PR #3, opened 2015-01-15
Fix ansible roles

sharph:
Execute graph builder from correct path
Install OTP over https instad of git:// (via SSH)


flibbertigibbet:
+1

While you're working on it, might as well update to use the Azavea OTP fork/develop branch (version 0.13).

PR #4, opened 2015-01-15
Feature/django

sharph:
Start OTP with "analyst" feature

Besides the ansible role to set up the database with PostGIS, I've added some code to settings.py to allow us to put our secrets in an external file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,3 +14,9 @@ Development Installation
 1. Make sure you have the development dependencies installed
 2. Place `septa_bus.zip`, `septa_rail.zip`, and `patco.zip` in the gtfs folder
 3. Run `vagrant up`
+4. `vagrant ssh`
+5. `. /app_env/activate`
+6. `cd /vagrant/python/cac_tripplanner`
+7. `./manage.py migrate`
+8. `./manage.py runserver [::]:8000`
+9. Access with your browser at http://localhost:8024/
--- a/deployment/ansible/roles/cac-tripplanner.app/handlers/main.yml
+++ b/deployment/ansible/roles/cac-tripplanner.app/handlers/main.yml
@@ -1,0 +1,3 @@
+- name: Enable PostGIS
+  command: "psql -d {{ postgres_db }} -c 'CREATE EXTENSION postgis;'"
+  sudo_user: postgres
--- a/deployment/ansible/roles/cac-tripplanner.app/meta/main.yml
+++ b/deployment/ansible/roles/cac-tripplanner.app/meta/main.yml
@@ -1,0 +1,5 @@
+---
+  dependencies:
+    - { role: "azavea.git" }
+    - { role: "azavea.nginx" }
+    - { role: "azavea.virtualenv" }
--- a/deployment/ansible/roles/cac-tripplanner.app/tasks/main.yml
+++ b/deployment/ansible/roles/cac-tripplanner.app/tasks/main.yml
@@ -9,12 +9,12 @@
     - python-psycopg2
 
 - name: Install pip packages
-  pip: name={{ item }}
+  pip: name={{ item.name }} version={{ item.version }}
        virtualenv=/app_env
   with_items:
-    - pyyaml
-    - django
-    - psycopg2
+    - { name: 'pyyaml', version: '3.11' }
+    - { name: 'django', version: '1.7.3' }
+    - { name: 'psycopg2', version: '2.5.4' }
 
 - name: Create postgres database
   postgresql_db: name={{ postgres_db }}
--- a/deployment/ansible/roles/cac-tripplanner.app/tasks/main.yml
+++ b/deployment/ansible/roles/cac-tripplanner.app/tasks/main.yml
@@ -1,0 +1,28 @@
+---
+- name: Install packages
+  apt: name={{ item }} state=present
+  with_items:
+    - postgresql-9.3
+    - postgresql-9.3-postgis-2.1
+    - libpq-dev
+    - python-dev
+    - python-psycopg2
+
+- name: Install pip packages
+  pip: name={{ item }}
+       virtualenv=/app_env
+  with_items:
+    - pyyaml
+    - django
+    - psycopg2
+
+- name: Create postgres database
+  postgresql_db: name={{ postgres_db }}
+  sudo_user: postgres
+  notify: Enable PostGIS
+
+- name: Create postgres user
+  postgresql_user: db={{ postgres_db }}
+                   name={{ postgres_user }}
+                   password={{ postgres_password }}
+  sudo_user: postgres
--- a/python/cac_tripplanner/cac_tripplanner/settings.py
+++ b/python/cac_tripplanner/cac_tripplanner/settings.py
@@ -1,0 +1,96 @@
+"""
+Django settings for cac_tripplanner project.
+
+For more information on this file, see
+https://docs.djangoproject.com/en/1.7/topics/settings/
+
+For the full list of settings and their values, see
+https://docs.djangoproject.com/en/1.7/ref/settings/
+"""
+
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+import os
+import yaml
+BASE_DIR = os.path.dirname(os.path.dirname(__file__))
+
+try:
+    secrets = yaml.safe_load(open('/etc/cac_secrets', 'r'))
+except (IOError, NameError):
+    secrets = {
+        'secret_key': '%&_DEVELOPMENT_SECRET_KEY_#42*pk!3y6lvk&1psyk=e=pr',
+        'database': {
+            'ENGINE': 'django.contrib.gis.db.backends.postgis',
+            'NAME': 'cac_tripplanner',
+            'USER': 'cac_tripplanner',
+            'PASSWORD': 'cac_tripplanner',
+            'HOST': '127.0.0.1',
+            'PORT': '5432'
+        },
+        'production': False
+    }
+
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/1.7/howto/deployment/checklist/
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = secrets['secret_key']
+
+# SECURITY WARNING: don't run with debug turned on in production!
+DEBUG = not secrets['production']
+
+TEMPLATE_DEBUG = not secrets['production']
+
+ALLOWED_HOSTS = []
+
+
+# Application definition
+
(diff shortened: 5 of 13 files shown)


deployment/ansible/roles/cac-tripplanner.app/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+    - postgresql-9.3
+    - postgresql-9.3-postgis-2.1
+    - libpq-dev
+    - python-dev
+    - python-psycopg2
+
+- name: Install pip packages
+  pip: name={{ item }}
+       virtualenv=/app_env
+  with_items:
```
CloudNiner: Wouldn't it be better to load these from a requirements.txt file so we can explicitly require version numbers?

python/cac_tripplanner/cac_tripplanner/settings.py
```diff
@@ -0,0 +1,96 @@
+"""
+Django settings for cac_tripplanner project.
```
CloudNiner: Typically we will template this file in ansible, so we can control dev/staging/prod variables at the group_vars level. Should we do that here as well?

deployment/ansible/roles/cac-tripplanner.app/tasks/main.yml
sharph: Yeah. We can put version numbers here but a requirements.txt is probably the cleaner way to do it.

ddohler: +1 to requiring versions; we can do something like:
`- { name: Django, version: 1.7.1 }` and then reference them with `{{ item.version }}`

CloudNiner:
Oops, saw you assigned to @kshepard .

Looks ok to me, should probably get a static build system in place as part of this card. Not quite sure what the best way to go about that is, should check with civic apps and/or @hectcastro since we haven't done a django project in awhile.

python/cac_tripplanner/cac_tripplanner/settings.py
kshepard: Yeah, I was just about to say the same thing. It'd be good to keep it in line with how we do it in other projects unless there's a good reason to change it.

sharph: I added code to load these types of settings from /etc/cac_secrets. I think this is cleaner than throwing the entire settings.py through the templating engine.

sharph: In other words, these vars can still be controlled at the group_vars level.

CloudNiner: Yeah, I actually do like this way and would be willing to go with it to see how it works out unless anyone has any major negatives they can think of.

deployment/ansible/roles/cac-tripplanner.app/tasks/main.yml
ddohler: We just migrated HunchLab away from requirements.txt to keeping the requirements in Ansible like this; I think that's cleaner because it keeps the application dependencies and versions within Ansible. If you use requirements.txt then you have to ensure that the requirements.txt file exists on whatever machine you're running Ansible against.

kshepard: +1 to @ddohler's suggestion.

CloudNiner: Ah makes sense. Mostly was worried about versioning, however we go about it.

python/cac_tripplanner/cac_tripplanner/settings.py
kshepard: Seems like it would be fine (and possibly a bit of an improvement) after thinking about it a bit. Sure, I'm game for giving it a try.

README.md
```diff
@@ -14,3 +14,9 @@ Development Installation
 1. Make sure you have the development dependencies installed
 2. Place `septa_bus.zip`, `septa_rail.zip`, and `patco.zip` in the gtfs folder
 3. Run `vagrant up`
+4. `vagrant ssh`
+5. `. /app_env/activate`
```
kshepard: Should be ./

sharph: No, `. /app_env/activate` is correct. /app_env/activate is not executable and `. ` tells bash to source the file.

sharph: Should be `/app_env/bin/activate` though...

python/cac_tripplanner/cac_tripplanner/settings.py
ddohler: I'm not sure I think this is a good idea; I personally find the apps where we template the whole settings.py file easier to work with than the ones where we split out the secrets, because it minimizes the amount of sleuthing you need to do to figure out what the value of a given setting is for a running application. That said, this type of thing rarely causes serious issues with a project, and this is a slightly different take on it than we've done before, so I'm willing to experiment.

notthatbreezy:
re: the static build system -- I think nyc-trees might be a good guide.

I think part of this card should also be splitting out the VMs into how they are eventually going to be deployed on AWS - e.g. an app VM with django, probably an OTPA VM since that will be in its own auto-scaling group behind an ELB probably, and then a services/catch-all VM for the database.

notthatbreezy:
Probably want gunicorn/nginx set up as well since most of the time when we develop django applications we try and keep production as close to development as possible since bugs sometimes slip in from relying on runserver vs. gunicorn all the time in development. Hunchlab, OTI, and other apps we've worked on probably serve as a good starting point.

Don't know what other people's thoughts on this are though.

README.md
kshepard: Oh, ok, I didn't know about that syntax for source before.

ddohler: Just a thought, we don't normally use virtualenvs inside Vagrant boxes; do we need it here?

kshepard:
+1, looks like a good start. I agree with others that it would be nice to flesh this out some more with a build system, switching to nginx, etc., which can be done in separate PRs.

deployment/ansible/roles/cac-tripplanner.app/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
+---
+- name: Install packages
+  apt: name={{ item }} state=present
+  with_items:
+    - postgresql-9.3
```
notthatbreezy: Might be good to use the azavea roles for postgresql and postgis

deployment/ansible/roles/cac-tripplanner.app/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+    - postgresql-9.3-postgis-2.1
+    - libpq-dev
+    - python-dev
```
notthatbreezy: There's an azavea role for installing [python](https://github.com/azavea/ansible-python) as well we could use

deployment/ansible/roles/cac-tripplanner.app/meta/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+  dependencies:
+    - { role: "azavea.git" }
+    - { role: "azavea.nginx" }
```
notthatbreezy: I think `azavea.nginx` is missing from roles.txt

deployment/ansible/roles/cac-tripplanner.app/handlers/main.yml
```diff
@@ -0,0 +1,3 @@
+- name: Enable PostGIS
+  command: "psql -d {{ postgres_db }} -c 'CREATE EXTENSION postgis;'"
+  sudo_user: postgres
```
hectcastro: PostGIS is enabled like this in NYC to help Ansible know the task is idempotent:

```yaml
- name: Add PostGIS extension
  sudo_user: postgres
  command: psql {{ postgresql_database }} -c "CREATE EXTENSION postgis"
  register: psql_result
  failed_when: >
    psql_result.rc != 0 and ("already exists" not in psql_result.stderr)
  changed_when: "psql_result.rc == 0"
```

deployment/ansible/roles/cac-tripplanner.app/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+    - python-psycopg2
```
hectcastro: [azavea.postgresql-support](https://github.com/azavea/ansible-postgresql-support) handles installing dependencies for the Ansible PostgreSQL modules.

deployment/ansible/roles/cac-tripplanner.app/tasks/main.yml
hectcastro: How long have you all been running with this approach? Sometimes the `pip` module doesn't detect changes to the `requirements.txt` file (I think when they're pinned to a Git SHA). Would be nice to get away from that if this has been successful.

python/cac_tripplanner/cac_tripplanner/settings.py
hectcastro: FWIW, we've been trying to pull everything in the Django settings file from the environment and then launch the Django application using `envdir`. Adds some noise before certain commands, but allows us to commit the settings files into the repository.

deployment/ansible/roles/cac-tripplanner.app/tasks/main.yml
notthatbreezy: It's been working on HunchLab for a week or so. We've been using the same approach for ruby gems for a while now and that has been good.

It definitely makes packer integration easier since it doesn't require pushing up the requirements.txt file to the instance.

deployment/ansible/roles/cac-tripplanner.app/meta/main.yml
sharph: Good catch. Why did it work for me?

sharph:
Added versions and missing nginx to roles.txt. I'm going to go ahead and merge and make other improvements in separate PR.

README.md
sharph: No, we don't. I was just going by conventions I've seen in other projects.

ddohler: Yeah, we don't normally put virtualenvs inside of Vagrant boxes since they serve overlapping purposes. The ones you've probably seen (Cicero and OpenDataPhilly) have them for historical reasons. Cicero is old and was migrated onto Vagrant, while with ODP we used pre-existing installation scripts / packages to save time and there was no reason to mess with it once it was working. It's definitely not a big deal though.

sharph: I've gotten rid of the virtualenv in my working branch.

PR #6, opened 2015-01-19
Add frontend boilerplate

moradology:
Not sure how to wire up dependencies - decided to just go ahead and add to settings.py's static_dirs.

Running 'gulp watch' from within '/vagrant/src' will automatically collectstatic for django runserver's usage.


deployment/ansible/roles/cac-tripplanner.app/defaults/main.yml
```diff
@@ -5,3 +5,6 @@ postgres_password: "cac_tripplanner"
 postgres_host: "192.168.8.25"
 django_workers: 1
 production: False
+
+app_sass_version: "3.4.9"
+gulp_dir: "/vagrant/src"
```
sharph: Change to `/opt/app`

python/cac_tripplanner/cac_tripplanner/settings.py
```diff
@@ -94,3 +94,9 @@
 # https://docs.djangoproject.com/en/1.7/howto/static-files/
 
 STATIC_URL = '/static/'
+STATIC_ROOT = os.path.join(BASE_DIR, 'cac_tripplanner', 'static')
+STATICFILES_DIRS = (
+    '/vagrant/src/app/scripts',
```
sharph: Here as well.

src/app/index.html
```diff
@@ -0,0 +1,95 @@
+<!doctype html>
```
notthatbreezy: We're still using django templates right?

moradology: Absolutely. Gulp's collectstatic call (by way of django) only pulls css and js files.

sharph:
LGTM as far as ansible is concerned. Maybe @CloudNiner might want to look at the gulp stuff.

deployment/ansible/roles/cac-tripplanner.app/tasks/main.yml
```diff
@@ -1,4 +1,6 @@
 ---
+- { include: jslibs.yml }
```
sharph: Nice. We should do this more often.

python/cac_tripplanner/cac_tripplanner/settings.py
```diff
@@ -94,3 +94,10 @@
 # https://docs.djangoproject.com/en/1.7/howto/static-files/
 
 STATIC_URL = '/static/'
+STATIC_ROOT = os.path.join(BASE_DIR, 'cac_tripplanner', 'static')
+BUILD_DIR = '/opt/app/src'
```
sharph: It might be good to put this in the secrets dict above so we can modify with group_vars later.

src/.editorconfig
```diff
@@ -0,0 +1,25 @@
...
+# editorconfig.org
+
+root = true
+
+
+[*]
+
+# change these settings to your own preference
+indent_style = space
+indent_size = 2
```
CloudNiner: Set to 4

src/.jshintrc
```diff
@@ -0,0 +1,18 @@
+{
+  "browser": true,
+  "esnext": true,
+  "bitwise": true,
+  "camelcase": true,
+  "curly": true,
+  "eqeqeq": true,
+  "immed": true,
+  "indent": 2,
```
CloudNiner: Set to 4

src/app/index.html
CloudNiner: In that case, we don't need this index file at all correct? Since our html files will be placed in the django templates directory?

CloudNiner:
Add README detailing how to get going on frontend dev would be helpful, since this is different from what we've done before.

Otherwise, LGTM

src/app/index.html
notthatbreezy: I agree, if this can be deleted we should dump this file to eliminate confusion

deployment/ansible/roles/cac-tripplanner.app/tasks/main.yml
notthatbreezy: I think this needs to go after django is installed and the application is setup since it calls collectstatic this include fails

deployment/ansible/roles/cac-tripplanner.app/tasks/jslibs.yml
```diff
@@ -0,0 +1,33 @@
...
+       user_install=no
+       state=present
+
+- name: Install Compass
+  gem: name=compass
+       user_install=no
+       state=present
+
+- name: Install bower and gulp
+  command: npm install -g bower gulp
```
notthatbreezy: Without using the npm module for ansible this task keeps failing after running once

deployment/ansible/roles/cac-tripplanner.app/meta/main.yml
```diff
@@ -2,4 +2,6 @@
   dependencies:
     - { role: "azavea.git" }
     - { role: "azavea.nginx" }
+    - { role: "azavea.nodejs" }
```
notthatbreezy: Had to update npm to get the npm prereqs to install.

src/gulpfile.js
```diff
@@ -0,0 +1,107 @@
+/* jshint node:true */
+'use strict';
+// generated on 2015-01-19 using generator-gulp-webapp 0.2.0
+var gulp = require('gulp');
+var shell = require('gulp-shell');
+var $ = require('gulp-load-plugins')();
+
+gulp.task('collectstatic', shell.task(['python /vagrant/python/cac_tripplanner/manage.py collectstatic --noinput']));
```
notthatbreezy: Not sure if there's a good way to template this out or not, but that absolute path in there probably isn't going to work for very long.

PR #7, opened 2015-01-19
Change nginx port to match user's port

sharph:
This change makes django redirects work.


moradology:
+1

notthatbreezy:
Not sure I understand what was going on here - but it might be a good idea to keep nginx on a standard port since that is how it will be deployed.

May want to look at using the following in the nginx config `proxy_set_header Host $http_host;` to preserve host and port for redirects since django supports that.

See: https://docs.djangoproject.com/en/1.7/ref/request-response/#django.http.HttpRequest.get_host and http://stackoverflow.com/questions/15414810/whats-the-difference-of-host-and-http-host-in-nginx

PR #10, opened 2015-01-20
Feature/homepage

moradology:


python/cac_tripplanner/templates/cac_tripplanner/cac_tripplanner_home.html
```diff
@@ -0,0 +1,79 @@
+{% extends "cac_tripplanner/base_cac_tripplanner.html" %}
```
CloudNiner: For the size of this app, and to follow typical django convention, these templates should probably just be:
```
python/cac_tripplanner/templates/base.html
python/cac_tripplanner/templates/home.html
```
If you need a subfolder to namespace your templates, you should just use different django apps instead.

moradology: I was trying to stay close to the two-scoops style - but you're right, this application probably doesn't warrant such heavy conventions.

src/gulpfile.js
```diff
@@ -5,7 +5,7 @@ var gulp = require('gulp');
 var shell = require('gulp-shell');
 var $ = require('gulp-load-plugins')();
 
-gulp.task('collectstatic', shell.task(['python /vagrant/python/cac_tripplanner/manage.py collectstatic --noinput']));
+gulp.task('collectstatic', shell.task(['python ../python/cac_tripplanner/manage.py collectstatic --noinput']));
```
ddohler: I think this is okay for now, but we may eventually want to revisit this since it requires the `src` and `python` directories to be siblings.

ddohler:
+1 -- my Django template knowledge is a bit rusty, but this looks good.

PR #13, opened 2015-01-20
Copy and serve static files for nginx

notthatbreezy:
Mostly fixes gulp and nginx so that they
work with each other. This copies the django
static files to dist/ and the other static
files there as well.

TODO: Figure out a sane strategy for static
files.


CloudNiner:
+1

PR #14, opened 2015-01-20
Feature/map page

CloudNiner:


python/cac_tripplanner/templates/base.html
```diff
@@ -3,12 +3,12 @@
...
-    <title>{% block title %}CAC Trip Planner{% endblock %}</title>
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <meta name="description" content="">
-    <meta name="author" content="">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=0">
+    <title>{% block title %}Open Transit Planning for Bikes, Public Transit, and Walking | Clean Air Council{% endblock %}</title>
 
     {% block cssimports %}
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.2/css/bootstrap.min.css">
```
CloudNiner: @moradology is this how this dependency should be added?

python/cac_tripplanner/cac_tripplanner/views.py
```diff
@@ -3,5 +3,8 @@
 def home(request):
     return render_to_response('home.html')
 
+def map(request):
```
notthatbreezy: Are we going to be using function based views?

CloudNiner: You know, I don't have a strong opinion. Seeing as there are going to be only a few html based views and maybe an API endpoint or two, function based may be a simpler choice?

notthatbreezy: Sounds fine to me, I default to class based because that's what we have been doing on other projects and what I started on; but, I don't see a lot of inheritance/re-use going on in this project so we can probably just use functions.

python/cac_tripplanner/templates/base.html
moradology: Either this or in a separate mixin file - I think the way you did it is best since every page we have is going to use it.

moradology:
+1

CloudNiner:
Updated with an additional commit that includes js stuff.

python/cac_tripplanner/cac_tripplanner/views.py
ddohler: A bit late to the party, but +1 for function-based views; I doubt this is going to be the type of app where class-based views would really shine.

notthatbreezy:
+1

PR #17, opened 2015-01-21
Finalize build system

kshepard:


python/cac_tripplanner/cac_tripplanner/settings.py
```diff
@@ -41,8 +41,10 @@
 
 TEMPLATE_DEBUG = not secrets['production']
 
-ALLOWED_HOSTS = []
+# TODO: add the load balancer DNS to this list when in production
+ALLOWED_HOSTS = ['127.0.0.1', 'localhost']
```
CloudNiner: Wouldn't it be best to configure this via the secrets dict?

kshepard: Oh yeah, that sounds like a better idea.

python/cac_tripplanner/cac_tripplanner/views.py
```diff
@@ -1,10 +1,9 @@
 from django.shortcuts import render_to_response
+from django.template import RequestContext
+from django.conf import settings
 
 def home(request):
-    return render_to_response('home.html')
+    return render_to_response('home.html', dict(debug=settings.DEBUG))
```
CloudNiner: Hmm I thought it was possible in templates to get settings values without passing them through.

Maybe that was a RequestContext() thing though.

CloudNiner: Ah, here it is:
https://docs.djangoproject.com/en/1.7/ref/templates/api/#django-core-context-processors-debug

notthatbreezy: Yea, we did too -- couldn't quite get it to work though. The docs seemed to suggest you specifically had to pass in `RequestContext(request)`, but that didn't seem to work either.

src/gulpfile.js
```diff
@@ -1,110 +1,134 @@
...
+var pythonRoot = '/opt/app/python/cac_tripplanner';
+
+var stat = {
+    scripts: staticRoot + '/scripts',
+    styles: staticRoot + '/styles',
+    images: staticRoot + '/images'
+};
+
+// The multiple-select module needs special treatment, because its bower.json file is incomplete
+var multiSelectRoot = 'bower_components/multiple-select';
```
CloudNiner: Of course...

python/cac_tripplanner/cac_tripplanner/views.py
notthatbreezy: Ohh, now looking at that, I think it was because we didn't have `INTERNAL_IPS` set correctly at the time.

deployment/ansible/roles/cac-tripplanner.app/files/default
```diff
@@ -9,7 +9,6 @@ server {
 	}
 
         location /static/ {
-                alias /opt/app/src/dist/;
-
+                alias /srv/cac/;
```
ddohler: This should probably be made into  `{{ root_static_dir }}`, I think.

CloudNiner:
+1

python/cac_tripplanner/cac_tripplanner/views.py
kshepard: That could be it. I'll give it another shot.

deployment/ansible/roles/cac-tripplanner.app/files/default
kshepard: Good point. I suppose it'll need to be moved from files to templates too.

deployment/ansible/roles/cac-tripplanner.app/tasks/jslibs.yml
```diff
@@ -11,24 +11,29 @@
        state=present
 
 - name: Install NPM dependencies
-  npm: path="{{ gulp_dir }}" state=present
+  npm: path="{{ root_src_dir }}" state=present
```
hectcastro: We switched to using `npm install` directly in NYC because of the discussion here: https://github.com/azavea/nyc-trees/pull/124#issuecomment-66536399

deployment/ansible/roles/cac-tripplanner.app/tasks/jslibs.yml
```diff
@@ -11,24 +11,29 @@
...
   sudo: no
 
-- name: Create Static Directories
-  file: path="{{ gulp_dir }}/{{ item }}" state=directory mode=0755
-  sudo: no
-  with_items:
-    - "dist/scripts"
-    - ".tmp/styles"
+- name: Create Static Directory
+  file: path="{{ root_static_dir }}" state=directory mode=0755 owner="{{ app_username }}"
```
hectcastro: This is pretty clever.

deployment/ansible/roles/cac-tripplanner.app/tasks/jslibs.yml
ddohler: This may explain some problems I'm having on Jenkins right now, I'll look into switching it.

PR #18, opened 2015-01-22
Calculate isochrone and find destinations within it

moradology:
This PR includes code for isochrone calculation/filtering based on isochrone results and the front end component which will make the request.

A potential weak spot is the JS component, which currently doesn't have robust safeguards against poorly formed requests.


python/cac_tripplanner/cac_tripplanner/views.py
```diff
@@ -1,9 +1,74 @@
...
 def map(request):
     return render_to_response('map.html', context_instance=RequestContext(request))
+
+
+class Reachable(View):
+    """Class based view for finding destinations of interest within the calculated isochrone"""
+    # TODO: make decisions on acceptable ranges of values that this endpoint will support
+
+    otp_router = 'default'
+    otp_url = 'http://192.168.8.26:8080/otp/routers/{router}/isochrone'.format(router=otp_router)
```
notthatbreezy: This URL should be a parameter in settings

python/cac_tripplanner/cac_tripplanner/views.py
```diff
@@ -1,9 +1,74 @@
...
+        isochrone_multipoly = MultiPolygon(polygons)
+        return isochrone_multipoly
+
+
+    def get(self, request, *args, **kwargs):
+        """When a GET hits this endpoint, calculate an isochrone and find destinations within it"""
+        params = request.GET
+
+        iso = self.isochrone(
+            lat=params.get('coords[lat]'),
```
notthatbreezy: Any reason not to just have parameters be lat,lng,mode, etc instead of the brackets?

python/cac_tripplanner/cac_tripplanner/views.py
```diff
@@ -1,9 +1,74 @@
...
+class Reachable(View):
+    """Class based view for finding destinations of interest within the calculated isochrone"""
+    # TODO: make decisions on acceptable ranges of values that this endpoint will support
+
+    otp_router = 'default'
+    otp_url = 'http://192.168.8.26:8080/otp/routers/{router}/isochrone'.format(router=otp_router)
+    algorithm = 'accSampling'
+
+    def isochrone(self, lat, lng, mode, date, time, max_travel_time, max_walk_distance):
+        """Make request to Open Trip Planner for isochrone geometry - convert it to GDAL Polygon"""
```
notthatbreezy: Docstring needs fleshing out

python/cac_tripplanner/cac_tripplanner/tests.py
```diff
@@ -0,0 +1,29 @@
...
+import json
+
+from django.test import TestCase
+
+from django.contrib.gis.geos import GEOSGeometry
+
+from destinations.models import Destination
+
+class CACTripPlannerIsochroneTestCase(object):
+    """ Test behavior of within-isochrone view """
```
notthatbreezy: Is there a test being run here?

python/cac_tripplanner/cac_tripplanner/views.py
moradology: I like the organization and that's what `$.param` output looks like. I'm open to modifying it, however.

python/cac_tripplanner/cac_tripplanner/urls.py
```diff
@@ -1,13 +1,15 @@
...
 from django.contrib.staticfiles import views as staticviews
 
+from .views import Reachable
+import settings
 
 urlpatterns = patterns('',
     url(r'^$', 'cac_tripplanner.views.home', name='home'),
-    url(r'^map/', 'cac_tripplanner.views.map', name='map'),
+    url(r'^map/$', 'cac_tripplanner.views.map', name='map'),
+    url(r'^reachable$', Reachable.as_view(), name='reachable'),
```
sharph: Should we put this endpoint under `map/`?

python/cac_tripplanner/cac_tripplanner/views.py
notthatbreezy: Doesn't matter was really just wondering. If it's not going to be used except by the javascript and not by humans, you can leave it; however, if it was something more user-facing I'd err on readability probably.

python/cac_tripplanner/cac_tripplanner/views.py
```diff
@@ -1,9 +1,74 @@
...
+            lat=params.get('coords[lat]'),
+            lng=params.get('coords[lng]'),
+            mode=params.getlist('mode[]'),
+            date=params.get('date'),
+            time=params.get('time'),
+            max_travel_time=params.get('maxTravelTime'),
+            max_walk_distance=params.get('maxWalkDistance')
+        )
+
+        matched_objects = Destination.objects.filter(point__within=iso)
```
sharph: Filter on `published`.

python/cac_tripplanner/cac_tripplanner/views.py
```diff
@@ -1,9 +1,74 @@
...
 def home(request):
     print request.META['REMOTE_ADDR']
     return render_to_response('home.html', context_instance=RequestContext(request))
 
+
 def map(request):
     return render_to_response('map.html', context_instance=RequestContext(request))
+
+
+class Reachable(View):
```
notthatbreezy: We had decided on using function based views earlier, may want to stick to that. Though I do think the class structure here groups related methods nicely.

python/cac_tripplanner/cac_tripplanner/views.py
```diff
@@ -1,9 +1,74 @@
...
+            'routerId': self.otp_router,
+            'fromPlace': lat + ',' + lng,
+            'mode': (',').join(mode),
+            'date': date,
+            'time': time,
+            'cutoffSec': max_travel_time,
+            'maxWalkDistance': max_walk_distance,
+            'algorithm': self.algorithm
+        }
+        isochrone_response = get(self.otp_url, params=payload)
```
sharph: `get` is such a generic name that we might want to be more explicit here with `requests.get()`.

src/app/scripts/cac/cac-isochrone.js
```diff
@@ -0,0 +1,58 @@
+CAC.Pages.Home = (function ($) {
```
CloudNiner: Need to change where this gets assigned to. Probably CAC.Isochrone? Or something slightly more generic?

src/app/scripts/cac/cac-isochrone.js
```diff
@@ -0,0 +1,58 @@
...
+            var requestUrl = isochroneUrl + urlParams;
+            var promise = $.ajax({
+                type: 'GET',
+                url: requestUrl,
+                contentType: 'application/json'
+            });
+            return promise;
+        } else {
+            console.log("INCOMPLETE PAYLOAD: UNABLE TO CALCULATE ISOCHRONE")
+            return false;
```
CloudNiner: For consistency and usability, you should create a promise here and reject it with an error message or object.

See jQuery deferred object: http://api.jquery.com/category/deferred-object/

python/cac_tripplanner/cac_tripplanner/views.py
```diff
@@ -1,9 +1,74 @@
...
+        print('pay', payload)
+
+        # Parse and traverse JSON from OTP so that we return only geometries
+        json_poly = json.loads(isochrone_response.content)[0]['geometry']['geometries']
+
+        # Iterate through the returned geometries and convert them to proper OGR geoms
+        polygons = []
+        for _, poly in enumerate(json_poly):
```
sharph: _ isn't used. You could use `map` here.

sharph: (or a list comprehension.)

src/app/scripts/cac/cac-isochrone.js
moradology: Good catch.

python/cac_tripplanner/cac_tripplanner/views.py
```diff
@@ -1,9 +1,74 @@
...
+            lng=params.get('coords[lng]'),
+            mode=params.getlist('mode[]'),
+            date=params.get('date'),
+            time=params.get('time'),
+            max_travel_time=params.get('maxTravelTime'),
+            max_walk_distance=params.get('maxWalkDistance')
+        )
+
+        matched_objects = Destination.objects.filter(point__within=iso)
+        return HttpResponse('{{"matched": "{0}"}}'.format(matched_objects), 'application/json')
```
sharph: I'm confused about what happens here. Does `matched_objects` serialize into GeoJSON?

python/cac_tripplanner/cac_tripplanner/views.py
sharph: I'm kind of torn too. If we go the function based route, we might move the isochrone function to a different file.

ddohler: We could do both; they have different use cases. http://www.django-rest-framework.org/api-guide/views/#function-based-views

As an aside, I think the name of this class could be made a bit clearer; it's a bit generic if you don't know the context.

python/cac_tripplanner/cac_tripplanner/views.py
ddohler: It would be nice not to have this call to an outside service in the view although this is definitely simpler for the moment since I think the alternative is modifying OTP.

python/cac_tripplanner/cac_tripplanner/tests.py
```diff
@@ -0,0 +1,68 @@
...
+        response = self.client.get(isochrone_url)
+        self.assertEqual('{"matched": "[<Destination: testWithin>]"}', response.content)
+
+
+
+
+
+
+
+
```
sharph: Extra newlines at the end of a file don't usually bother me but this seems like a lot of them.

sharph:
+1

PR #19, opened 2015-01-22
Feature/content templates

notthatbreezy:
Still needs styling, which I'll add if we get html prototypes, or @mtedeschi could just edit these templates, not sure if one way is preferred over the other -- doesn't really matter to me.

### DIFF
--- a/deployment/ansible/roles/cac-tripplanner.app/defaults/main.yml
+++ b/deployment/ansible/roles/cac-tripplanner.app/defaults/main.yml
@@ -10,3 +10,13 @@ app_sass_version: "3.4.9"
 
 root_src_dir: "/opt/app/src"
 root_static_dir: "/srv/cac"
+
+cac_python_dependencies:
+    - { name: 'pyyaml', version: '3.11' }
+    - { name: 'django', version: '1.7.3' }
+    - { name: 'psycopg2', version: '2.5.4' }
+    - { name: 'pytz', version: '2014.10' }
+    - { name: 'django-ckeditor', version: '4.4.7' }
+    - { name: 'django-extensions', version: '1.4.9' }
+    - { name: 'ipython', version: '2.3.1' }
+    - { name: 'django-wpadmin', version: '1.7.0'}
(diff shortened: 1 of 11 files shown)


deployment/ansible/roles/cac-tripplanner.app/defaults/main.yml
```diff
@@ -10,3 +10,13 @@ app_sass_version: "3.4.9"
 
 root_src_dir: "/opt/app/src"
 root_static_dir: "/srv/cac"
+
+cac_python_dependencies:
```
CloudNiner: Is it worth it to split this into a dependencies and a dev dependencies so that things like ipython don't get installed in production?

python/cac_tripplanner/cms/models.py
```diff
@@ -5,14 +5,25 @@
...
+            (tips_and_tricks, 'Tips and Tricks'),
+        )
+
     title = models.CharField(max_length=80)
     slug = models.SlugField()
     author = models.ForeignKey(User)
     teaser = RichTextField()  # above the fold
     content = RichTextField(null=True, blank=True)  # below the fold
+    publish_date = models.DateTimeField(blank=True, null=True)
     published = models.BooleanField(default=False)
```
CloudNiner: It would probably be better if published were removed here, and a method is_published() was added to the model, since keeping publish_date and published in sync is a pain

notthatbreezy: True, that's a good point.

deployment/ansible/roles/cac-tripplanner.app/defaults/main.yml
notthatbreezy: It doesn't really hurt anything.

Also, though we strive not to, inevitably sometimes it is necessary to debug something on the server and having `django_extensions` and `ipython` there makes that easier.

CloudNiner:
+1 after replacing published field with a method

deployment/ansible/roles/cac-tripplanner.app/defaults/main.yml
CloudNiner: Agree. Up to you.

PR #22, opened 2015-01-22
Replace cac-tripplanner.otp role with azavea.opentripplanner

CloudNiner:
OTP box reprovisioned successfully, and updated the septa_rail gtfs file on fileshare.


moradology:
Might make sense to get this merged in before team lunch so people can let their boxes work on destroying/reupping

deployment/ansible/otp.yml
```diff
@@ -7,5 +7,4 @@
       apt: update_cache=yes
 
   roles:
-    - { role: "cac-tripplanner.otp"}
```
sharph: I understand azavea.opentripplanner will be run by dependency, but maybe it makes sense to define it here explicitly too since at the end of the day we want this VM to run the OTP server.

sharph: @hectcastro Maybe you have an idea of what the convention should be? I've run into this more than once.

sharph:
+1

deployment/ansible/otp.yml
hectcastro: I try not to create duplicates if a role is already listed as a dependency because it can cause the steps to get run multiple times (it'll be an OK instead of a CHANGED, but that still takes up some amount of time).

CloudNiner: Makes sense. I don't think its worth the duplication since the role dependencies are still explicitly stated in the role itself.

PR #23, opened 2015-01-22
Add layer control with two basemaps + stubbed overlays

CloudNiner:


src/bower.json
```diff
@@ -7,5 +7,10 @@
     "leaflet": "^0.7.3",
     "multiple-select": "~1.1.0",
     "ramda": "0.8.0"
+  },
+  "overrides": {
+    "leaflet": {
+      "ignore": true
```
CloudNiner: @ddohler this should make leaflet available to tests, but not include it in the vendor build, since it was easier to include them via CDN. The css is broken if it is included via gulp.

notthatbreezy: We can include the css via gulp, I don't think @kshepard and I realized it was needed when we did gulp.

notthatbreezy:
+1

src/bower.json
CloudNiner: True, and it was being included. The bigger issue is that img paths get mangled by the static files copy so things end up in the wrong place and fixing that would require an effort similar to the multipleselect library, but more complicated since there are more images.

notthatbreezy: Got it.

kshepard: Maybe we need to modify the way bower images are copied over. We're currently sticking them all in one images directory, which sounds nice in theory, but pretty much all css uses relative image links and would require css overrides. We should probably try to maintain that directory structure when copying them over, or something along those lines.

CloudNiner: I agree that this would be the ideal solution. That _should_ be trivial, but maybe it's not.

PR #24, opened 2015-01-22
Add random tips/tricks and profile to home page

notthatbreezy:
This commit adds random tips & tricks and a community profile to the home page.

Some custom managers are used to DRY out queries and simplify things.


moradology:
+1. Good use of Django orm managers

PR #25, opened 2015-01-22
Update .jshintrc

moradology:


notthatbreezy:
Might be better to just excise the console statements since setting `devel: true` won't help us catch console statements that may break some browsers like IE.

PR #27, opened 2015-01-22
Configure and upgrade Gunicorn

kshepard:


deployment/ansible/roles/cac-tripplanner.app/templates/upstart-cac-tripplanner-app.conf.j2
```diff
@@ -0,0 +1,14 @@
...
+{% else %}
+start on (local-filesystems and net-device-up IFACE!=lo)
+{% endif %}
+stop on shutdown
+
+respawn
+setuid {{ app_username }}
+chdir {{ root_app_dir }}
+
+exec gunicorn --config {{ root_conf_dir }}/gunicorn.py cac_tripplanner.wsgi
```
moradology: Seems to me like this could be added to azavea.gunicorn with a conditional flag or the like (and with project specific text removed). Is there a reason we don't have something like that?

kshepard: That's a good point. What do you think @hectcastro?

notthatbreezy: Yeah, I don't think we have an azavea gunicorn role, but it probably isn't a bad idea. Only potential issue is that there are probably quite a few overrides necessary for anyone to use it at all.

notthatbreezy:
+1 provisioned and loaded fine

kshepard:
Cool, thanks for testing it. Mine provisioned fine too, but I was starting from a fresh VM. I wasn't exactly sure what would happen with a VM that already had the old version of Gunicorn (from apt) running. I suppose if anyone runs into errors, just manually stop the existing Gunicorn service.

deployment/ansible/roles/cac-tripplanner.app/templates/upstart-cac-tripplanner-app.conf.j2
hectcastro: Yep, there isn't one right now. Not a bad idea, but as Chris said, it would need to expose a bunch of the attributes so that people could override them for their specific needs.

Also, here is a list of the existing roles: https://galaxy.ansible.com/list#/users/6541

PR #28, opened 2015-01-22
Mark build unstable on linting errors

ddohler:
The build still currently fails because one of the tests relies on having a running OTP server, which we don't have at the moment. But this will mark the build unstable if linting fails, rather than failing the whole build.

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+function mark_unstable {
+    java -jar jenkins-cli.jar -s $JENKINS_URI set-build-result unstable
+}
+
+set -x
+trap 'mark_unstable' ERR
+
+# Python linting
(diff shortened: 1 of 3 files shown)


moradology:
+1

scripts/lint.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+function mark_unstable {
+    java -jar jenkins-cli.jar -s $JENKINS_URI set-build-result unstable
+}
+
+set -x
+trap 'mark_unstable' ERR
```
hectcastro: Interesting approach.

ddohler: I sort of don't like that this puts some Jenkins-specific stuff directly into this script, but I tried doing something like `./scripts/lint.sh || java -jar jenkins-cli.jar -s $JENKINS_URI set-build-result unstable` and it didn't want to work.

PR #29, opened 2015-01-23
Use/Watch NFS to speed up gulp-watch

notthatbreezy:
This switches the project to use NFS
for the app VM to speed up gulp watches.

Also tunes NFS to reduce the period that
the NFS client will cache attribtues of
directories and files to 2 seconds.


CloudNiner:
So many plus ones.

Gotta remember to add this to other projects.

moradology:
:kiss:

PR #31, opened 2015-01-26
Add link shortening functionality

ddohler:
The links are currently not very short (24-character keys) because I was conservative about potential link collisions and used UUIDs. We might want to change this (to use hashing or shorter keys or both), so I put the link-shortening logic in a separate class so that it's easier to change. This also includes a hit-tracking object which allows us to track when shortened links are dereferenced.

### DIFF
--- a/python/cac_tripplanner/cac_tripplanner/settings.py
+++ b/python/cac_tripplanner/cac_tripplanner/settings.py
@@ -67,6 +67,7 @@ INSTALLED_APPS = (
     # Project Apps
     'cms',
     'destinations',
+    'shortlinks',
 )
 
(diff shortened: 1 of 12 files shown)


sharph:
One way to generate very short urls might be to use a [base58](http://en.wikipedia.org/wiki/Base58) encoding of the primary key. This will keep the URLs as short as possible and also remove the need for a key column.

ddohler:
Interesting, I like it--wish I had discovered that when I was researching how to do this. I'm pretty much out of time on this card, but I'll put in a to-do because that sounds like it will work pretty well.

sharph:
+1

PR #32, opened 2015-01-26
Feature/routing ui

moradology:
This changeset organizes scripts into folders, sets up a simple interface to OTP's planning endpoint, creates some logic for handling mode selection (and persistence through a session should we desire that), and adds some mock data on which plans might be calculated.
Also: it sets up the geolocation API to watch user location changes.


moradology:
This might also be good for @notthatbreezy to take a look at.

moradology:
screen shot 2015-01-26 at 10 17 13 am 2

python/cac_tripplanner/templates/base.html
```diff
@@ -30,18 +30,23 @@
...
+    <script src="{% static 'scripts/vendor/lodash.js' %}"></script>
 
     <script src="{% static 'scripts/main/cac/cac.js' %}"></script>
-    <script src="{% static 'scripts/main/cac/cac-map-control.js' %}"></script>
-    <script src="{% static 'scripts/main/cac/cac-map-overlays.js' %}"></script>
-    <script src="{% static 'scripts/main/cac/cac-pages-home.js' %}"></script>
-    <script src="{% static 'scripts/main/cac/cac-pages-map.js' %}"></script>
-    <script src="{% static 'scripts/main/cac/cac-isochrone.js' %}"></script>
+    <script src="{% static 'scripts/main/cac/user/cac-user-modes.js' %}"></script>
+    <script src="{% static 'scripts/main/cac/map/cac-map-control.js' %}"></script>
```
CloudNiner: Good call.

src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -0,0 +1,127 @@
...
+                userMarker.setLatLng(latlng);
+            } else {
+                userMarker = new L.CircleMarker(latlng)
+                  .on('click', function() {
+                      $(document).trigger('MOS.Map.Control.CurrentLocationClicked', latlng);
+                  });
+                userMarker.setRadius(50);
+                map.addLayer(userMarker);
+            }
+            if (targetLocation.lat === latlng[0] && targetLocation.lng === latlng[1]) {
```
CloudNiner: This would probably be good to buffer with a small area, or to a given decimal precision. Otherwise, it'll never be true.

src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -0,0 +1,127 @@
...
+            maximumAge: 0
+        };
+        var success = function(pos) {
+            var latlng = [pos.coords.latitude, pos.coords.longitude];
+            if (userMarker) {
+                userMarker.setLatLng(latlng);
+            } else {
+                userMarker = new L.CircleMarker(latlng)
+                  .on('click', function() {
+                      $(document).trigger('MOS.Map.Control.CurrentLocationClicked', latlng);
```
CloudNiner: Typically events are fired either by wrapping the object (in this case MapControl) itself, or exposing an events property that you fire the events on. This avoids polluting the global namespace and having a bunch of other event checkers execute since the document is the tree root for event propagation.

So, an example:
```
// In Map.Control
function MapControl(options) {
    ...
   this.events = $({});
}

function locateUser() {
    ...
    MapControl.events.trigger('MOS.Map.Control.CurrentLocationClicked', latlng);
}

// In Map.Pages
...
mapControl.events.on('MOS.Map.Control.CurrentLocationClicked', function () { ... });
```
}

src/app/scripts/cac/map/cac-map-routing.js
```diff
@@ -0,0 +1,56 @@
...
+    function planTrip(coordsFrom, coordsTo) {
+        var deferred = $.Deferred();
+        var urlParams = prepareParamString(coordsFrom, coordsTo);
+        var requestUrl = routingUrl + '?' + urlParams;
+        $.ajax({
+            url: requestUrl,
+            type: 'GET',
+            contentType: 'application/json'
+        }).then(deferred.resolve);
+        return deferred.promise();
```
CloudNiner: _Should_ just be able to do:
```
return $.ajax();
```
here

src/app/scripts/cac/map/cac-map-routing.js
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Helper function to prepare the parameter string for consumption by the OTP api
+     *
+     * @param coordsFrom {array} The coords in lat-lng which we would like to travel from
+     * @param coordsTo {array} The coords in lat-lng which we would like to travel to
+     *
+     * @return {string} A set of get params, ready for consumption
+     */
+    function prepareParamString(coordsFrom, coordsTo) {
+        var currentTime = moment();
```
CloudNiner: Ah. Nice.

src/app/scripts/cac/pages/cac-pages-map.js
```diff
@@ -0,0 +1,69 @@
...
+            var coords = lat + ',' + lng;
+            $('section.directions input.origin').val(coords);
+        });
+        $(document).bind('MOS.Map.Control.DestinationClicked', function(e, feature) {
+            var coords = feature.geometry.coordinates[1] + ',' + feature.geometry.coordinates[0];
+            $('section.directions input.destination').val(coords);
+        });
+
+        // Plan a trip using information provided
+        $('section.directions button[type=submit]').click(function() {
```
CloudNiner: Consider breaking some of the anonymous functions here into named private functions. To keep `this` bound to the instance of Pages.Map, you can do:
```
$('selector').click($.proxy(foo, this));

function foo() { ... };
```
See http://api.jquery.com/jquery.proxy/

CloudNiner:
+1, although I would still update the 'am I at the target location' check to be a bit more lenient.

PR #33, opened 2015-01-26
Move views to appropriate apps

sharph:


ddohler:
+1 - I also like that this matches the structure of our existing projects better, i.e. the "main" app handles primarily configuration, not functionality.

PR #34, opened 2015-01-28
Feature/display routes

notthatbreezy:
This commit adds displaying of routes and
route summaries on the map page.

Route functionality was refactored
to separate itself from the map related
javascript files since the route/itinerary
stuff was mostly separated from the map.

Styling is still required for the directions;
however, mostly everything is separated into
classes and such so that shouldn't be too difficult


notthatbreezy:
screenshot from 2015-01-28 10 15 37

notthatbreezy:
This leaves the detail view for a separate task in the interest of moving things along. Clicking on a specific itinerary changes the styling.

python/cac_tripplanner/templates/map.html
```diff
@@ -59,6 +59,30 @@
                     </div>
                 </div>
             </div>
+            <div class="sidebar-clip">
```
notthatbreezy: Not sure why this is here, will remove

src/app/scripts/cac/routing/cac-routing-itinerary.js
```diff
@@ -0,0 +1,132 @@
...
+     * @param {object} otpItinerary OTP itinerary
+     *
+     * @return {string} string to use for labeling an itinerary
+     */
+    function getVia(otpItinerary) {
+        var steps = _(otpItinerary.legs).map(function(leg) {
+            return leg.steps;
+        }).flatten();
+
+        return steps.max(function(step) {
```
CloudNiner: _.max can be shorthanded to just take a string property name, e.g. `return steps.max('distance')`

src/app/scripts/cac/routing/cac-routing-plans.js
```diff
@@ -0,0 +1,109 @@
...
+    function planTrip(coordsFrom, coordsTo, map) {
+        var deferred = $.Deferred();
+        var urlParams = prepareParamString(coordsFrom, coordsTo);
+        var requestUrl = routingUrl + '?' + urlParams;
+        return $.ajax({
+            url: requestUrl,
+            type: 'GET',
+            contentType: 'application/json'
+        }).then(function(data) {
+            removeItineraryLayers(map);
```
CloudNiner: I think this is probably the wrong place for this map interaction. Do it on Pages.Map when this promise resolves, since not all planTrips() necessarily care about a map.

CloudNiner:
@notthatbreezy refactored a bit. See if you like this any better, if not we can revert and I'll continue reviewing.

src/app/scripts/cac/routing/cac-routing-itinerary.js
moradology: Alternatively, you could include max in the chain above it:
```javascript
return _(otpItinerary.legs)
  .map(function(leg){return leg.steps;})
  .flatten()
  .max(function(steps){return step.distance;})
  .streetName;
```

CloudNiner: Oh, map has the same shorthand, so it can even be
```
return _(otpItinerary.legs).map('steps').flatten().max('distance').streetName
```

moradology: Interesting. I assumed you were talking about a JS builtin - had no idea underscore/lodash supported this.

src/app/scripts/cac/cac.js
```diff
@@ -4,6 +4,7 @@ CAC = (function () {
     var module = {
         Map: {},
         Pages: {},
+        Routing: {},
```
moradology: :+1: for clean ontology.

moradology:
This looks good. +1 and I'll rebase to match the organizational decisions made here.

CloudNiner:
I didn't touch the css either, likely it was mostly copy pasted from the prototype.

notthatbreezy:
Looks great to me -- thanks for taking this when I was out sick!

PR #35, opened 2015-01-28
Add search destination API endpoint

CloudNiner:


python/cac_tripplanner/destinations/views.py
```diff
@@ -63,3 +64,35 @@ def get(self, request, *args, **kwargs):
...
+
+        """
+        params = request.GET
+        lat = params.get('lat', None)
+        lon = params.get('lon', None)
+        text = params.get('text', None)
+        output_fields = ('name', 'description', 'point', 'address', 'city', 'state', 'zip')
+
+        results = []
+        if lat is not None and lon is not None:
```
flibbertigibbet: If `lat` and `lon` are strings here, might as well just do:
```python
if lat and lon:
```
(will be truthy if non-null and non-empty).

python/cac_tripplanner/destinations/views.py
```diff
@@ -63,3 +64,35 @@ def get(self, request, *args, **kwargs):
...
+            searchPoint = Point(float(lon), float(lat))
```
flibbertigibbet: Might want to put some error handling around this, in case the cast to `float` or the point creation fails, and return a handy message about the coordinates being invalid.

python/cac_tripplanner/destinations/views.py
```diff
@@ -63,3 +64,35 @@ def get(self, request, *args, **kwargs):
...
+            results = (Destination.objects.filter(published=True)
+                .distance(searchPoint)
+                .order_by('distance')[:10])
```
flibbertigibbet: Not sure why there's slicing happening here and below, but it might be good to avoid if possible, or comment if not.  Is there a chance that the thing might be empty or smaller than the slice size?  Should check for that if so.

flibbertigibbet:
LGTM, though I'm not familiar with this corner of the codebase.

python/cac_tripplanner/destinations/views.py
CloudNiner: Just limits the number of results. I made it an optional parameter.

python/cac_tripplanner/destinations/views.py
CloudNiner: Just limits the number of results. I made it an optional parameter.

PR #37, opened 2015-02-23
Feature/deployment tweaks

CloudNiner:


Vagrantfile
```diff
@@ -111,7 +111,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     if testing?
         app.vm.synced_folder ".", "/opt/app"
     else
-        app.vm.synced_folder ".", "/opt/app", :nfs => true, :mount_options => ['actimeo=2']
+        app.vm.synced_folder ".", "/opt/app", :nfs => true, :mount_options => [
+            ("nfsvers=3" if ENV.fetch("CAC_NFS_VERSION", 4).to_i == 3),
```
sharph: Did you fix your nfs troubles with this?

sharph: Also, might something like CAC_USE_NFS_3 be better here? Seems weird that you are specifying a boolean with what appears to be any arbitrary version.

sharph:
+1

Vagrantfile
CloudNiner: Yikes, that's what i get for copy-pasting. I'll update to what you suggest.

Fix is from here: https://github.com/azavea/nih-wayfinding/pull/52

PR #40, opened 2015-02-24
Feature/home page destinations

CloudNiner:


deployment/ansible/roles/cac-tripplanner.app/templates/nginx-default.j2
```diff
@@ -4,11 +4,15 @@ server {
 	server_name _;
 
 	location / {
-                proxy_set_header Host $http_host;
-		proxy_pass http://127.0.0.1:8000;
+        proxy_set_header Host $http_host;
+        proxy_pass http://127.0.0.1:8000;
```
kshepard: Seems like the indentation in the file has gone awry.

kshepard:
+1

deployment/ansible/roles/cac-tripplanner.app/templates/nginx-default.j2
CloudNiner: Ugh. \t is the worst.

PR #41, opened 2015-02-24
Feature/copy all source data

flibbertigibbet:


deployment/ansible/roles/cac-tripplanner.otp-data/files/validate_feed.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/python
+
+# Check GTFS feeds have no errors and are current, using feedvalidator.py
```
CloudNiner: Did you write this script or did it come from somewhere? There are quite a things that don't match our style guidelines, like short var names, no docstrings. Either way, I think it's fine as is, but if it came from somewhere would probably be helpful to cite the source.

deployment/ansible/roles/cac-tripplanner.otp-data/defaults/main.yml
```diff
@@ -4,7 +4,5 @@
 #otp_data_dir: /var/otp
 otp_osm_source: https://s3.amazonaws.com/metro-extracts.mapzen.com/philadelphia_pennsylvania.osm.pbf
 otp_osm_filename: /var/otp/philadelphia.osm.pbf
-otp_gtfs_sources:
-  - septa_rail
-  - septa_bus
-  - patco
+otp_repo: "https://github.com/azavea/OpenTripPlanner.git"
```
CloudNiner: I'm not sure we should be using the azavea fork unless absolutely necessary.

`otp_version` can be a tag name, so we can just bring down v0.13 of the parent repository. If we do need to use the azavea fork in the future, we can trivially switch to using it then.

deployment/ansible/roles/cac-tripplanner.otp-data/files/validate_feed.py
flibbertigibbet: It's based on my [validator script for the PATCO GTFS](https://github.com/flibbertigibbet/patco-gtfs/blob/develop/validate_feed.py) that I use with Travis, so yes to both.  It's just a tiny script for a convenience check, so it doesn't seem to me to be something to fill with docstrings (we'll never build out docs for it.)

deployment/ansible/roles/cac-tripplanner.otp-data/defaults/main.yml
flibbertigibbet: I was thinking we might want to create a CAC-specific branch on our fork for modifications, but as that hasn't proven necessary yet, I'll set it back to use the upstream branch/tag.

CloudNiner:
nice +1

PR #44, opened 2015-02-28
Minor styling/markup edits based on feedback from meeting

mtedeschi:
There isn't a lot here yet, but once we hear from the partner that's doing branding work, I can hopefully jump in and start doing styling and refining things. Let me know if you need specific stuff now that this is working!


flibbertigibbet:
:+1:

PR #46, opened 2015-03-02
Feature/fix python tests

flibbertigibbet:
Python linting; build OTP on Jenkins, to fix failing Python test that expects OTP endpoint to be live.


kshepard:
+1

PR #47, opened 2015-03-02
Feature/add js tests

flibbertigibbet:
Adds some basic JavaScript tests, and comments out the python test that requires an OTP response (to re-enable when OTP files are loaded onto Jenkins).

### DIFF
--- a/src/karma/karma.conf.js
+++ b/src/karma/karma.conf.js
@@ -18,7 +18,8 @@ module.exports = function(config) {
       '/srv/cac/scripts/jquery.js',
       '/srv/cac/scripts/vendor.js',
       '/srv/cac/scripts/main.js',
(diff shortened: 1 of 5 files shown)


src/karma/karma.conf.js
```diff
@@ -18,7 +18,8 @@ module.exports = function(config) {
       '/srv/cac/scripts/jquery.js',
       '/srv/cac/scripts/vendor.js',
       '/srv/cac/scripts/main.js',
-      'test/spec/test.js'
+      'test/spec/*.js',
+      'test/spec/**/*.js',
```
kshepard: There's an extra trailing comma here, as well as in the karma-dev conf.

kshepard:
+1

PR #48, opened 2015-03-03
Download OTP data files from S3 on test server

flibbertigibbet:
Also re-enables python test that requires OTP to be live.

### DIFF
--- a/deployment/ansible/roles/cac-tripplanner.otp-data/tasks/main.yml
+++ b/deployment/ansible/roles/cac-tripplanner.otp-data/tasks/main.yml
@@ -5,6 +5,11 @@
 - name: Copy Feed Validator
   copy: src=validate_feed.py dest="{{ otp_data_dir }}"
 
+- name: Download OTP Data
+  local_action: command aws s3 sync s3://cleanair-otp-data "{{ otp_data_dir }}/"
+  sudo: no
+  when: production or test
+
 - name: Copy OTP Data
   copy: src=./otp_data/ dest="{{ otp_data_dir }}/" owner={{ansible_ssh_user}} group={{ansible_ssh_user}} mode=0664
   notify: Validate GTFS
--- a/deployment/ansible/group_vars/all
+++ b/deployment/ansible/group_vars/all
@@ -5,4 +5,6 @@ app_username: "vagrant"
 # open trip planner vars
 otp_process_mem: 4G
 
+s3_otp_data: cleanair-otp-data
+
 postgresql_support_libpq_version: 9.4.*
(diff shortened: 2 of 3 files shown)


deployment/ansible/roles/cac-tripplanner.otp-data/tasks/main.yml
```diff
@@ -5,6 +5,11 @@
 - name: Copy Feed Validator
   copy: src=validate_feed.py dest="{{ otp_data_dir }}"
 
+- name: Download OTP Data
+  command: aws s3 sync s3://cleanair-otp-data "{{ otp_data_dir }}/"
+  sudo: no
+  when: production or test
```
kshepard: What's the reason for not also downloading the files in development? Seems like they may come in handy, at least for running local tests.

kshepard:
+1

notthatbreezy:
Might want to use the s3 module for ansible

deployment/ansible/roles/cac-tripplanner.otp-data/tasks/main.yml
flibbertigibbet: It would require setting up the AWS account credentials locally (the files aren't public), and might interfere with testing different input files locally.  If neither of those things are a concern, though, then sure, we could run this everywhere.

flibbertigibbet:
Neat, I didn't know there was an ansible module for that.  Looking at [the docs](http://docs.ansible.com/s3_module.html), I don't see a way to download all the contents of a bucket, just to download individual files by name.

Do you know if the ansible module provides an equivalent to `sync` @notthatbreezy?

deployment/ansible/roles/cac-tripplanner.otp-data/tasks/main.yml
kshepard: Oh, ok, I was assuming the files were public.

notthatbreezy:
That's a good point - I don't see a way to do sync. I think eventually we might want to individually specify GTFS data (to keep things updated), though maybe not so using the s3 command is probably good enough.

Is the `awscli` getting installed as a dependency somewhere else? I couldn't find it in any of the existing roles.

flibbertigibbet:
The idea was that this was for the test server, so as @ddohler suggested in chat, I set up the Jenkins server with `awscli` and the credentials.  So, neither the credentials or `awscli` package are set up from the repo.

flibbertigibbet:
I want the OTP data directory to not list specific files because they will likely change in future, and I want the OTP data role to not concern itself with what gets used.  We will likely add elevation data in future, and may throw in more or different GTFS files.  It's also convenient to not depend on exact file names for the GTFS we have.

notthatbreezy:
Ah ok - might not be a bad idea to add in a specific task to install it -- that way it covers when/if the test vagrant environment gets destroyed and needs rebuilt. Up to you though.

flibbertigibbet:
Oh, I guess I need to invoke the command with `local_action` then, so ansible will run the sync from the host (the AWS setup is on the Jenkins host, not the vagrant box).

flibbertigibbet:
Ok, I'm going to go ahead and merge this, to see what Jenkins makes of it.

PR #55, opened 2015-03-03
Make VM memory usage configurable via environment variables.

flibbertigibbet:
Also reduce memory used by OTP graph build.

(Jenkins host machine doesn't have enough memory to support the existing defaults.)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -6,9 +6,9 @@ require "yaml"
 
 if ENV['CAC_TRIPPLANNER_MEMORY'].nil?
   # OpenTripPlanner needs > 1GB to build and run
-  MEMORY_MB = "8192"
+  OTP_MEMORY_MB = "4096"
 else
-  MEMORY_MB = ENV['CAC_TRIPPLANNER_MEMORY']
+  OTP_MEMORY_MB = ENV['CAC_TRIPPLANNER_MEMORY']
 end
 
 if ENV['CAC_TRIPPLANNER_CPU'].nil?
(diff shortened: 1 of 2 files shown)


deployment/ansible/group_vars/all
```diff
@@ -3,7 +3,7 @@ postgresql_support_libpq_version: "9.4.0-1.pgdg14.04+1"
 app_username: "vagrant"
 
 # open trip planner vars
-otp_process_mem: 4G
+otp_process_mem: 2G
```
hectcastro: Potentially useful, although you may not care to do it here – Ansible has a [`lookup`](http://docs.ansible.com/playbooks_lookups.html#more-lookups) function that can pull from the environment where Ansible is being executed.

flibbertigibbet: Neat, good to know.  However much we need to do the build should be constant, but I'll keep that in mind for future reference.

Vagrantfile
```diff
@@ -133,8 +139,14 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
 
+    if ENV['CAC_APP_MEMORY'].nil?
+      APP_MEMORY_MB = "2048"
+    else
+      APP_MEMORY_MB = ENV['CAC_APP_MEMORY']
+    end
+
     config.vm.provider :virtualbox do |v|
-      v.memory = 2048
+      v.memory = APP_MEMORY_DB
```
notthatbreezy: Should be `APP_MEMORY_MB`

notthatbreezy:
+1 after fixing the typo, verified that provisions successfully

PR #56, opened 2015-03-03
Fix references to `this` broken by linting

flibbertigibbet:
Remaining references to `self` are necessary to avoid linter errors; this reverts those references that broke functionality (and also happened to not be linter errors).

### DIFF
--- a/src/app/scripts/cac/search/cac-typeahead.js
+++ b/src/app/scripts/cac/search/cac-typeahead.js
@@ -50,13 +50,14 @@ CAC.Search.Typeahead = (function ($) {
     return CACTypeahead;
 
     function onTypeaheadSelected(event, suggestion, dataset) {
+        var self = this;
         var typeaheadKey = $(event.currentTarget).data('typeahead-key') || defaultTypeaheadKey;
 
         if (dataset === 'currentlocation') {
-            this.events.trigger('cac:typeahead:selected', [typeaheadKey, suggestion]);
+            self.events.trigger('cac:typeahead:selected', [typeaheadKey, suggestion]);
         } else {
             CAC.Search.Geocoder.search(suggestion.text, suggestion.magicKey).then(function (location) {
-                this.events.trigger('cac:typeahead:selected', [typeaheadKey, location]);
+                self.events.trigger('cac:typeahead:selected', [typeaheadKey, location]);
             });
         }
     }
--- a/src/app/scripts/cac/search/cac-typeahead.js
+++ b/src/app/scripts/cac/search/cac-typeahead.js
@@ -24,7 +24,6 @@ CAC.Search.Typeahead = (function ($) {
         minLength: 2
     };
     var defaultTypeaheadKey = 'default';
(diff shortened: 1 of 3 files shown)


src/app/scripts/cac/search/cac-typeahead.js
```diff
@@ -57,7 +57,7 @@ CAC.Search.Typeahead = (function ($) {
             self.events.trigger('cac:typeahead:selected', [typeaheadKey, suggestion]);
         } else {
             CAC.Search.Geocoder.search(suggestion.text, suggestion.magicKey).then(function (location) {
-                self.events.trigger('cac:typeahead:selected', [typeaheadKey, location]);
+                this.events.trigger('cac:typeahead:selected', [typeaheadKey, location]);
```
notthatbreezy: Still get an error here when trying to geocode

```
Uncaught TypeError: Cannot read property 'trigger' of undefined cac-typahead.js:60
```

flibbertigibbet:
Okay, all `this` references have been restored; defining `self` apparently does not work in all contexts.  JSHint will now ignore the strict check for `this` references.

src/app/scripts/cac/search/cac-typeahead.js
```diff
@@ -54,10 +53,10 @@ CAC.Search.Typeahead = (function ($) {
         var typeaheadKey = $(event.currentTarget).data('typeahead-key') || defaultTypeaheadKey;
 
         if (dataset === 'currentlocation') {
-            self.events.trigger('cac:typeahead:selected', [typeaheadKey, suggestion]);
+            this.events.trigger('cac:typeahead:selected', [typeaheadKey, suggestion]);
         } else {
             CAC.Search.Geocoder.search(suggestion.text, suggestion.magicKey).then(function (location) {
-                self.events.trigger('cac:typeahead:selected', [typeaheadKey, location]);
+                this.events.trigger('cac:typeahead:selected', [typeaheadKey, location]);
```
notthatbreezy: This (no pun intended) still throws an error -- the whole function needs to be reverted so it should be

```javascript
    function onTypeaheadSelected(event, suggestion, dataset) {
        var self = this;
        var typeaheadKey = $(event.currentTarget).data('typeahead-key') || defaultTypeaheadKey;

        if (dataset === 'currentlocation') {
            self.events.trigger('cac:typeahead:selected', [typeaheadKey, suggestion]);
        } else {
            CAC.Search.Geocoder.search(suggestion.text, suggestion.magicKey).then(function (location) {
                self.events.trigger('cac:typeahead:selected', [typeaheadKey, location]);
            });
        }
    }
```

flibbertigibbet:
I saw no console errors in testing typeahead; reverted function anyways.

notthatbreezy:
You have to hit geocode, there are no errors just from the typahead.

flibbertigibbet:
Okay... so are you fine with this now?

notthatbreezy:
+1

PR #57, opened 2015-03-04
Add build tasks to gulp-test

flibbertigibbet:


flibbertigibbet:
Trivial, merging.

src/gulpfile.js
```diff
@@ -160,7 +160,10 @@ gulp.task('common:build', ['clean'], function() {
     return gulp.start('copy:vendor-css', 'copy:vendor-images', 'sass', 'collectstatic');
 });
 
-gulp.task('test', sequence(['jshint',
+gulp.task('test', sequence([
+            'development',
+            'production',
```
notthatbreezy: Does having both development and production together make sense? They both run clean so production is just deleting everything that development produced.

flibbertigibbet: The [app install task](https://github.com/azavea/cac-tripplanner/blob/develop/deployment/ansible/roles/cac-tripplanner.app/tasks/jslibs.yml) already runs `gulp-production`, so it's unclear why gulp hasn't been finding all its files on Jenkins.

notthatbreezy: I'm not sure what issues are occurring on Jenkins, just pointing out that running development here isn't going to do anything because production is just going to delete everything that got copied over during the development task.

flibbertigibbet: I thought perhaps it was copying something that wasn't getting deleted, and was unsure why Jenkins was having issues that weren't appearing locally.  Merged so Jenkins could run it.  This has been reverted in #58, which I tested manually on the Jenkins VM.

PR #60, opened 2015-03-04
Fix isochrone fetch.

flibbertigibbet:
Return isochrone along with list of matched locations from Django endpoint.  Display isochrone when search button clicked in 'Discover Places' tab on map page.


src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -90,7 +95,8 @@ CAC.Map.Control = (function ($, L, _) {
             } else {
                 userMarker = new L.CircleMarker(latlng)
                   .on('click', function() {
-                      events.trigger('MOS.Map.Control.CurrentLocationClicked', latlng);
+                      // TODO: not implemented
+                      events.trigger('CAC.Map.Control.CurrentLocationClicked', latlng);
```
kshepard: Ha, good catch.

src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -101,7 +107,7 @@ CAC.Map.Control = (function ($, L, _) {
             deferred.fail(function(){return 'ERROR(' + error.code + '): ' + error.message; });
         };
 
-        if ('geolocation' in navigator) {
+        if (navigator.geolocation) {
```
kshepard: This looks cleaner, but it causes problems in various browsers. See the "First things first" section here: http://themarklee.com/2013/10/27/super-simple-geolocation/

flibbertigibbet: Good to know.  I'd pulled that from [the previous working branch](e3bbebb41e019ff0567ea23d9e7d1b652d151dce) and assumed it was necessary.  I'll revert that.

kshepard:
Since cac-map-isochrone.js has been removed, the include can also be removed from the html template.

flibbertigibbet:
Good catch.  I was looking around for includes, but didn't find any in the JavaScript directory.  I see, it's over in the Django template.

src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -110,17 +116,103 @@ CAC.Map.Control = (function ($, L, _) {
...
+            });
+        };
+
+        // default to current date and time
+        var now = new Date();
+        // months are zero-based
+        var dateStr = [(now.getMonth() + 1),
+                        now.getDate(),
+                        now.getFullYear()
+                      ].join('-');
```
kshepard: I feel like this should be easier to accomplish. Maybe we should start using something like Moment.js at some point.

kshepard:
+1

src/app/scripts/cac/map/cac-map-control.js
flibbertigibbet: Yes, I'm thinking this will get moved and significantly modified when we implement user preferences and options.  Right now, there's no way to set the time for any searches, which I think will be pretty important.

PR #61, opened 2015-03-04
Feature/fix reachable tests

flibbertigibbet:
We will probably change the image field to be required, when we get around to requiring three image fields of different sizes.


flibbertigibbet:
Simple fix, merging.

PR #63, opened 2015-03-05
Add geocoder support to admin interface

kshepard:


kshepard:
Screenshot:

admin-geocoding

python/cac_tripplanner/templates/admin/cac-geocoding-map.js
```diff
@@ -0,0 +1,22 @@
+{% extends 'gis/admin/osm.js' %}
+
+{% block extra_layers %}
+
+var $address = $('[name=address]');
+var module = {{ module }};
+var map = module.map;
+var projLatLng = new OpenLayers.Projection('EPSG:4326');
+var projWM = new OpenLayers.Projection('EPSG:900913'); // 3857 doesn't work here for some reason
```
flibbertigibbet: Looks like it may be necessary to [declare an alias](http://docs.openlayers.org/library/spherical_mercator.html#sphericalmercator-and-epsg-aliases) for 3857 in OpenLayers.  (I'm fine with the Google code here.)

flibbertigibbet:
+1

python/cac_tripplanner/destinations/admin.py
```diff
@@ -11,6 +11,17 @@ class DestinationAdmin(admin.OSMGeoAdmin):
...
 
+    # Override map_template for custom address geocoding behavior
+    map_template = 'admin/cac-geocoding-map.html'
+
+    # Include geocoder dependencies
+    extra_js = [
+        'https://code.jquery.com/jquery-2.1.3.min.js',
+        '/static/scripts/vendor/lodash.js',
+        '/static/scripts/main/cac/cac.js',
+        '/static/scripts/main/cac/search/cac-geocoder.js'
```
CloudNiner: Nice! Glad this was reusable.

python/cac_tripplanner/templates/admin/cac-geocoding-map.js
kshepard: Cool, nice find! I'll update the comment with the reason, but will keep the code as-is, since finding the right place in the template to add the override will definitely be more trouble than it's worth.

python/cac_tripplanner/destinations/admin.py
kshepard: Me too, it worked out surprisingly well.

PR #64, opened 2015-03-05
Feature/parse rss

CloudNiner:


CloudNiner:
Any opposition to just using a cron job to run this, something like once an hour or every other hour?

python/cac_tripplanner/destinations/management/commands/load_events.py
```diff
@@ -0,0 +1,93 @@
...
+            return None
+        return elements[0].firstChild.data
+
+    def handle(self, *args, **options):
+        """ Retrieve and populate FeedEvent table from an RSS Feed """
+
+        url = 'http://www.uwishunu.com/category/events/feed/'
+        # Get 403 forbidden without changing user-agent
+        headers = {'User-Agent': 'Mozilla/5.0 (X11; Linux i686; rv:10.0) Gecko/20100101 Firefox/10.0'}
+        request = urllib2.Request(url, None, headers)
```
kshepard: It'd be good to have some error handling here, with problems being written to a log file in order to make it easy to diagnose failures.

kshepard:
Cron seems fine to me for now. We'll just need to move it off the webserver if we end up needing to scale up.

notthatbreezy:
Yeah - I think we have to solve the same problem on some other projects, so once this gets deployed I think we'll have a solution that will allow us to make sure the task is only done once. The easiest solution that would get this job done is to query the EC2 API, sort the EC2 instances in a repeatable fashion, then have the cron job bail out if it isn't the first instance in the sorted EC2 list.

kshepard:
+1, I wasn't able to test it out due to some of my local db changes, but looks good.

CloudNiner:
@kshepard added a basic cron job via ansible, might want to take a look

python/cac_tripplanner/destinations/management/commands/load_events.py
CloudNiner: Created the log file as part of the cron job, from what I've seen the urllib and xml stack traces that would be printed are actually fairly useful in terms of diagnosing issues. Not sure additional handling is worth it in such a simple function?

kshepard: Sounds good. I wrote that comment before the log file existed, but now that it does and the stack traces are being written to it, that's fine.

kshepard:
:+1:

deployment/ansible/roles/cac-tripplanner.app/tasks/main.yml
```diff
@@ -50,6 +50,17 @@
...
   notify: Restart nginx
 
+- name: Touch cron job log file if it does not exist
+  command: touch {{ app_cron_event_feed_log }}
+           creates={{ app_cron_event_feed_log }}
+
+- name: Set cron job log file permissions
+  file: path={{ app_cron_event_feed_log }} owner={{ app_username }} group={{ app_username }} mode=0664
+
+# TODO: Add logic in production to only run on a single webserver
```
flibbertigibbet: As I recall, we put a bunch of points on this card.  Would be good to go ahead and do this now.

CloudNiner: There are two other projects that require the exact same functionality, and there is time devoted to solving the problem for those. Going to hold off on working on that, and instead write the API endpoint for these events.

PR #65, opened 2015-03-05
Feature/feed event api

CloudNiner:


python/cac_tripplanner/destinations/views.py
```diff
@@ -130,3 +133,24 @@ def get(self, request, *args, **kwargs):
...
         data = serializers.serialize('json', results, fields=output_fields)
         return HttpResponse(data, 'application/json')
+
+class FeedEvents(View):
+    """ API endpoint for the FeedEvent model """
+
+    def get(self, request, *args, **kwargs):
+        """ GET 20 most recent feed events that are published
+
+        TODO: Additional filtering, dynamic limits?
```
CloudNiner: I can add more params here, but I'm not sure ATM what is actually necessary, since I think the display of these events on the map is pretty basic (i.e. just show the clickable dots on a map with popup)

kshepard: I think we may have differing thoughts on what is supposed to happen with these feed events. I was under the impression that they'd show up in the search results alongside the featured locations, and then they'd be selectable as a routing option. Would be good to have @flibbertigibbet and/or @jbranigan weigh in to make sure we implement this correctly.

flibbertigibbet: Yes, that was my impression as well, that they'd go into the featured locations and be used for routing.  We'll also want to remove them once event dates pass, I imagine.

CloudNiner: I spoke with @jbranigan yesterday and he said that these FeedEvents would be added to the map as an additional toggleable marker layer that defaults to on. They are not to be shown in the featured destinations section of the homepage. I didn't ask whether they should show up in the search results.

CloudNiner: I also limited to 20 events because there is no end date in the feed. Only the publish date. The event date is occasionally buried in the content, which would be way too much effort to parse I think.

python/cac_tripplanner/destinations/management/commands/load_events.py
```diff
@@ -90,4 +91,4 @@ def handle_feed_item(self, item):
 
         feed_event, created = FeedEvent.objects.update_or_create(guid=guid, defaults=updated_item)
         if created:
-            self.stdout.write('Added event with guid: "%s"' % guid)
+            self.stdout.write('%s: Added event with guid: "%s"' % (str(datetime.utcnow()), guid))
```
kshepard: I think I missed this before, but we should use the new-style string interpolation for these.

kshepard:
+1, this looks like a good start. I imagine it'll need to be modified a bit as we figure out exactly how these events will be integrated into the UI.

python/cac_tripplanner/destinations/management/commands/load_events.py
CloudNiner: Derp. Updated.

PR #66, opened 2015-03-05
Finish featured locations admin input

kshepard:


flibbertigibbet:
Are we not going to do any check on the image fields to see if they actually match the dimensions?

kshepard:
@flibbertigibbet: that's a good idea, I can add validation for that.

python/cac_tripplanner/destinations/models.py
```diff
@@ -25,7 +26,10 @@ class Destination(models.Model):
                                help_text=('The map automatically updates as the address is typed, '
                                           'but may be overridden manually if incorrect.'))
     point = models.PointField()
-    image = models.ImageField(upload_to='destinations/', null=True)
+    image = models.ImageField(upload_to='destinations/', null=True,
```
notthatbreezy: Now that we have an AWS account set up, should we create a bucket to upload images to or is that a separate task (I'm not sure what's involved with that)

kshepard: That also sounds like a good idea. I'll look into that.

flibbertigibbet:
I imagine the python tests will need updating for the changed model.

kshepard:
Added image validation and s3 storage (configurable). This could use another review.

python/cac_tripplanner/destinations/migrations/0007_destination_website_url.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('destinations', '0006_add_city_state_defaults'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='destination',
+            name='website_url',
+            field=models.URLField(default='http://github.com/azavea/cac-tripplanner'),
```
flibbertigibbet: Might be better to show an error to the user if they haven't entered a link, or make the field not be required; we probably don't want this link to make it into production use.

python/cac_tripplanner/destinations/views.py
```diff
@@ -69,13 +79,9 @@ def get(self, request, *args, **kwargs):
...
-            pnt = obj['point']
-            obj['point'] = json.loads(pnt.json)
-            img = obj.get('image')
-            if img:
-                obj['image'] = img.url
-            else:
-                obj.pop('image')
+            obj['point'] = json.loads(obj['point'].json)
+            obj['image'] = self.image_to_url(obj, 'image')
+            obj['wide_image'] = self.image_to_url(obj, 'wide_image')
```
flibbertigibbet: Will this work ok if an image field is null?  (Looks like they're both still nullable.)

flibbertigibbet:
Image dimensions check looks nice.  Could we also do a file size check?  Seem to recall I managed to get an nginx error on attempting to upload a GIF.  An allowed extension message/check might be good, too.  Another thought:  we might want to run an image minimizer on the uploads.

flibbertigibbet:
Apart from above considerations, +1

python/cac_tripplanner/destinations/migrations/0007_destination_website_url.py
kshepard: Ok, I can make it an optional field.

python/cac_tripplanner/destinations/views.py
kshepard: Yep, it works fine with nulls: image_to_url has a null check.

kshepard:
An image size check sounds like a good idea, I'll add that. I probably don't have time to add an image minimizer at the moment, but I'll add an issue for it so we can investigate it in the future.

PR #67, opened 2015-03-06
Feature/wire homepage buttons

flibbertigibbet:
Saves form data from the home page in local storage (or cookie, if local storage not supported) and redirects to map page when 'go' buttons clicked.
Map page loads preferences, if present, then initiates appropriate directions or 'explore' search.
Modifies travel mode drop-downs; do not use multi-select, instead have unique options (some selections imply or exclude others, so this is simpler).

Map page still needs some work; it's not reading several of the options in the UI for its queries, and does not update stored preferences, just reads them.

Note:  attempted to add tests for map page, but test runner was very unhappy around leaflet initialization.


notthatbreezy:
+1 - was unable to load the isochrone after hitting "Get Directions" - django is returning 500s -- not sure if that's part of this task or not

flibbertigibbet:
Isochrones are working fine for me; I imagine something is up with your VM.  Also, going to 'Get Directions' hits OTP for directions, not the Django isochrone endpoint.

notthatbreezy:
Hmm ok - I provisioned all of the VMs successfully from your branch. It looks like it is making a request here `http://localhost:8024/map/reachable?coords%5Blat%5D=39.9638352&coords%5Blng%5D=-75.1603638&mode%5B%5D=WALK&mode%5B%5D=TRANSIT&date=3-6-2015&time=13%3A30%3A58+GMT-0500+(EST)&maxTravelTime=1800&maxWalkDistance=1609`, then a 500 error gets thrown when it tries to grab the first item in a list `json_poly = json.loads(isochrone_response.content)[0]['geometry']['geometries']`

flibbertigibbet:
Ok.  I see, you have no features.  The previously added endpoint doesn't check for empty result.

flibbertigibbet:
Will modify isochrone endpoint separately.

PR #70, opened 2015-03-06
Add check and test for empty isochrone.

flibbertigibbet:
Unlikely to occur, but may happen if origin outside graph bounds, or inaccessible.

See #67 comments.

### DIFF
--- a/python/cac_tripplanner/destinations/views.py
+++ b/python/cac_tripplanner/destinations/views.py
@@ -45,7 +45,11 @@ class FindReachableDestinations(View):
         isochrone_response = requests.get(self.isochrone_url, params=payload)
 
         # Parse and traverse JSON from OTP so that we return only geometries
-        json_poly = json.loads(isochrone_response.content)[0]['geometry']['geometries']
+        try:
+            json_poly = json.loads(isochrone_response.content)[0]['geometry']['geometries']
+        except:
+            # No isochrone found.  Is GTFS loaded?  Is origin within the graph bounds?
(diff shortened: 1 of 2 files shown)


python/cac_tripplanner/destinations/views.py
```diff
@@ -45,7 +45,11 @@ def isochrone(self, lat, lng, mode, date, time, max_travel_time, max_walk_distan
         isochrone_response = requests.get(self.isochrone_url, params=payload)
 
         # Parse and traverse JSON from OTP so that we return only geometries
-        json_poly = json.loads(isochrone_response.content)[0]['geometry']['geometries']
+        try:
+            json_poly = json.loads(isochrone_response.content)[0]['geometry']['geometries']
```
notthatbreezy: Requests has a nice helper method you can use to save the call to json.loads, `isochrone_response.json()`

notthatbreezy:
+1

python/cac_tripplanner/destinations/views.py
flibbertigibbet: Oh, we have requests in the project.  I saw the feed parser was just using urllib.

notthatbreezy: I haven't looked at the feed parser, but at least here we're using requests https://github.com/azavea/cac-tripplanner/blob/2a90ebd68885efc0549988167223a2c0a45bb6a7/python/cac_tripplanner/destinations/views.py#L45

PR #71, opened 2015-03-06
Add marker on successful discover places location search

CloudNiner:
Has the following behavior:
  - map pans to marker center when displayed
  - marker is cleared when the discover places tab is hidden


flibbertigibbet:
+1

PR #74, opened 2015-03-06
Feature/add time selector

flibbertigibbet:
Adds extra options section to directions tab.
Adds mode and time selectors to directions, and uses them in planning trips.

### DIFF
--- a/src/app/scripts/cac/pages/cac-pages-map.js
+++ b/src/app/scripts/cac/pages/cac-pages-map.js
@@ -36,6 +36,9 @@ CAC.Pages.Map = (function ($, Handlebars, _, MapControl, Routing, MockDestinatio
 
         $('.sidebar-options .view-more').click($.proxy(showOptions, this));
 
+        // initiallize date/time picker
+        $('#datetimeDirections').datetimepicker();
+
         $('#sidebar-toggle-directions').on('click', function(){
             $('.explore').addClass('hidden');
             $('.directions').removeClass('hidden');
--- a/src/app/scripts/cac/pages/cac-pages-map.js
+++ b/src/app/scripts/cac/pages/cac-pages-map.js
@@ -159,6 +159,9 @@ CAC.Pages.Map = (function ($, Handlebars, _, MapControl, Routing, MockDestinatio
         if (!method) {
             return; // no user preferences set
         }
+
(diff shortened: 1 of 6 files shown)


flibbertigibbet:
image

src/app/scripts/cac/pages/cac-pages-map.js
```diff
@@ -57,8 +61,9 @@ CAC.Pages.Map = (function ($, Handlebars, _, MapControl, Routing, MockDestinatio
 
     return Map;
 
-    function showOptions() {
-        var moreOpt = '.sidebar-options .more-options';
+    function showOptions(event) {
+        var parent = $(event.target).parent().parent();
```
CloudNiner: I tend to strongly avoid selections like this, as it is extremely brittle, and I've been bitten by it before. At least with an id/class selector of some kind, you avoid breaking the selection if you move the dom tree around.

src/app/scripts/cac/pages/cac-pages-map.js
```diff
@@ -57,8 +61,9 @@ CAC.Pages.Map = (function ($, Handlebars, _, MapControl, Routing, MockDestinatio
...
+        var moreOpt = $('.sidebar-options .more-options', parent);
 
         $(moreOpt).toggleClass('active');
         $(moreOpt).parent().find('a.view-more').text(function() {
```
CloudNiner: Hmm an alternative to these parent selectors would be to select some wrapper element that is higher up the tree, and select the relevant children of it with standard selectors.

CloudNiner:
+1, but strongly consider switching up those selectors

flibbertigibbet:
image

src/app/scripts/cac/pages/cac-pages-map.js
flibbertigibbet: Yes, but how else do I find it relative to the clicked link?  It was toggling both when either tab's expansion was clicked, and I couldn't figure out why `proxy` didn't seem to switch the context.

flibbertigibbet:
A thousand sorries, but I'm going to go ahead and merge this is with the icky `parent().parent()` selector.  I have to go now, and will be out for a while, or else I'd spend more time on it.

Issue #77, opened 2015-03-10
Fix tests


kshepard:
Both the python and js tests are currently failing.

flibbertigibbet:
Fixed by #84 and #85.

PR #81, opened 2015-03-10
Wire up feature directions clicks on explore page

kshepard:
This makes it so the featured destinations that show up in the
left-panel of the explore page can be clicked, which changes
to the directions tab and routes to that destination.

Note: I wasn't able to write a test for this, because the test
suite is currently broken per issue #77.

### DIFF
--- a/src/app/scripts/cac/pages/cac-pages-map.js
+++ b/src/app/scripts/cac/pages/cac-pages-map.js
@@ -154,8 +154,26 @@ CAC.Pages.Map = (function ($, Handlebars, _, moment, MapControl, Routing, MockDe
         var $container = $('<div></div>').addClass('destinations');
         $.each(destinations, function (i, destination) {
             var $destination = $(CAC.Map.Templates.destinationBlock(destination));
+
             $destination.click(function () {
-                // TODO: What to do on click?!?!
+                // TODO: see issue #78 regarding refactors to improve this
+
+                showDirectionsTab();
+
+                // Set origin
+                var from = UserPreferences.getPreference('origin');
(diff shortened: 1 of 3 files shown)


src/app/scripts/cac/pages/cac-pages-map.js
```diff
@@ -154,8 +154,26 @@ CAC.Pages.Map = (function ($, Handlebars, _, moment, MapControl, Routing, MockDe
         var $container = $('<div></div>').addClass('destinations');
         $.each(destinations, function (i, destination) {
             var $destination = $(CAC.Map.Templates.destinationBlock(destination));
-            $destination.click(function (event) {
-                // TODO: What to do on click?!?!
+
+            $destination.click(function () {
```
CloudNiner: I didn't know about it when I wrote the original TODO here, but there is an intermediate view for the sidebar where on click we show the title, image and full description for a destination. The mockup is in the wireframes. 

Either way, we'll need this code because there is a button on that intermediate view that is for getting directions.

flibbertigibbet: The unused `event` was throwing a linter warning; I suspect why it was removed here.  (I've already removed it in #85).

flibbertigibbet: The unused `event` was throwing a linter warning; I imagine that's why it was removed here (I've already removed it in #85).

kshepard:
I just rebased/resolved conflicts, and am going to merge this before it gets too stale. As @CloudNiner mentioned, once the intermediate view is in place, this click event will need to be shifted over to the button there.

PR #83, opened 2015-03-16
Fix destination.website_url migration and model inconsistency

CloudNiner:
Migrations failed when updating to current develop wtih error:

`django.db.utils.IntegrityError: column "website_url" contains null values`

Made the assumption that the model definition was the correct one.


kshepard:
+1, oops -- totally my fault for modifying the model after the migration was in place.

PR #85, opened 2015-03-16
Feature/fix build errors

flibbertigibbet:
Fix JavaScript tests and linter warning.


flibbertigibbet:
As a note, it's a mystery to me why the project is set up to load tests with the mocha test framework, but runs them via the jasmine test framework.

CloudNiner:
LGTM

CloudNiner:
Yeah, I'm not sure why that is either. Can we just replace jasmine with mocha in karma.conf.js? Since we call `mocha.run` in the `test/index.html`?

flibbertigibbet:
No, because our tests are written using jasmine-style assertions.

PR #91, opened 2015-03-16
Remove tuple from secrets dictionary

### LINKED ISSUE #75: Secrets file cannot be successfully parsed
There is a tuple in the secrets dictionary which cannot be successfully represented as YAML, so the parser fails.

kshepard:
Fixes #75


flibbertigibbet:
:+1:

PR #92, opened 2015-03-16
Feature/save options on map

### LINKED ISSUE #79: Use UserPreferences consistently
There are currently some bugs related to not using UserPreferences consistently. For example, the marker shown for the origin on the explore page is set via searching on the home page, but changing the location from the explore page and searching does not update this location.

flibbertigibbet:
* Closes #79.
* Uses isochrone inputs from UI
* Fix python linter warnings
* Save empty JS linter file, if no warnings found (so Jenkins knows).

### DIFF
--- a/src/app/scripts/cac/map/cac-map-control.js
+++ b/src/app/scripts/cac/map/cac-map-control.js
@@ -145,6 +145,7 @@ CAC.Map.Control = (function ($, L, _) {
             $.ajax({
                 type: 'GET',
                 data: payload,
+                cache: false,
                 url: isochroneUrl,
                 contentType: 'application/json'
             }).then(deferred.resolve);
@@ -157,7 +158,7 @@ CAC.Map.Control = (function ($, L, _) {
     /**
      * Get travelshed and destinations within it, then display results on map.
     */
-    function fetchIsochrone() {
+    function fetchIsochrone(coordsOrigin, when, mode, exploreMinutes) {
         var deferred = $.Deferred();
         // clear results of last search
         clearDiscoverPlaces();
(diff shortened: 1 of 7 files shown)


src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -174,37 +175,37 @@ CAC.Map.Control = (function ($, L, _) {
...
-            params.coords.lng = data[1];
-            getIsochrone(params);
-        }, function(error) {
-            console.log('Could not geolocate user');
-            console.error(error);
-            // use default location
-            getIsochrone(params);
-        });
+        if (coordsOrigin) {
+            deferred.resolve(getIsochrone(params));
```
kshepard: I think this deferred.resolve should be removed (same as how the getIsochrone calls in the next block are called).

src/app/scripts/cac/pages/cac-pages-map.js
```diff
@@ -179,10 +190,39 @@ CAC.Pages.Map = (function ($, Handlebars, _, moment, MapControl, Routing, MockDe
...
+    function clickedExplore() {
+        var exploreMinutes = $('#exploreTime').val();
+        var mode = $('#exploreModeSelector').val();
+
+        // TODO: add date/time selector to 'explore' extra options panel?
+        var when = moment();
+
+        // store search inputs to preferences
+        UserPreferences.setPreference('method', 'explore');
+        UserPreferences.setPreference('originText', $('#exploreOrigin').val());
```
kshepard: The originText is being set, but it looks like the actual origin coordinates aren't. This would lead to a scenario where when the page is reloaded, the text will be set, but it will be using a different set of stored coordinates (probably from when it was stored on the home page).

kshepard:
+1 aside from my other comments.

src/app/scripts/cac/pages/cac-pages-map.js
flibbertigibbet: This is set now from `onTypeaheadSelected`, when the full location object is accessible.

src/app/scripts/cac/map/cac-map-control.js
flibbertigibbet: Good catch; missed that one.

PR #93, opened 2015-03-16
Add image upload fields for articles

kshepard:


python/cac_tripplanner/cms/models.py
```diff
@@ -1,9 +1,14 @@
+from ckeditor.fields import RichTextField
```
flibbertigibbet: Minor nit:  don't we usually import non-django things below django things?

python/cac_tripplanner/destinations/models.py
```diff
@@ -1,26 +1,13 @@
+from ckeditor.fields import RichTextField
```
flibbertigibbet: Same import ordering here.

flibbertigibbet:
+1

python/cac_tripplanner/cms/models.py
kshepard: I was reviewing the coding guidelines yesterday and tried to match it to that: "Imports should be grouped into three sections at the top of the file: Python built-in modules, third-party modules, and local source code (stuff in source control)." Maybe that needs to be revised to have a fourth section for django modules? In the meantime I'll switch it as you suggested.

CloudNiner: I thought this was handled as:
built-in modules
non-django third-party modules
django modules
django third-party modules
local source

Either way, we should move the discussion off line and clarify whatever we choose in the coding guidelines.

PR #94, opened 2015-03-17
Feature/frontend refactor

CloudNiner:
Part 1: Moves js code for each of the sidebar tabs into its own control

Wanted to get this reviewed before it gets any bigger and I do any more

@flibbertigibbet might need to pair with you to merge this with your latest PR. It's a nightmare.


CloudNiner:
@flibbertigibbet Just finished the merge conflicts. Will you pull down this branch and verify that everything works as expected?

flibbertigibbet:
Yup, looks alright.

src/app/scripts/cac/control/cac-control-sidebar-directions.js
```diff
@@ -0,0 +1,198 @@
...
+
+    return SidebarDirectionsControl;
+
+    function planTrip() {
+        if (!(directions.origin && directions.destination)) {
+            setDirectionsError();
+            return;
+        }
+
+        var picker = $('#datetimeDirections').data('DateTimePicker');
```
kshepard: For all of these css selector ids/classes, I think it would be better to use the jQuery plugin pattern of having them specified as defaults and then allowing them to be overridden with options, if desired, when the controller is initialized. Moving all of them outside the controller logic would make things cleaner.

src/app/scripts/cac/control/cac-control-sidebar-tab.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+CAC.Control.SidebarTab = (function ($) {
+
+    'use strict';
+
+    var defaults = {
+        tabWrapperSelector: '.sidebar-tabs'
+    };
+
+    var events = {
```
kshepard: I like this events object and think it should also be implemented for the other controls.

kshepard:
+1, looking much better already. The pages modules are so much more manageable now.

src/app/scripts/cac/control/cac-control-sidebar-directions.js
CloudNiner: Hmm, ok. There's a ton of them. I'll try to roll as many of them as I can, and we can refactor any other modules as we go with this pattern. 

However, going to merge this as is for now and add the selector defaults in another PR, as @flibbertigibbet wants to use this branch for something else.

PR #98, opened 2015-03-17
Use flag in new plugin release to always write jshint report.

flibbertigibbet:
Fixes issue with Jenkins failing build if empty lint file found.


flibbertigibbet:
Pretty trivial.  Going to merge this to check that it pleases Jenkins.

PR #99, opened 2015-03-17
Feature/prepopulate destinatinos

CloudNiner:


flibbertigibbet:
Oh, interesting.  Looks like Django has shifted from [fixtures to migrations](https://docs.djangoproject.com/en/1.7/howto/initial-data/) for pre-loading data.  Okay.

Would be nice if we could get some images to go with these; I assume we'll want those for demo.

+1

CloudNiner:
Yeah, not sure how I feel about that. Also not sure how to add default images. Thoughts?

flibbertigibbet:
@kshepard might have some ideas about a good way to get them on our bucket and get the references to them.  Maybe @jbranigan has some ideas about where we should go for appropriate images (design team?)

kshepard:
The bucket is only used if it's specified in the secrets file. Then it uses it for the media directory. In development, it just stores the images locally. There's no way to mix and match. Off the top of my head, I'm not sure the best way to go about adding default images in a data migration.

CloudNiner:
Yeah, ok. I added a default on the destinations sidebar to use placehold.it, maybe we just have a default image for every place the images are displayed if none exists. 

Also, since the ImageField is just a string path to the image in the static dir, maybe we could just put some default images in the static directory and update the image paths in this migration?

CloudNiner:
Going to merge this for now, and maybe figure out images later when we have some?

jbranigan:
I grabbed some images from Google (not licensed) to be used only for testing. I resized them - they should all be accurate - and placed them in /projects/CleanAirCouncil_TripPlanner/data/test-images/

PR #100, opened 2015-03-17
Switch to Azavea OpenTripPlanner fork.

flibbertigibbet:
CAC branch of v0.13 has elevation data loading enabled.

Elevation files found in data directory will be used to build graph.


CloudNiner:
+1 might want to update the README with the proper files to place into the otp_data dir.

PR #102, opened 2015-03-17
Feature/pages map selectors

CloudNiner:
Also removes typeahead init from pages-map and moves a separate instance of typeahead to each sub control


kshepard:
+1

PR #103, opened 2015-03-17
Fix map interaction under sidebar

flibbertigibbet:


kshepard:
+1, tested it out and it works great.

PR #106, opened 2015-03-17
Fix OTP routing

kshepard:
This pulls in the routing URL setting from Django
into javascript in order to remove the hardcoded URL.

### DIFF
--- a/python/cac_tripplanner/templates/base.html
+++ b/python/cac_tripplanner/templates/base.html
@@ -69,11 +69,6 @@
 
     {% endblock %}
 
-    <script type="text/javascript">
-      // Settings pulled in from server-side
-      CAC.Settings.routingUrl = '{{ routing_url }}';
-    </script>
-
     {% block jspage %}
     {% endblock %}
   </body>
--- a/python/cac_tripplanner/templates/base.html
+++ b/python/cac_tripplanner/templates/base.html
@@ -69,6 +69,11 @@
 
     {% endblock %}
 
+    <script type="text/javascript">
+      // Settings pulled in from server-side
+      CAC.Settings.routingUrl = '{{ routing_url }}';
+    </script>
+
     {% block jspage %}
     {% endblock %}
   </body>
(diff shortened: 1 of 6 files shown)


python/cac_tripplanner/templates/base.html
```diff
@@ -69,6 +69,11 @@
 
     {% endblock %}
 
+    <script type="text/javascript">
```
flibbertigibbet: Shouldn't this block be in `map.html`, rather than the base page?  Looks like it's on the `map` view only.

src/app/scripts/cac/routing/cac-routing-plans.js
```diff
@@ -25,7 +23,7 @@ CAC.Routing.Plans = (function($, L, moment, _, UserPreferences, Itinerary) {
     function planTrip(coordsFrom, coordsTo, when, mode, arriveBy) {
         var deferred = $.Deferred();
         var urlParams = prepareParamString(coordsFrom, coordsTo, when, mode, arriveBy);
-        var requestUrl = routingUrl + '?' + urlParams;
+        var requestUrl = Settings.routingUrl + '?' + urlParams;
```
flibbertigibbet: I know what you're working on doesn't touch the formatting part, but I just noticed that perhaps it would be better to pass the parameters object directly to the `$.ajax` call's `data` argument, to let it do the URL formatting, rather than using a combination of `prepareParamString` invoking `$.param`, and then having to concatenate with the `?` here.

flibbertigibbet:
+1, apart from consideration of template snippet move.

python/cac_tripplanner/templates/base.html
kshepard: That's a good point, I'll move it. We may want to move it back here if we end up using the settings on other pages.

src/app/scripts/cac/routing/cac-routing-plans.js
kshepard: That sounds way cleaner to me. I'll try it out.

kshepard:
The branch has been updated with those changes. I'll merge when Jenkins stabilizes.

PR #107, opened 2015-03-17
Remove auth params from public S3 resources

kshepard:
@hectcastro -- you may want to take a look at this as well. It works great from my testing.


notthatbreezy:
Nice, +1

hectcastro:
Confirmed that it works in EC2 with IAM roles. Thanks, @kshepard. :+1:

PR #108, opened 2015-03-18
Feature/uwishunu updates

CloudNiner:
This can be reviewed awhile, just need to fix and add some additional tests.


python/cac_tripplanner/destinations/management/commands/load_events.py
```diff
@@ -24,10 +26,17 @@ def get_property(self, item, property_name):
             return None
         return elements[0].firstChild.data
 
+    def parse_date(self, string_datetime):
+        """ Parse a date from the feed, return datetime localized to US/Eastern """
+        if not string_datetime:
+            return None
+        parsed_date = datetime.strptime(string_datetime, '%a, %d %b %Y %H:%M:%S +0000')
```
flibbertigibbet: Might consider wrapping this in a `try`, in case the feed has any weirdly formatted date strings.

flibbertigibbet:
+1

PR #111, opened 2015-03-19
Add data migration to include images for the default destinations

CloudNiner:


kshepard:
I've copied this default images up to the development S3 bucket so they also work with S3 storage enabled. This will need to be done manually whenever a new bucket is to be used. It may be worth adding a comment somewhere that states this.

kshepard:
+1, tested, works great!

CloudNiner:
Good point. Ill add it in the settings file.

PR #114, opened 2015-03-19
Feature/directions detail

CloudNiner:
Very WIP, but here's the public interface for the directions list. Going to implement next.


kshepard:
+1, interface looks good to me.

PR #119, opened 2015-03-20
Feature/more conditional options

flibbertigibbet:
This finishes up implementing the extra options on the map page.

### DIFF
--- a/src/app/scripts/cac/control/cac-control-sidebar-explore.js
+++ b/src/app/scripts/cac/control/cac-control-sidebar-explore.js
@@ -2,7 +2,7 @@
  *  View control for the sidebar explore tab
  *
  */
-CAC.Control.SidebarExplore = (function ($, MapTemplates, Typeahead, UserPreferences) {
+CAC.Control.SidebarExplore = (function ($, BikeOptions, MapTemplates, Typeahead, UserPreferences) {
 
     'use strict';
 
@@ -34,6 +34,7 @@ CAC.Control.SidebarExplore = (function ($, MapTemplates, Typeahead, UserPreferen
         destinationDirections: 'cac:control:sidebarexplore:destinationdirections'
     };
 
+    var bikeOptions = null;
     var datepicker = null;
     var mapControl = null;
     var typeahead = null;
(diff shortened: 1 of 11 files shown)


src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -120,6 +144,23 @@ CAC.Map.Control = (function ($, L, _) {
     }
 
     /**
+     * Show/hide sidebar options based on the selected mode.
+     * Expects both tabs to have the same selector names for the toggleable divs.
+     */
+    function changeMode(selectors) {
```
CloudNiner: This a really weird spot for this. Can it not go on the map page control?

CloudNiner: Or better yet, if it needs to be on both sidebars, add a new BikeOptionsControl that encapsulates all the bike options logic. Then the bike triangle definition can come off the map control as well, since that seems like a weird spot for that as well.

flibbertigibbet:
Okay, refactored.  I'd put it on the map control because both sidebars had reference to it.

src/app/scripts/cac/control/cac-control-sidebar-explore.js
```diff
@@ -60,15 +75,45 @@ CAC.Control.SidebarExplore = (function ($, MapTemplates, Typeahead, UserPreferen
...
+        var date = picker.date();
+        if (!date) {
+            // use current date/time if none set
+            date = moment();
+        }
+
+        var otpOptions = { mode: mode };
+
+        if (mode.indexOf('BICYCLE') > -1) {
+            var bikeTriangleOpt = $('option:selected', options.selectors.bikeTriangleDiv);
```
CloudNiner: Seems like the BikeControl is still leaky here, since the selectors are used both in the bike control and outside it. If you expose methods on the bikecontrol that provide the three values you're pulling, then the change mode method doesn't need the selectors argument and you can just set the selectors as configurable defaults in that control.

CloudNiner:
+1

PR #120, opened 2015-03-20
Feature/markers origin destination

flibbertigibbet:
* Show markers on trip origin and destination points.
* Clear query results on tab switch
* Do not show query results if user switched to the other tab

### DIFF
--- a/src/app/scripts/cac/control/cac-control-sidebar-directions.js
+++ b/src/app/scripts/cac/control/cac-control-sidebar-directions.js
@@ -15,6 +15,7 @@ CAC.Control.SidebarDirections = (function ($, Control, BikeOptions, MapTemplates
             buttonPlanTrip: 'section.directions button[type=submit]',
             checkboxArriveBy: 'input[name="arriveByDirections"]:checked',
             datepicker: '#datetimeDirections',
+            directions: '.directions',
             arriveByButton: 'input[name="arriveByDirections"]:eq(1)',
             maxWalkDiv: '#directionsMaxWalk',
             modeSelector: '#directionsModeSelector',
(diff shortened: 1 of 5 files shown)


src/app/scripts/cac/control/cac-control-sidebar-directions.js
```diff
@@ -92,6 +93,14 @@ CAC.Control.SidebarDirections = (function ($, Control, BikeOptions, MapTemplates
         bikeOptions.changeMode(options.selectors);
     }
 
+    function clearDirections() {
+        mapControl.setOriginDestinationMarkers(null, null);
+        mapControl.clearItineraries();
+        itineraryListControl.hide();
+        directionsListControl.hide();
+        $('.directions').removeClass('show-results');
```
kshepard: This should be defined in the selectors object. Same goes for the code not added in this PR that adds the class.

src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -282,4 +293,56 @@ CAC.Map.Control = (function ($, L, _) {
         map.panTo(latLng);
     }
 
+    /**
+     * Show markers for trip origin/destination.
+     * Will unset the markers if either co-ordinate set is null/empty.
+     *
+     * @param {Array} originCoords Start point co-ordinates [lat, lng]
+     * @param {Array} destinationCoords End point co-ordinates [lat, lng]
```
kshepard: Typo (I think): no hyphen in coordinates.

kshepard:
+1

src/app/scripts/cac/map/cac-map-control.js
flibbertigibbet: Different spelling of the same, but ok.

PR #126, opened 2015-03-23
Limit rich text editor options.

flibbertigibbet:
Explicitly define what is available in the toolbar, removing options such as smileys and image embedding.


flibbertigibbet:
image

CloudNiner:
LGTM, that's convenient

PR #130, opened 2015-03-23
Remember last entry for other map tab.

flibbertigibbet:
Reverts behavior change in refactor.
Both tabs on map page pre-populate with last search term.


kshepard:
+1

PR #131, opened 2015-03-24
Feature/styling changes cleanup

flibbertigibbet:


src/app/styles/components/_sidebar-explore.scss
```diff
@@ -135,7 +136,8 @@
         padding: 0 15px;
         position: relative;
         width: 100%;
-        height: 0;
+        height: 100%;      // TODO: remove and dynamically resize
+        max-height: 400px; // TODO: remove and dynamically resize
```
CloudNiner: For consistency and ease of changing if necessary, the '400px' could be a sass var since its used in multiple places.

CloudNiner:
+1

flibbertigibbet:
This is temporary, so we can see the sidebar results until they dynamically resize.

PR #132, opened 2015-03-24
Use bootstrap-select

flibbertigibbet:
Use [bootstrap-select](https://github.com/silviomoreto/bootstrap-select) for drop-downs, so @mtedeschi  can more easily add styling.


CloudNiner:
+1, assuming the gulp prod build still works correctly

flibbertigibbet:
`npm run gulp-development` executes successfully, and tests still pass.  Why is the prod build a concern with this?

CloudNiner:
Oh nvmnd, this is just the map file, sorry.

PR #140, opened 2015-03-24
Feature/dynamic resize sidepanel

### LINKED ISSUE #134: Make side panel details resize dynamically
Map page side panel results div is fixed in size for both tabs.  It should set its height based on the size of the results to display, up to some max.

There are some TODOs in the scss with fixed sizes to remove.

flibbertigibbet:
Fixes #134.


CloudNiner:
+1

PR #141, opened 2015-03-26
Set accept header for isochrone endpoint.

flibbertigibbet:
Fixes non-deterministic behavior where OTP occasionally returns a shapefile
instead of GeoJSON.


CloudNiner:
+1

PR #144, opened 2015-03-27
Feature/add loader

flibbertigibbet:
Add spinkit CSS loader animation to display during asynchronous queries.


CloudNiner:
The spinner should really be its own newable control, and be passed a selector to draw into, with show/hide methods, but other than that this LGTM.

flibbertigibbet:
Ok.  I'm not sure what the benefit is of making it a control and adding it dynamically, as it doesn't require templating.

CloudNiner:
Well, there is a template, its this:
```
<div class="sk-spinner sk-spinner-three-bounce hidden">
            <div class="sk-bounce1"></div>
            <div class="sk-bounce2"></div>
            <div class="sk-bounce3"></div>
</div>
```

Two other benefits are:
- a spinner can be added anywhere else on the page trivially
- We dont have to come back to the project a year from now and remember all of the specific selectors, template and add/remove class calls that make the spinner work as intended

flibbertigibbet:
That's not really a template,as it's static content.  It's just an HTML snippet, that's being shown/hidden with the bootstrap 'hidden' class, like a bunch of the other content.  I generally think of newable objects as being things with dynamic content and mutable state.  I'll go make the control, but not sure if that's really saving more complexity that it adds.

CloudNiner:
True, I would agree, although I tend to extend that definition to include anything that is reusable. Which this definitely is. I also think this saves complexity if done right, since you don't have to manually add and keep track of classes/html every time you add a spinner. You just do:
```
var destinationsSpinner = new CAC.Spinner(selector);
...
destinationsSpinner.show();
...
destinationsSpinner.hide();
```

In this case, up to you either way, especially if there aren't going to be additional spinners to add.

PR #147, opened 2015-03-27
Fix getting directions from explore tab.

### LINKED ISSUE #129: No way back to features list from detail view
Once a feature is clicked in the 'explore' tab, the app switches to the 'directions' tab.  From there, navigating back to the 'explore' tab now shows the detail description of the feature (which was not visible before), but there's no way to go back to the summary list of all the features.  Also, the only way to view the detailed description of a feature is to switch back to the 'explore' tab after seeing directions to it.

### LINKED ISSUE #145: Itinerary directions list does not update when directions change
Changing directions to/from address will search for new directions and draw line with correct origin destinations, but going to view the directions details, old origin/destination sometimes show.

flibbertigibbet:
Fixes #145.  Fixes #129.


kshepard:
+1, works great. While you're at it, may also be worth adding a little styling to the "Back" and "Get Directions" links so that they have a pointer on hover.

PR #150, opened 2015-03-27
Feature/collapse layer switcher

flibbertigibbet:
Don't know if we actually want to use this, but here's a somewhat janky but functional minimize/maximize button for the layer switcher.


kshepard:
Cool! I think with some minor styling tweaks this could work well.

jbranigan:
Is this something that can be merged?

flibbertigibbet:
I wasn't sure if we wanted to try this out or not.  I'll close this PR and open another one, since this one is stale.

PR #174, opened 2015-04-06
Closes #152: Always show all featured dests on map and sidebar

### LINKED ISSUE #152: Always display featured locations on the map
This wasn't necessarily clear in the wireframes, but if a user pans outside of their isochrone, the other destinations should be there. They don't necessarily need to be listed in the sidebar, although that is an option as well.

CloudNiner:
Simplest way to show all destinations is to return them sorted
by distance with the isochrone. Could refactor to pull list
of destinations from /api/destinations, but not really worth it
atm, since that would require moving around some app logic
to fetch/display them.


flibbertigibbet:
+1

PR #179, opened 2015-04-06
Used cached distance to features.

### LINKED ISSUE #178: distances to destinations requeried on going back from details
When a featured destination in the sidebar gets clicked, its details show.  On clicking the 'back' button in the details section, the featured destinations are listed again, without the distances, which then get queried again and repopulated.

flibbertigibbet:
Fixes #178.

### DIFF
--- a/src/app/scripts/cac/control/cac-control-sidebar-explore.js
+++ b/src/app/scripts/cac/control/cac-control-sidebar-explore.js
@@ -216,10 +216,15 @@ CAC.Control.SidebarExplore = (function ($, BikeOptions, MapTemplates, Routing, T
      */
     function setDestinationDistance(destination, $container) {
 
+        // helper to set the text snippet
+        function setDestinationMinutesText(distanceMinutes) {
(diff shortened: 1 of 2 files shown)


src/app/scripts/cac/control/cac-control-sidebar-explore.js
```diff
@@ -215,6 +215,15 @@ CAC.Control.SidebarExplore = (function ($, BikeOptions, MapTemplates, Routing, T
      * @param {Object} $container jQuery-selected HTML snippet in the sidebar for the destination
      */
     function setDestinationDistance(destination, $container) {
+
+        // first check if we have distance cached
+        if (destination.durationMinutes) {
+            $container.find(options.selectors.distanceMinutesText)
+                .text(destination.durationMinutes + ' minutes away');
```
kshepard: Since this bit of code now appears twice, it may be worth abstracting it to a function that takes a $container and number of minutes. Then if we need to change the formatting, it can be done in a single place.

kshepard:
+1

PR #181, opened 2015-04-07
Feature/drag origin dest

flibbertigibbet:


src/app/scripts/cac/search/cac-geocoder.js
```diff
@@ -54,4 +56,27 @@ CAC.Search.Geocoder = (function ($) {
         return dfd.promise();
     }
 
+    function reverse(lat, lng) {
+        var dfd = $.Deferred();
+
+        var params = {
+            location: [lng, lat].join(','),
```
kshepard: May want to increase the default search radius (via the 'distance' parameter). I was playing around with dragging the markers and there were many times where I'd drop it in the middle of a park and it wouldn't return any results.

flibbertigibbet: Ah, okay.  I hadn't seen that yet.

kshepard:
+1

PR #182, opened 2015-04-07
Move to v0.14 on main OTP repo

flibbertigibbet:
This switches OTP to the new 0.14 release, and updates the build and serve parameters for the changes.  It switches back to the main repo, as they've since re-enabled loading elevation data.

### DIFF
--- a/deployment/ansible/roles/cac-tripplanner.otp-data/templates/otp.conf.j2
+++ b/deployment/ansible/roles/cac-tripplanner.otp-data/templates/otp.conf.j2
@@ -1,0 +1,13 @@
+## NOTICE: This file is written by ansible, and any changes made here will be overwritten on next provision.
+#           Modify cac-tripplanner.otp-data/templates/otp.conf.j2 to make changes stick.
+description "start OpenTripPlanner process"
+
+start on {{ upstart_start_on }}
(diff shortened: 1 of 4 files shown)


deployment/ansible/roles/cac-tripplanner.otp-data/templates/otp.conf.j2
```diff
@@ -0,0 +1,13 @@
+## NOTICE: This file is written by ansible, and any changes made here will be overwritten on next provision.
+#           Modify azavea.opentripplanner/templates/otp.conf.j2 to make changes stick.
```
CloudNiner: `cac-trippplanner.otp-data`?

Although, on a related note, I'd be ok with modifying the azavea opentripplanner role to require you to override a template, since that's way more flexible.

CloudNiner:
+1

deployment/ansible/roles/cac-tripplanner.otp-data/templates/otp.conf.j2
flibbertigibbet: Yeah, good idea.  Wasn't sure what would be a good way to change the template based on version, especially since we can use commits or other non-release tags, and forks.

PR #183, opened 2015-04-07
Show loader while reverse geocoding dragged marker.

flibbertigibbet:
Also adds error message display on failed reverse geocode.


kshepard:
+1

PR #184, opened 2015-04-07
Add step-by-step interactivity

kshepard:
Closes #166

Issue #158, opened 2015-03-31
Add info to routing result


jbranigan:
Primarily for step-by-step:

*   Time, distance
*   Mode icons (walk, bike, bus, train)
*   Include which bus or train to get on (and remove turn directions while on buses)

Realizing we may have lifted the set-by-step parser from a different project that was walk-only. Hopefully we are getting the required data returned.

PR #185, opened 2015-04-07
Feature/add routing info

flibbertigibbet:
This gets most of the way to addressing #158.

Will do the rest after #184 is merged.


kshepard:
+1

PR #191, opened 2015-04-09
Extend geocoder search categories

flibbertigibbet:
The 'Populated Place' category adds city and neighborhood names; the suggest service will preferentially suggest street names first, if there are some that also match.  Adding 'Postal' as a category seems to make the service better at finding some addresses (e.g., addresses on Girard would not geocode before.)

### DIFF
--- a/src/app/scripts/cac/search/cac-typeahead.js
+++ b/src/app/scripts/cac/search/cac-typeahead.js
@@ -121,6 +121,15 @@ CAC.Search.Typeahead = (function ($) {
 
     function suggestAdapterFactory() {
         var url = 'http://geocode.arcgis.com/arcgis/rest/services/World/GeocodeServer/suggest';
+
+        // search categories from here:
+        // https://developers.arcgis.com/rest/geocode/api-reference/geocoding-category-filtering.htm
+        var searchCategories = ['Address',
+                                'Postal',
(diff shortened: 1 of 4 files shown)


flibbertigibbet:
image

src/app/scripts/cac/search/cac-typeahead.js
```diff
@@ -128,7 +137,7 @@ CAC.Search.Typeahead = (function ($) {
                 '-75.126531',
                 '39.967842'
             ].join(','),
-            category: 'Address,POI',
+            category: searchCategories,
```
kshepard: I think these fields are also used on the geocoder. Do you know if they need to change there as well?

kshepard:
+1, passes the "East Falls" test.

src/app/scripts/cac/search/cac-typeahead.js
flibbertigibbet: So they are.  I'll move the categories to a shared module; the bounds should be shared, too.

kshepard: Awesome, sounds good.

CloudNiner:
You could try a center and radius instead, so that it prefers closer destinations. But, that has other potential issues.

PR #193, opened 2015-04-09
Extend search area

flibbertigibbet:
Extend search area to cover counties here:
http://pecpa.org/wp-content/uploads/Recreation-The-Circuit-map-photo-Patrick-1260x979.jpg


flibbertigibbet:
Trivial, merging

PR #197, opened 2015-04-10
Do not download OSM on provision.

flibbertigibbet:
Expect to find OSM file(s) with GTFS, other OTP files.

We will want to use a custom extract to cover our specific bounding box, rather than use the MapZen Philadelphia extract.


kshepard:
+1

PR #199, opened 2015-04-10
Automatically send OTP query on any field change

### LINKED ISSUE #155: Map page refresh on options changed
We should update the map as users tweak settings - both for directions and isochrone (within reason).

kshepard:
Wires up listeners to all input fields and debounces the
called functions. Also only allows for one open OTP request
at a time.

Closes #155

### DIFF
--- a/src/app/scripts/cac/control/cac-control-sidebar-directions.js
+++ b/src/app/scripts/cac/control/cac-control-sidebar-directions.js
@@ -10,7 +10,7 @@ CAC.Control.SidebarDirections = (function ($, Control, BikeOptions, Geocoder, Ma
     var METERS_PER_MILE = 1609.34;
 
     // Number of millis to wait on input changes before sending directions request
-    var directionDebounceMillis = 750;
+    var DIRECTION_DEBOUNCE_MILLIS = 750;
 
     var defaults = {
         selectors: {
@@ -229,12 +229,13 @@ CAC.Control.SidebarDirections = (function ($, Control, BikeOptions, Geocoder, Ma
             $(options.selectors.directions).addClass(options.selectors.resultsClass);
             itineraryListControl.show();
         });
-    }, directionDebounceMillis);
+    }, DIRECTION_DEBOUNCE_MILLIS);
 
     return SidebarDirectionsControl;
 
     function changeMode() {
         bikeOptions.changeMode(options.selectors);
+        planTrip();
     }
 
     function clearDirections() {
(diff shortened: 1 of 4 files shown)


src/app/scripts/cac/control/cac-control-sidebar-directions.js
```diff
@@ -102,21 +114,20 @@ CAC.Control.SidebarDirections = (function ($, Control, BikeOptions, Geocoder, Ma
         setDirections: setDirections
     };
 
-    return SidebarDirectionsControl;
-
-    function changeMode() {
```
flibbertigibbet: Were these functions deleted by accident?  `changeMode` and `clearDirections` seem to still be needed.

src/app/scripts/cac/control/cac-control-sidebar-directions.js
```diff
@@ -9,6 +9,9 @@ CAC.Control.SidebarDirections = (function ($, Control, BikeOptions, Geocoder, Ma
 
     var METERS_PER_MILE = 1609.34;
 
+    // Number of millis to wait on input changes before sending directions request
+    var directionDebounceMillis = 750;
```
flibbertigibbet: Might make sense to either capitalize the added constants, or change `METERS_PER_MILE` to lower case, for consistency.

src/app/scripts/cac/control/cac-control-sidebar-directions.js
```diff
@@ -93,6 +102,9 @@ CAC.Control.SidebarDirections = (function ($, Control, BikeOptions, Geocoder, Ma
 
         setFromUserPreferences();
         changeMode();
+
+        // Respond to changes on all direction input fields
+        $(options.selectors.directionInput).on('input change', planTrip);
```
flibbertigibbet: What happens when the mode is changed?  As it already has its own handler, I think it would be better to remove the new input class from it, and instead call `planTrip` from the end of its handler.  (Same with isochrone controller.) I imagine it works due to the debounce timeout letting the mode change handler finish first, but better to be explicit with that.

src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -66,6 +66,16 @@ CAC.Map.Control = (function ($, Handlebars, L, _) {
         'under <a href="http://creativecommons.org/licenses/by-sa/3.0">CC BY SA</a>.'
     ].join('');
 
+    // Variables used for limiting to one isochrone request at a time.
```
flibbertigibbet: Change to block comment here?

flibbertigibbet:
+1 pending mode change handler modification

flibbertigibbet:
Do we not want to trigger updates on typeahead address field changes?  Looks like that's not happening.

kshepard:
I intentionally left that out since there was a card for adding behavior to the typeahead selection and I didn't want to interfere with it. Since you're now finished up with that card, I think it would definitely make sense to wire that up as well.

PR #201, opened 2015-04-13
Feature/ansible secrets

flibbertigibbet:


deployment/ansible/group_vars/all
```diff
@@ -1,12 +0,0 @@
----
```
kshepard: It feels a little weird to me to template this file, since a lot of those variables won't ever be customized (e.g. the OTP version, jarfile, repo, etc). I'm thinking it would make more sense to template the development group_vars file instead -- put the secrets in there, and have the all file contain the things that are more or less static.

kshepard:
+1 aside from templating development instead of all.

deployment/ansible/group_vars/all
flibbertigibbet: Agreed, this didn't feel better than just templating `settings.py` directly.  Not sure if separately templating each of the environment `group_vars` files is ideal either, but ok, that's an improvement.

PR #202, opened 2015-04-13
Feature/typeahead enter key

flibbertigibbet:
When enter key is pressed in a typeahead field, automatically select the first suggestion.

Uses the [undocumented `autoselect` option](https://github.com/twitter/typeahead.js/issues/32).


kshepard:
+1

RickMohr:
Heads up that the (undocumented) `autoselect` option was removed in version 0.11.1

flibbertigibbet:
Good to know. Thanks @RickMohr!

PR #203, opened 2015-04-13
Set times to destinations after templates have been added to DOM.

### LINKED ISSUE #198: Times to destination are only shown on initial page load search
When another search is issued, the time queries appear to be sent, but the times are never populated.

flibbertigibbet:
Fixes #198.

Occasionally travel times would return after the templated snippet had been added to the DOM, but the snippet would be updated, and not the DOM.  This tracks the snippets in the DOM by destination ID, and sets all travel times after snippets have been added.


kshepard:
+1

PR #204, opened 2015-04-13
Add packer configuration for app/OTP/bastion

kshepard:


deployment/packer/cac.json
```diff
@@ -0,0 +1,94 @@
+{
+  "variables": {
+    "local_project_directory": ".",
+    "aws_access_key": "",
```
flibbertigibbet: Could these be set from the `aws_access_key_id` and `aws_secret_access_key` vars set in the `group_vars` for `cac_secrets`?

flibbertigibbet:
+1

deployment/packer/cac.json
kshepard: Interesting thought. However, those keys in cac_secrets are actually going to be removed very shortly when we switch to using IAM roles.

notthatbreezy: Packer will the common AWS SDK paths (I believe) to grab keys, so passing them in might not be required. We might want to look at some of the authorization helpers in other projects for examples. That way this should be usable with MFA tokens as well, but would require running another command (to authenticate with MFA) before running the packer commands.

kshepard: Cool, I do like the idea of using MFA tokens for this process.

PR #205, opened 2015-04-14
Issue query on address changes

kshepard:
Finishes off #199


python/cac_tripplanner/cac_tripplanner/settings.py
```diff
@@ -36,6 +36,8 @@
         'otp_url': 'http://192.168.8.26:8080/otp/routers/{router}/',
         'allowed_hosts': ['127.0.0.1', 'localhost'],
         'internal_ips': ['0.0.0.0', '127.0.0.1'],
+        'internal_ips': ['0.0.0.0', '127.0.0.1'],
```
flibbertigibbet: Duplicate line here?

flibbertigibbet:
+1

python/cac_tripplanner/cac_tripplanner/settings.py
kshepard: Hm, yes, good catch!

PR #207, opened 2015-04-14
Prefer locations near user in typeahead.

### LINKED ISSUE #154: Autocomplete results enhancements
*   Some destinations don't autocomplete (King of Prussia Mall, Longwood Gardens) - this may be related to the extent of the suggest/search radius
*   City/area names don't geocode (i.e. East Falls)
*   Should discuss options when a user's search can't be resolved
*   This kind of needs to work as well as Google Maps (I know) - we may want to consider switching libraries
*   This was the primary hindrance to usability
*   Requires discussion

flibbertigibbet:
If current location not available, set bounding box on suggest service instead.
Due to ESRI bug, cannot specify both at the moment.
Closes #154.


kshepard:
+1

PR #208, opened 2015-04-14
Feature/fit isochrone

flibbertigibbet:
Zoom map to fit isochrone bounds.

Also makes isochrone marker draggable, and adds ability to display error messages on discover tab, to match behavior of directions tab.


kshepard:
+1

PR #214, opened 2015-04-20
Fix problems on page load

kshepard:
The explore and directions module were both calling changeMode upon
page load, and it was causing problems with extra requests and also
changing preferences (so if you reloaded the explore page it would
actually redirect you to the directions page).


flibbertigibbet:
:+1:

PR #215, opened 2015-04-21
Feature/wire mode buttons

flibbertigibbet:


src/app/scripts/cac/control/cac-control-sidebar-directions.js
```diff
@@ -116,17 +112,12 @@ CAC.Control.SidebarDirections = (function ($, Control, BikeOptions, Geocoder, Ma
...
-     * This function has been debounced to cut down on requests.
+     * Throttled to cut down on requests.
      */
-    var planTrip = _.debounce(function() {
-        // If there's already an open plan request, wait until it finishes.
-        // This is automatically handled gracefully via debounce.
-        if (activePlanRequest) {
-            pendingPlanRequest = true;
-            planTrip();
+    var planTrip = _.throttle(function() {
```
kshepard: Is throttle really better than debounce here? I haven't pulled down this code and tested, but I'd think using throttle would allow for the user to change inputs in such a way that there most recent change never triggers a new request.

kshepard:
+1, lots of nice cleanup.

src/app/scripts/cac/control/cac-control-sidebar-directions.js
flibbertigibbet: `_.throttle` makes both a leading-edge and a trailing-edge call.  I did like having the logic there to ensure there was only one query out at a time, but that seems to be more important for the isochrone fetches, and this seems to have sorted out some issues with jittery behavior.

PR #219, opened 2015-04-23
Add bike share locations layer with popups

flibbertigibbet:


flibbertigibbet:
image

Curiously, the locations are all closed for four minutes a day.

kshepard:
+1

PR #221, opened 2015-04-24
Feature/about faq pages

flibbertigibbet:
Add editable About and FAQ pages to admin interface that cannot be added or deleted.


kshepard:
I was assuming these pages would be used for the links in the footer: "Terms of Use", "Data Policy", and "Privacy Policy". And if that is the case, those links should also be wired up to their respective pages. It'd probably be good to check in with @jbranigan on this to make sure.

kshepard:
+1 after verifying the pages are correct and potentially wiring the links up.

flibbertigibbet:
Yeah, I noticed we had three footer links, but only two new pages, and wasn't sure what the plan was with that.

flibbertigibbet:
Merging this.  Will put out separate PR once there's clarification on links, etc.

PR #223, opened 2015-04-24
Save featured location destination to preferences in expected format.

flibbertigibbet:
Fixes issue where if featured location picked for destination, coordinates for it not found on page reload.


kshepard:
+1

PR #224, opened 2015-04-27
Feature/pois

flibbertigibbet:
Enable POIs in typeahead suggestions.  Fixes issue with POIs not geocoding properly when the place name comes before its street address.  Disables ranking search results by distance in favor of using bounding box, as requested by @jbranigan.


src/app/scripts/cac/search/cac-typeahead.js
```diff
@@ -154,6 +154,9 @@ CAC.Search.Typeahead = (function (_, $, SearchParams) {
             f: 'pjson'
         };
 
+        // TODO: re-enamble result distance ranking when bug fixed to work with searchExtent
```
kshepard: Minor typo

kshepard:
+1, seems to work really well.

PR #226, opened 2015-04-27
Hide other itineraries when one clicked.

flibbertigibbet:
Also highlight itineraries on map and in sidebar on hover.

### DIFF
--- a/src/app/scripts/cac/control/cac-control-itinerary-list.js
+++ b/src/app/scripts/cac/control/cac-control-itinerary-list.js
@@ -21,7 +21,8 @@ CAC.Control.ItineraryList = (function ($, Handlebars, Utils) {
 
     var events = $({});
     var eventNames = {
-        itineraryClicked: 'cac:control:itinerarylist:itineraryclicked'
+        itineraryClicked: 'cac:control:itinerarylist:itineraryclicked',
+        itineraryHover: 'cac:control:itinerarylist:itineraryhover'
     };
 
     var $container = null;
@@ -38,6 +39,7 @@ CAC.Control.ItineraryList = (function ($, Handlebars, Utils) {
         eventNames: eventNames,
         setItineraries: setItineraries,
         setItinerariesError: setItinerariesError,
+        showItineraries: showItineraries,
         show: show,
         hide: hide,
         toggle: toggle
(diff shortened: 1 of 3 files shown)


src/app/scripts/cac/control/cac-control-itinerary-list.js
```diff
@@ -106,10 +110,30 @@ CAC.Control.ItineraryList = (function ($, Handlebars, Utils) {
         events.trigger(eventNames.itineraryClicked, itinerary);
     }
 
+    /**
+     * Handle hover event on itinerary list item
+     */
+    function onItineraryHover() {
+        var itineraryId = this.getAttribute('data-itinerary');
```
kshepard: Alternatively, this can be done with: `$(this).data('itinerary')`, which is arguably slightly cleaner.

src/app/scripts/cac/control/cac-control-sidebar-directions.js
```diff
@@ -253,6 +261,21 @@ CAC.Control.SidebarDirections = (function ($, Control, BikeModeOptions, Geocoder
         }
     }
 
+    /**
+     * Handles hover events to highlight a given itinerary
+     */
+    function onItineraryHover(event, itinerary) {
+        if (itinerary) {
+            $(options.selectors.itineraryById + currentItinerary.id + '"]')
```
kshepard: I don't love this selector that requires also supplying the closing bracket when using it, particularly because it needs to happen twice here. Maybe a better way to go would be to have a selector for .block-itinerary, and a function getItineraryById that abstracts the dirty work of this query.

kshepard:
+1

src/app/scripts/cac/control/cac-control-sidebar-directions.js
flibbertigibbet: Me either.  Went with this to reduce the amount of things to join.  Will go with something else.

src/app/scripts/cac/control/cac-control-itinerary-list.js
flibbertigibbet: Borrowed that bit from the click handler.  Looking at the jQuery docs, it wasn't clear to me whether `data` was appropriate for fetching the value, or just setting it.

PR #228, opened 2015-04-27
Reverse geocode OSM node labels.

### LINKED ISSUE #227: Handle OSM nodes in place names
Occasionally OpenTripPlanner won't have a name for a direction step origin or destination, and so will label it `osm:node:[some number]` in the response.

The directions UI should hide or re-label these, perhaps by reverse-geocoding the point, or using a generic label.

image

flibbertigibbet:
Fixes #227.


kshepard:
+1

PR #235, opened 2015-04-30
Don't run dist-upgrade when creating AMIs

kshepard:
We are always grabbing the latest available AMI, so there should
be no reason to perform an upgrade. And if there is a change, such
as a kernel update, it could require user input when installing
the bootloader, which will cause Packer to fail.


flibbertigibbet:
+1

PR #240, opened 2015-04-30
Feature/social direction sharing

flibbertigibbet:
Not all the way there, but mostly.  Would be nice to get this on a test server, since there are some things that can't be tested if the content isn't publicly available.

### DIFF
--- a/python/cac_tripplanner/templates/base.html
+++ b/python/cac_tripplanner/templates/base.html
@@ -7,6 +7,23 @@
     <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=0">
     <title>{% block title %}Open Transit Planning for Bikes, Public Transit, and Walking | Clean Air Council{% endblock %}</title>
 
+    {% block extrametatags %}
+    <meta property="og:title" content="Trip Planner" />
+    <meta property="og:size_name" content="Go Philly Go" />
+    <meta property="og:url"
+    content="{% if request.is_secure %}https{% else %}http{% endif %}://{{ request.META.HTTP_HOST }}" />
+    <!-- TODO: reword? -->
+    <meta property="og:description"
+        content="CAC Trip Planner, for planning transit, walking, and biking trips around the Delaware Valley region.  Features parks and other destinations." />
+    <!-- TODO: update app ID once we have real app -->
+    <meta property="fb:app_id" content="1594368497497826" />
+    <meta property="og:type" content="article" />
+    <meta property="og:locale" content="en_US" />
+    <meta property="article:author" content="Clean Air Council" />
+    <meta property="article:publisher" content="CAC Trip Planner" />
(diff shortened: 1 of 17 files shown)


python/cac_tripplanner/templates/base.html
```diff
@@ -31,9 +47,19 @@
     <script type="text/javascript" src="https://code.jquery.com/jquery-2.1.3.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/leaflet/0.7.3/leaflet.js"></script>
 
-
     {% if debug %}
 
+    <!-- facebook loads async; might not be there during module init! -->
+    <!-- loading debug version here -->
```
CloudNiner: Is there any downside to loading debug version in production?

CloudNiner: nvmnd, just saw the if debug tag.

src/app/scripts/cac/control/cac-control-directions-list.js
```diff
@@ -207,10 +230,122 @@ CAC.Control.DirectionsList = (function ($, Handlebars, UserPreferences, Utils) {
         }
     }
 
+    function shareOnFacebook(event) {
```
CloudNiner: These three methods probably make more sense in a separate module, since they're all share functionality and dont really appear to use too many vars in the closure.

src/app/scripts/cac/control/cac-control-directions-list.js
```diff
@@ -207,10 +230,122 @@ CAC.Control.DirectionsList = (function ($, Handlebars, UserPreferences, Utils) {
...
+                caption: caption,
+                picture: pictureUrl,
+                link: event.data.url,
+                redirect_uri: options.useHost + '/map'
+            };
+            /* jshint camelcase:true */
+
+            var url = feedUrl + $.param(params);
+            window.open(url, '_blank');
+            event.returnValue = false;
```
CloudNiner: Are these necessary, since you call the exact same methods no matter what just after?

CloudNiner:
LGTM, but strongly consider splitting the share functions to a separate module, so they can be reused elsewhere.

src/app/scripts/cac/control/cac-control-directions-list.js
flibbertigibbet: Sure, can do that.

src/app/scripts/cac/control/cac-control-directions-list.js
flibbertigibbet: You mean setting the `event.returnValue` and `preventDefault`? Nope, not necessary, but doesn't hurt either.  If the button somehow got changed to a submit type, it would be needed.  Those calls came from the documentation boilerplate from both Facebook and Twitter, so I assume they serve some purpose in some cases.

PR #243, opened 2015-05-01
Use shortened links for social sharing

flibbertigibbet:


CloudNiner:
+1 I've seen quite a few console statements sneak in, you might want to stub `window.console` so that pages don't barf if console isn't defined. Only typically an issue in older browsers or IE when dev tools are closed. Or just remove the console statements.

flibbertigibbet:
Oh, meant to remove those.  Good catch.

Issue #234, opened 2015-04-29
Style event popups and print directions


mtedeschi:
- [ ] Style event and bike location map popups
- [ ] Style the print directions page view

flibbertigibbet:
I'm doing a little work on the print directions view, will put in formatted durations.

mtedeschi:
This looks like it's in good shape, but will assign it to me and take a look again.

PR #255, opened 2015-05-04
Style print page

jfrankl:
Style print page. Related to #234. Includes:

- Wrap page in container
- Adding some styling elements that were deeply nested within sidebar classes

image

image

### DIFF
--- a/python/cac_tripplanner/templates/directions.html
+++ b/python/cac_tripplanner/templates/directions.html
@@ -1,10 +1,12 @@
 {% extends "base.html" %}
 {% block content %}
-{% include "header-mixin.html" %}
-<section class="directions">
-  <div class="directions-list">
-  </div>
-</section>
+<div class="print container">
+  {% include "header-mixin.html" %}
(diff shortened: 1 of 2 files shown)


python/cac_tripplanner/templates/directions.html
```diff
@@ -1,10 +1,12 @@
 {% extends "base.html" %}
 {% block content %}
-{% include "header-mixin.html" %}
-<section class="directions">
-  <div class="directions-list">
-  </div>
-</section>
+<div class="print container">
+  {% include "header-mixin.html" %}
+  <section class="directions-print">
```
flibbertigibbet: Not having the `directions` class here will likely break some of the code that expects it, which may be why the `osm:node` labels are showing up in the screenshot (those should refresh to street names).

flibbertigibbet:
+1 from me.  Are we waiting on @mtedeschi to check it out, or shall I merge this?

jbranigan:
LGTM. I think we can go for it.

PR #261, opened 2015-05-06
Fetch random locations by shuffled id, for efficiency

flibbertigibbet:
Address TODO for handling random destination fetch.  Just get the IDs to randomize first, then pull the records to return by ID.


kshepard:
+1

PR #262, opened 2015-05-06
Only return unique itineraries.

### LINKED ISSUE #256: Same route shown multiple times
image
The same route was just repeated 3 times as my options. There should only be one option displayed if there is only one reasonable route.

flibbertigibbet:
Fixes #256.
This is a workaround for opentripplanner/OpenTripPlanner#1894.

### DIFF
--- a/src/app/scripts/cac/control/cac-control-itinerary-list.js
+++ b/src/app/scripts/cac/control/cac-control-itinerary-list.js
@@ -2,7 +2,7 @@
  *  View control for the itinerary list
  *
  */
-CAC.Control.ItineraryList = (function ($, Handlebars, Utils) {
+CAC.Control.ItineraryList = (function (_, $, Handlebars, Utils) {
 
     'use strict';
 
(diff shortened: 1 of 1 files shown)


src/app/scripts/cac/control/cac-control-itinerary-list.js
```diff
@@ -52,7 +52,21 @@ CAC.Control.ItineraryList = (function ($, Handlebars, Utils) {
...
+        // Due to issue: https://github.com/opentripplanner/OpenTripPlanner/issues/1894
+        // itineraries with transit + (bike/walk) can return 3 identical itineraries if only
+        // bike/walk used, and not transit.
+        itineraries = [];
+        if (newItineraries.length > 0) {
+            // _.findWhere returns null if list is empty, so prime list with first itinerary
+            itineraries.push(newItineraries[0]);
+        }
+        _.each(newItineraries, function(itinerary) {
+            if (_.findWhere(itineraries, itinerary) === null) {
```
kshepard: This could probably be:
`if (!_.findWhere(itineraries, itinerary))`

kshepard:
+1

src/app/scripts/cac/control/cac-control-itinerary-list.js
flibbertigibbet: Yeah, I guess it's never going to evaluate to 0. That looks better.

flibbertigibbet: Interestingly, that didn't seem to work. I'll just leave it as-is.

PR #263, opened 2015-05-07
Enforce minimum sizes for article images,

### LINKED ISSUE #231: Adjust sizes/dimensions for Django article images
Wide image
Minimum size: 1280x400 pixels

Narrow image
Minimum size: 400x600 pixels (this is good as-is)

flibbertigibbet:
and change aspect ratio expected for wide image.

Closes #231.

### DIFF
--- a/python/cac_tripplanner/cms/models.py
+++ b/python/cac_tripplanner/cms/models.py
@@ -103,7 +103,7 @@ class Article(models.Model):
     modified = models.DateTimeField(auto_now=True)
     content_type = models.CharField(max_length=4, choices=ArticleTypes.CHOICES)
     wide_image = models.ImageField(upload_to=generate_filename, null=True,
-                                   help_text='The wide image. Will be displayed at 600x300.')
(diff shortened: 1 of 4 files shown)


python/cac_tripplanner/cms/models.py
```diff
@@ -103,7 +103,7 @@ class ArticleTypes(object):
     modified = models.DateTimeField(auto_now=True)
     content_type = models.CharField(max_length=4, choices=ArticleTypes.CHOICES)
     wide_image = models.ImageField(upload_to=generate_filename, null=True,
-                                   help_text='The wide image. Will be displayed at 600x300.')
+                                   help_text='The wide image. Will be displayed at 1280x400.')
```
kshepard: This is lame, but Django will require a migration here even though the help_text isn't stored anywhere in the db.

kshepard:
+1

python/cac_tripplanner/cms/models.py
flibbertigibbet: That's odd. Good catch.

PR #264, opened 2015-05-07
Go to explore tab details from homepage destination

### LINKED ISSUE #259: Destination information
When clicking on a destination on the home page, the directions come up but that's it. There's no way now for me to get information on the destination. I think ideally clicking on the destination on the home page would bring me to a page with info on the destination and photos, a link to their website, etc. On there I would have a place I could click to get directions there. At a minimum, if it takes me directly to directions I should be able to get that pop up with a little info on it.

flibbertigibbet:
Closes #259.


kshepard:
:+1:

PR #271, opened 2015-05-07
Feature/link shortener encoding

flibbertigibbet:
* Use base 58 instead of base 64 encoding for shortened URLs.
  - The existing shortened URLs in the database will need to be dropped
* Update minor versions on some of the app pip-installed packages

I found [python-basehash](https://github.com/bnlucas/python-basehash) while working on this, which might be suitable if we want shortened URLs to always be the same for a shortened address. However, I doubt there will be many cases of the exact same trip getting shared, so I'm not sure how useful that would be.


kshepard:
+1

PR #272, opened 2015-05-07
Put in bike routes layer

flibbertigibbet:
Adds layer with bike routes (from CartoDB visualization).
This replaces the unimplemented bicycle parking layer.

### DIFF
--- a/src/app/scripts/cac/map/cac-map-overlays.js
+++ b/src/app/scripts/cac/map/cac-map-overlays.js
@@ -11,13 +11,11 @@ CAC.Map.OverlaysControl = (function ($, L) {
     }
 
     OverlaysControl.prototype.bikeShareOverlay = bikeShareOverlay;
-    OverlaysControl.prototype.bikeParkingOverlay = bikeParkingOverlay;
+    OverlaysControl.prototype.bikeRoutesOverlay = bikeRoutesOverlay;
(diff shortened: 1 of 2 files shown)


src/app/scripts/cac/map/cac-map-overlays.js
```diff
@@ -33,9 +31,14 @@ CAC.Map.OverlaysControl = (function ($, L) {
         return bikeShareFeatureGroup;
     }
 
-    // TODO: Implement
-    function bikeParkingOverlay() {
-        return L.featureGroup([]);
+    function bikeRoutesOverlay() {
+        // https://cac-tripplanner.cartodb.com/viz/501dbdc8-f4ea-11e4-8c9e-0e018d66dc29/public_map
```
kshepard: What's this one? And should it be removed?

kshepard:
+1

src/app/scripts/cac/map/cac-map-overlays.js
flibbertigibbet: That's the link to the visualization itself.  I thought it might be helpful to document it, since the viz ID isn't in the tile URL itself.

kshepard: Ah, ok, makes sense to keep it around then.

PR #273, opened 2015-05-07
Better fix for duplicate itineraries.

flibbertigibbet:
Supercedes #262.

The previous fix sometimes removed alternate routes that weren't actually duplicates, as `_.findWhere` only does a shallow object comparison, and also the alternate routes were still being mapped.

### DIFF
--- a/src/app/scripts/cac/routing/cac-routing-plans.js
+++ b/src/app/scripts/cac/routing/cac-routing-plans.js
@@ -28,7 +28,25 @@ CAC.Routing.Plans = (function($, L, moment, _, UserPreferences, Itinerary, Setti
             data: urlParams
         }).then(function(data) {
             if (data.plan) {
-                var itineraries = _(data.plan.itineraries).map(function(itinerary, i) {
+                // Ensure unique itineraries.
+                // Due to issue: https://github.com/opentripplanner/OpenTripPlanner/issues/1894
+                // itineraries with transit + (bike/walk) can return 3 identical itineraries if only
+                // bike/walk used, and not transit.
+                // TODO: remove this workaround once OTP issue resolved.
+                var lastItinerary = null;
+                var planItineraries = _.reject(data.plan.itineraries, function(itinerary) {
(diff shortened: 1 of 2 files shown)


src/app/scripts/cac/routing/cac-routing-plans.js
```diff
@@ -28,7 +28,25 @@ CAC.Routing.Plans = (function($, L, moment, _, UserPreferences, Itinerary, Setti
...
         }).then(function(data) {
             if (data.plan) {
-                var itineraries = _(data.plan.itineraries).map(function(itinerary, i) {
+                // Ensure unique itineraries.
+                // Due to issue: https://github.com/opentripplanner/OpenTripPlanner/issues/1894
+                // itineraries with transit + (bike/walk) can return 3 identical itineraries if only
+                // bike/walk used, and not transit.
+                // TODO: remove this workaround once OTP issue resolved.
+                var lastItinerary = null;
+                var planItineraries = _.reject(data.plan.itineraries, function(itinerary) {
```
kshepard: Have you taken a look at [`_.uniq`](https://lodash.com/docs#uniq)? Seems like it may work great here.

kshepard:
+1, though possibly switch to _.uniq if it's a good fit.

src/app/scripts/cac/routing/cac-routing-plans.js
flibbertigibbet: Oh, I didn't realize we were already using lodash on this project.  I'll check it out.

flibbertigibbet: Tried `_.uniq`, but it didn't seem to work (duplicates weren't filtered).  I think it may be shallow as well; I found somebody created a [lodash-deep](https://www.npmjs.com/package/lodash-deep) project that has `_.deepUniq` defined.  I think I'll leave it as-is for now. Property ordering appears to be consistent, so the naive `JSON.stringify` works fine.

PR #274, opened 2015-05-08
Feature/user alerts

flibbertigibbet:
* Add template for dismissable Bootstrap alerts in map sidebar.
* Move map templates into map template module.

This does not add an alert template for the home page, as I don't see a need for one at the moment.  The only thing to alert for there would be missing/invalid input from the address fields, which can be handled with input box styling.


flibbertigibbet:
Some test alerts:
alert_gpg_directions_list
alert_gpg_discover
alert_gpg_itinerary_list

kshepard:
+1

Issue #267, opened 2015-05-07
Edit color of map pins


WarkulwizDesign:
Can the map pin colors be edited so that they are most distinctive - esp. for the events? The gray pin for events gets lost in the map. The start destination pin is the same color as the current bike share location. I'd suggest changing the bike share locations to the cyan color since that resonates as the color for Indego.
![wda-editmappins](https://cloud.githubusercontent.com/assets/12275457/7521873/440d3b92-f4bf-11e4-86f3-ced9e72637e7.png)

jbranigan:
Cartography for pins and base maps is underway: https://github.com/azavea/cac-tripplanner/issues/156

jbranigan:
And thanks for the notes!

WarkulwizDesign:
Sorry for the repeat!!

flibbertigibbet:
Closed with #277

Issue #156, opened 2015-03-31
Cartography enhancements


jbranigan:
*   Isochrone style
*   Primary/secondary directions lines styles
*   Icons
*   Basemap options

jbranigan:
This is for the isochrone and directions lines...

Solid Line:
stroke-linejoin="round" stroke-linecap="round" stroke="#d02d2d" stroke-opacity=".75" stroke-width="5" fill="none"

Dash Line:
stroke-linejoin="round" stroke-linecap="round" stroke="#2c7fb8" stroke-opacity=".75" stroke-width="5" stroke-dasharray="5, 8" fill="none"

Isochrone Blob
stroke-linejoin="round" stroke-linecap="round" fill-rule="evenodd" stroke="#5c2482" stroke-opacity="0.4" stroke-width="2" fill="#5c2482" fill-opacity="0.3"

flibbertigibbet:
The [Leaflet GeoJSON styling options](http://leafletjs.com/reference.html#path-options) don't have a `fill-rule`, and `fill` doesn't apply for lines, but otherwise these styles should work well.

jbranigan:
Cool. I'm reopening to account for icon and basemap changes, unless you'd prefer I open new issues for those.

jbranigan:
For featured location icons, any preferences on one of these three? I'm partial to the first one.

![](http://google.github.io/material-design-icons/maps/svg/ic_beenhere_24px.svg)
![](http://google.github.io/material-design-icons/maps/svg/ic_local_attraction_24px.svg)
![](http://google.github.io/material-design-icons/maps/svg/ic_satellite_24px.svg)

Featured location icon color can be the same green we used for Indego.
Let's change Indego icons to #0083CC. That's the light blue from their logo.

Still more ideas to come, this is still WIP...

flibbertigibbet:
The [FontAwesome markers](https://github.com/lvoogdt/Leaflet.awesome-markers) we're using use png images for the markers, so we're limited in the available colors.  There's 'blue' and 'cadetblue' available; their 'blue' is pretty close.

jbranigan:
The 'blue' looks pretty good to me.

jbranigan:
Can we switch out the Stamen base maps for CartoDB's positron and dark matter? Positron would be the default. In the layer switcher, we can just call the layers "light" and "dark"

jbranigan:
All looks good

PR #277, opened 2015-05-08
Change icons and colors

flibbertigibbet:
Changes suggested by #156 and #267.

Changes featured location icon, and switches around icon colors: featured locations are green, bike share stations are blue, home is purple, and Uwishunu events are orange.

image


flibbertigibbet:
What do you think, @jbranigan?

kshepard:
+1

Issue #248, opened 2015-05-01
"Current Location" popup won't find location or close, sometimes option isn't there


CorynWolk:
This one came from another Council staff member. They were running Firefox, as am I, and I haven't had quite the same issue, so I'm not sure what's going on. In his case, when they clicked "Share location" on the browser pop-up, it didn't actually find what location he was at. But then, whether he clicked "share location" or not, when he tried to type his location into the bar, the pop-up would pop-up and prevent him from typing.

On my browser, the Share My Location option is generally working (though it's not always accurate in actually finding my location), but it doesn't always show up when I'd expect it to--sometimes it only gives me the options of typing something in or the default City Hall, or if I delete the starting point, it just says "Select an origin" rather than giving me the choice to have it use my location without typing it in. I'm not sure if that's intentional or not.

(I don't know how the backend works, so I can't tell how related these two issues are. If they're totally unrelated, I'm sorry to incorrectly group them!)

kshepard:
Closed by #314

PR #278, opened 2015-05-11
Feature/cleared typeahead

flibbertigibbet:
* Trigger event when typeahead input field cleared
* Listen to event and delete location object associated with field (do not use default or previous location)
* Handle missing locations (clear results; do not attempt to fetch results; set error class on input)
* Fix styling issues with typeahead, placeholder text, and input error class

Addresses one of the issues in #248.


kshepard:
+1

PR #282, opened 2015-05-12
Feature/handle invalid locations

flibbertigibbet:
Better handling of missing or invalid locations on map page.

Also fixes user preferences tests.


kshepard:
+1

PR #283, opened 2015-05-13
Handle variable-length shortened URLs

flibbertigibbet:
Shortened base 58-encoded trip plan URLs, although usually 22 characters, are not always that length.  This allows for a range of shortened URL lengths.


kshepard:
+1

Issue #281, opened 2015-05-12
set maxWalkDistance for bike trips


flibbertigibbet:
The behavior of `maxWalkDistance` has changed in the latest version of OpenTripPlanner.  Now, instead of only affecting total allowed walk distance to transit, it now generally means how far one is willing to travel via any mode to transit.  So, it now effectively works as "max bike distance" for biking/transit trips.

We could add a "max bike distance" dialog now for bike/transit trips (as suggested in #159), or we should at least set `maxWalkDistance` to a high default value for bike/transit trips, to avoid messages saying trip is not possible, when it is possible by biking alone.

kshepard:
The behavior changed from 0.14 to 0.15? If so, what do you think about reverting to 0.14? Seems like that would be the easiest fix.

flibbertigibbet:
The behavior in 0.14 was less desirable. Should only revert if it 0.14 does return bike-only trips for bike/transit requests (I don't think it did).  Bumping `maxWalkDistance` in that case should be simple.

PR #288, opened 2015-05-15
Allow longer bike riding when using public transit

### LINKED ISSUE #286: Palmyra Cove not working
image
Start address shouldn't be the issue, the end address is based on "get directions" for Palmyra Cove

sharph:
palmyra

Closes #286 and #281

### DIFF
--- a/src/app/scripts/cac/control/cac-control-sidebar-directions.js
+++ b/src/app/scripts/cac/control/cac-control-sidebar-directions.js
@@ -12,6 +12,9 @@ CAC.Control.SidebarDirections = (function ($, Control, BikeModeOptions, Geocoder
     // Number of millis to wait on input changes before sending directions request
     var DIRECTION_THROTTLE_MILLIS = 750;
 
+    // default maxWalk when biking (in miles)
+    var MAXWALK_BIKE = 300;
+
     var defaults = {
         selectors: {
             bikeTriangleDiv: '#directionsBikeTriangle',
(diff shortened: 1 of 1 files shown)


sharph:
One sec... I'm going to move that magic number to the top of the file or out of it.

sharph:
New commit pushed.

src/app/scripts/cac/control/cac-control-sidebar-directions.js
```diff
@@ -12,6 +12,9 @@ CAC.Control.SidebarDirections = (function ($, Control, BikeModeOptions, Geocoder
     // Number of millis to wait on input changes before sending directions request
     var DIRECTION_THROTTLE_MILLIS = 750;
 
+    // default maxWalk when biking
+    var MAXWALK_BIKE = 300;
```
flibbertigibbet: Default maxWalkDistance is greater; it is 800.  This is only .19 miles, right? Also would be good to rename constant and/or add note regarding units/what this is in miles.

src/app/scripts/cac/control/cac-control-sidebar-directions.js
```diff
@@ -162,6 +165,9 @@ CAC.Control.SidebarDirections = (function ($, Control, BikeModeOptions, Geocoder
             $.extend(otpOptions, {optimize: 'TRIANGLE'},
                      bikeModeOptions.options.bikeTriangle[bikeTriangle]);
             UserPreferences.setPreference('bikeTriangle', bikeTriangle);
+
+            // allow longer bike riding when using public transit
+            $.extend(otpOptions, { maxWalkDistance: MAXWALK_BIKE * METERS_PER_MILE });
```
flibbertigibbet: Also need to update UserPreferences `maxWalk` here.

src/app/scripts/cac/control/cac-control-sidebar-directions.js
sharph: This is 300 miles. I'll add a comment.

src/app/scripts/cac/control/cac-control-sidebar-directions.js
sharph: I'm not sure it should be updated -- just sent with the OTP request. Otherwise when you switch back to walking mode, won't the user selectable preference be overridden with the bike default?

sharph:
Comment about miles added

src/app/scripts/cac/control/cac-control-sidebar-directions.js
flibbertigibbet: No; if the mode switches to non-bike, the max walk preference gets updated or unset (see block below). I guess it's fine for now since we don't have a user control for changing it for bike yet.

PR #296, opened 2015-05-15
Use command ansible module instead of pip for majorkirby install

sharph:
Ansible's pip module installs to /tmp/src for some reason so call the
command directly instead


sharph:
Closes #289

kshepard:
+1, weird!

Issue #295, opened 2015-05-15
Use S3 URL for media


jbranigan:
Instead of the local cms/media location.
Affects home page, content pages.

kshepard:
Closed via #300

Issue #290, opened 2015-05-15
Admin - editing destination throws 500


jbranigan:
Got a 500 error on trying to save a location change.

kshepard:
Closed via #300

PR #300, opened 2015-05-15
Various admin fixes

kshepard:
Fixes the following issues: 
#290 -- destination image validation was using the old validate_image signature
#295 -- needed to remove the media_url usage in order to work properly with s3
#231 -- updated article image width to 1440


python/cac_tripplanner/cms/models.py
```diff
@@ -30,15 +30,9 @@ def random(self):
         Note: This does not return a queryset so cannot be chained,
```
flibbertigibbet: I suppose this note doesn't apply any more, since we are using a queryset.

flibbertigibbet:
+1

python/cac_tripplanner/cms/models.py
kshepard: Cool, I'll update it.

PR #306, opened 2015-05-17
Mobile testing and more styling fixes

mtedeschi:


python/cac_tripplanner/templates/home.html
```diff
@@ -56,7 +56,7 @@ <h3 class="text-center hidden-xs">There are 2 easy ways to use the map:</h3>
                 <input id="directionsTo" class="typeahead" type="search"
                     data-typeahead-key="to" placeholder="City Hall" />
                 <span class="text">using</span>
-                <div id="directionsMode" class="btn-toolbar transit-types-container">
+                <div id="directionlgode" class="btn-toolbar transit-types-container">
```
flibbertigibbet: Was this `id` change intentional?  This will break the javascript code that uses the `id`.

kshepard: Sounds like an overzealous find/replace of sm (small) to lg (large). Looks safe to revert.

PR #308, opened 2015-05-18
Fix origin set in discover tab.

flibbertigibbet:
Fix two separate issues causing missing origin when getting directions from featured location, both of which had been masked by use of default locations.

* If using current location by default on page load, set location object
* Fix originText not being set when address selected in typeahead (it never was).


kshepard:
+1. Also, to make sure we don't duplicate effort: no need for further "current location" fixes, since I have a branch in the works that's removing that logic.

PR #313, opened 2015-05-18
Feature/add directions page link

flibbertigibbet:
Display modal with shortened link to directions page.

Also fixes issue where directions list always set up twice (so events would trigger twice, including sharing actions.)


kshepard:
+1

flibbertigibbet:
image

image

PR #322, opened 2015-05-20
Various fixes related to setting/loading locations

kshepard:
The major change here is unifying the `from` and `origin` settings.


flibbertigibbet:
+1

PR #327, opened 2015-05-20
Load CartoDB library from CDN

flibbertigibbet:
Load cartodb.js from CDN, rather than compiling it into minified vendor.js.


kshepard:
+1

PR #328, opened 2015-05-20
Fix some naming problems ('destination' vs. 'to')

kshepard:
Going to merge this in order to do a deploy. I'll address any comments in the morning.


src/app/scripts/cac/pages/cac-pages-home.js
```diff
@@ -247,9 +247,11 @@ CAC.Pages.Home = (function ($, BikeModeOptions, Templates, UserPreferences) {
 
         var $input;
         var $other;
+        var prefKey;
 
         // Make sure to keep the directionsFrom origin in sync with the explore origin
         if (key === 'origin' || key === 'from') {
```
flibbertigibbet: `from` and `origin` are the same now, right?

kshepard: They are, but there still need to be discrete typeahead elements on this page. I didn't change the templates at all, so there are typeaheads for both a directionsFrom and and exploreOrigin. It may make sense to modify the typeahead keys in the template to get rid of this 'from', however it is currently working as expected.

PR #330, opened 2015-05-21
Change leaflet references.

flibbertigibbet:
Use `cartodb.L` instead of `L`, except when referencing libraries built against leaflet (Polyline and AwesomeMarkers).


kshepard:
+1

PR #331, opened 2015-05-21
Feature/mobile directions link

flibbertigibbet:
Also disables broken test.

### DIFF
--- a/src/app/scripts/cac/pages/cac-pages-home.js
+++ b/src/app/scripts/cac/pages/cac-pages-home.js
@@ -226,6 +226,23 @@ CAC.Pages.Home = (function ($, BikeModeOptions, Templates, UserPreferences) {
         }).then(function(data) {
             if (data.destinations && data.destinations.length) {
                 var destination = data.destinations[0];
+
+                // check user agent to see if mobile device; if so, redirect to Google Maps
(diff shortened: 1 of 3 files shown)


kshepard:
+1, cool, works on my Android.

kshepard:
And does it definitely work on iOS devices? That StackOverflow link seems to mention the map:// protocol a bunch, which I don't see here.

flibbertigibbet:
I decided not to go with the Apple maps map:// protocol, partly because the code to check for iOS device and version looked pretty brittle, partly because the mobile Google Maps site works just as well on iOS :)

flibbertigibbet:
Tested in Chrome emulation.

jbranigan:
Worked on iOS, I think. This action was supposed to be associated with the "Get Directions" button at the bottom of the featured location details. It happened for me when I clicked on the item on the home page, never actually seeing the details.

flibbertigibbet:
Oh, I was under the impression that the user was never to be allowed to get to the map page if on a mobile device.

jbranigan:
Mike's got all of the map and directions picker stuff hidden - or it will be. Also, it's affecting use on iPad - we should have normal behavior there.

flibbertigibbet:
Tablets should be allowed desktop site access?

jbranigan:
Yes, the scope is for desktops and tablets, just not phones. Mike is working on conditional displays based on (I think) a max width of 767px.

PR #338, opened 2015-05-26
Tablet styling + small design fixes

mtedeschi:


flibbertigibbet:
Clearing input on the 'discover' tab is still weird; it results in the location and search icons overlaying one another.  Otherwise looks good.

PR #339, opened 2015-05-26
Redirect mobile users from feature details directions.

flibbertigibbet:
Redirect non-tablet mobile devices to Google directions from featured location details "Get Directions" button.


kshepard:
+1, haven't tested it on various devices, but the code looks good. If you haven't already, I'd check this against the recent open styling PR to make sure that pixel width boundary still applies.

PR #342, opened 2015-05-27
Increase default number of gunicorn workers

flibbertigibbet:
In doing load testing, I discovered we only have one gunicorn worker, and that this presents a bottleneck.  According to [the docs](http://gunicorn-docs.readthedocs.org/en/latest/design.html), (2 * # of CPUs) + 1 is a good number, so that would put us at five for all environments.  Five seems to work well in load testing, too.


kshepard:
+1

PR #347, opened 2015-05-28
Increase max walk distance to 2 miles.

flibbertigibbet:
Otherwise Palmyra Cove Nature Park appears inaccessible with defaults, as getting there with walk+transit requires over 1.5 mi of walking.

### DIFF
--- a/python/cac_tripplanner/templates/map.html
+++ b/python/cac_tripplanner/templates/map.html
@@ -89,7 +89,7 @@
                         <h6>Maximum walking distance</h6>
                         <div class="input-group">
                             <input type="number" class="form-control isochrone-input"
-                                placeholder="0.5" value="0.5" min="0.1" step="0.1" >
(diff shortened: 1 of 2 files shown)


kshepard:
Don't know if it's a comparable situation, but @jbranigan ran into this problem with other featured locations and fixed it by moving the point slightly, which is fine since a lot of these locations are large and have different entrances. If that doesn't make sense in this case, and we're good with changing the default walk distance to 2 miles, then this looks good to me. +1

flibbertigibbet:
There's no transit across the Palmyra bridge, and bikes must be walked on it, so it's not about the entrance point for that location.  I think we may be better off with a large max walk distance anyways, since users may not think to go check extra settings if they can't get a route somewhere.

kshepard:
Cool, I agree with that logic. If the route shows too much walking using the default setting, then the user will see it and can try lowering it.

PR #352, opened 2015-05-29
Fixes #349.

### LINKED ISSUE #349: Not able to move the map once directions selected
I just tested this and was able to move the map on my browser, chrome. But see comment below from Erik at SEPTA:

"This is very cool, Nick, very sleek interface. I particularly like the integration with the BikeShare locations and real-time docking info.

One technical problem I’m having is after selecting a set of parameters in the “Get Directions” field, I’m not able to click on the map and move it. Could be an issue with my browser, Safari. Not sure how I can “print screen” this problem per your instructions below since it’s not a static issue, but I thought I’d send it your way anyway.

I’ll let you know if I see any other issues.

Erik"

flibbertigibbet:
Remove second pan/zoom of map when directions retrieved. Do not pan to moved directions marker.

Also fixes issue with directions fit sometimes seemingly to not take sidebar into account.


kshepard:
+1

PR #354, opened 2015-05-29
Feature/auto scaling

flibbertigibbet:


kshepard:
+1, cool! Haven't tested it, but I'd love to watch it work its magic under some load.

Issue #348, opened 2015-05-28
directions starting in a weird place


nickrogers:
typed in "135 south 19th street, philadelphia, pa" and had directions show me somewhere in the Northeast. I didn't type in the zipcode, but if the site isn't sure then maybe it should prompt me for a zipcode? I would think City and State would be good enough though.
![image](https://cloud.githubusercontent.com/assets/12200440/7867436/185d2148-0544-11e5-91a4-765a4ecdb6f0.png)

nickrogers:
Actually, I'm just realizing that it put me in my most recent search, the Holmesburg Junction Station.

Of note, I also didn't type in the zipcode for my ending address, but it got that right.

jbranigan:
Same problem as https://github.com/azavea/cac-tripplanner/issues/336

Issue #252, opened 2015-05-04
Sometimes gives directions to different location if doesn't recognize address, or seems to remember/use previous destination instead


CorynWolk:
There've been a couple variations on this. This last time, I had tried searching directions from my location (found by network) to 36th and Lancaster Ave. The site didn't recognize "36th St and Lancaster Ave, Philadelphia" and tried giving me a few kind of random addresses before finally giving me directions to a point a half mile west on Lancaster Ave. I then went back to the homepage and tried getting directions to "49th St and Hazel Ave, Philadelphia." Instead, the site gave me directions to the same point on Lancaster Ave, but replaced the finish point with "49th St and Hazel Ave." (I clicked the magnifying glass a few times to try to get it to refresh the directions, but they stayed the same.)
![destination and directions mixup issue](https://cloud.githubusercontent.com/assets/12202740/7459213/5be831a8-f268-11e4-9226-4a774f4cca96.jpg)

flibbertigibbet:
The ESRI geocoder search service (used for the suggestions) does not recognize intersections.

jbranigan:
@CorynWolk, we spent some time last week trying to fine-tune how the location search will work. As @flibbertigibbet mentioned, this is an unsupported feature in the geocoding service. We have address, neighborhood, city, zip, featured locations, and business names, but no intersections.

jbranigan:
We'll try to handle this a bit more gracefully, though.

CorynWolk:
Thanks! So @jbranigan is there any workaround for that? It seems like if there's no way to get the website to recognize intersections, we should maybe make a note near the search bar instructing users not to use them. (Otherwise they might do what I did and assume it's a matter of punctuation or a random intersection not being in the database.)

Issue #241, opened 2015-05-01
Directions to Valley Forge not working


nickrogers:
![image](https://cloud.githubusercontent.com/assets/12200440/7432276/bd823646-eff4-11e4-9fbd-ef095ee7f3f8.png)

I asked for directions from my location (CAC office) to "Valley Forge National Historic Park, N Outer Line Dr, King Of Prussia, Pennsylvania" (which was an option when I started typing Valley Forge) and the directions took me to City Hall.

jbranigan:
Nice job on the first issue submission, @nickrogers. 

I just tried the same thing and it gave me the correct result, but initially blipped directions to City Hall before correcting. We'll dig a bit deeper.

jbranigan:
Fixed in https://github.com/azavea/cac-tripplanner/issues/362

PR #362, opened 2015-06-04
Don't allow searching without typeahead selection

### LINKED ISSUE #336: directions start at city hall even though the person put their home address in
image
"I also put in my home address but it gave me directions from city hall. It still shows the origin as my home address on the left but the map shows the origin as city hall. It seems to do this every time I ask for directions. I attached a screen shot. "

kshepard:
There have been many issues reported that are caused by the user
inputting an address, then either tabbing or clicking away from
the input, thus bypassing the typeahead selection. This commit
ensures that a typeahead value was selected, and will warn the
user otherwise (via red-bordered input box).

Closes #336, #252, #241, #348


flibbertigibbet:
+1

PR #377, opened 2015-06-12
Feature/issue 374

CloudNiner:
That header still needs fixed, but here's a screenshot with the map:
cac-directions-map

### DIFF
--- a/src/app/scripts/cac/pages/cac-pages-directions.js
+++ b/src/app/scripts/cac/pages/cac-pages-directions.js
@@ -1,11 +1,22 @@
 CAC.Pages.Directions = (function ($, _, DirectionsList, Itinerary, Settings, Utils) {
     'use strict';
 
+    var center = [39.95, -75.1667];
+    var zoom = 12;
+
+    // TODO: Remove duplicate from cac-map-control.js
+    var cartodbAttribution = [
+        '&copy; <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors, ',
+        '&copy; <a href="http://cartodb.com/attributions">CartoDB</a>'
+    ].join('');
+
     var defaults = {
         selectors: {
             directionsContainer: '.directions-list'
         }
     };
+
+    var map = null;
 
     function Directions(options) {
         this.options = $.extend({}, defaults, options);
@@ -31,6 +42,8 @@ CAC.Pages.Directions = (function ($, _, DirectionsList, Itinerary, Settings, Uti
             }
         });
 
+        loadMap();
+
         $.ajax({
             url: Settings.routingUrl,
             type: 'GET',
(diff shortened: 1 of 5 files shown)


src/app/scripts/cac/pages/cac-pages-directions.js
```diff
@@ -40,10 +51,63 @@ CAC.Pages.Directions = (function ($, _, DirectionsList, Itinerary, Settings, Uti
...
+            });
+            basemap.addTo(map);
+        }
+
+        function setMapItinerary(itinerary) {
+            map.fitBounds(itinerary.geojson.getBounds(), {
+                padding: [30,30]
+            });
+            itinerary.geojson.addTo(map);
+            console.log(itinerary);
```
flibbertigibbet: Log statement here.

src/app/scripts/cac/pages/cac-pages-directions.js
```diff
@@ -40,10 +51,63 @@ CAC.Pages.Directions = (function ($, _, DirectionsList, Itinerary, Settings, Uti
...
+                attribution: cartodbAttribution
+            });
+            basemap.addTo(map);
+        }
+
+        function setMapItinerary(itinerary) {
+            map.fitBounds(itinerary.geojson.getBounds(), {
+                padding: [30,30]
+            });
+            itinerary.geojson.addTo(map);
```
flibbertigibbet: Should fit to the geojson bounds here, instead of setting to predefined center/zoom above.

flibbertigibbet:
Might be a good idea to disable scroll/zoom interactivity on the map div, so mobile users can get down to the directions.

src/app/scripts/cac/pages/cac-pages-directions.js
```diff
@@ -1,12 +1,21 @@
 CAC.Pages.Directions = (function ($, _, DirectionsList, Itinerary, Settings, Utils) {
     'use strict';
 
+    var center = [39.95, -75.1667];
+    var zoom = 12;
+    var cartodbAttribution = [
```
flibbertigibbet: Should probably comment a TODO here about the stuff duplicated from the map controller.

flibbertigibbet:
+1

src/app/scripts/cac/pages/cac-pages-directions.js
```diff
@@ -40,10 +51,63 @@ CAC.Pages.Directions = (function ($, _, DirectionsList, Itinerary, Settings, Uti
...
             var params = data.requestParameters;
             var itinerary = new Itinerary(itineraries[itineraryIndex], itineraryIndex, params);
+            setMapItinerary(itinerary);
             directionsListControl.setItinerary(itinerary);
         }, function (error) {
             console.log('error: ', error);
         });
+
+        function loadMap() {
+            var mapOptions = {
```
CloudNiner: All map interactivity should be disabled here already. I don't think I missed anything?

flibbertigibbet: Oh cool, missed that.

jbranigan:
If you get directions from here to Conshohocken and the map isn't interactive, how are you going to figure out where you missed that right turn? 

I originally thought a relatively short (250px-ish) map would be small enough so that mobile users would still be able to touch non-map sections of the page for scrolling.

flibbertigibbet:
A mobile map smaller than full screen size is pretty much unusable. Interaction with it is likely to be unintentional.

CloudNiner:
Hmm, my misunderstanding. I thought it was decided that this would be static during our last conversation. I do think it would be difficult with the map at its current size to scroll past it without it being very annoying, but I can revert to interactive if you'd prefer.

jbranigan:
The whole mobile side of this site is to just try to help the users out a bit. It was specifically precluded in the scope. I figured a small map I can zoom in on is better than nothing. A static map would be useful for very short trips only. Neither are really useful, but I think some interaction would be preferable.

PR #378, opened 2015-06-12
Closes #360: Add transit departure times and step distances to directions

### LINKED ISSUE #360: html language showing up in directions, and time suggestion
image

Also, we have a start and end time of day, but not for the bus pickup. Consider putting a time down for any public transportation pickups/dropoffs in the trip?

CloudNiner:
directions-sidebar


flibbertigibbet:
+1

PR #383, opened 2015-06-15
Fix error when end_date is missing

flibbertigibbet:
Noticed the following error in the server logs:
```
Jun 15 06:44:42 ip-10-0-1-25 event-feed.log:      if end_date < self.now or publication_date is None or end_date is None:
Jun 15 06:44:42 ip-10-0-1-25 event-feed.log:  TypeError: can't compare datetime.datetime to NoneType
```


CloudNiner:
+1

PR #398, opened 2015-06-23
Unset destination after loading it in explore tab.

### LINKED ISSUE #397: Explore nearby locations gave me one destination already selected (and not the closest one), rather than the list
image

image

flibbertigibbet:
Fixes #397.
Caused by switch in preference key in #322.


kshepard:
+1

PR #399, opened 2015-06-23
Highlight and pan to destination selected on home page.

### LINKED ISSUE #340: Last searched travel shed showing up when clicking on destination
From the home page I clicked on Palmyra Cove, and the screen it took me to had a 30 minute travel shed from the Holmesburg Junction Septa Station, which I believe was the last place I had searched. I think a travel shed form the destination might be appropriate - to show the area around that destination that is X minutes away. Kind of a reverse travel shed.

image

flibbertigibbet:
Closes #340.
To function, must not pan/zoom to fit isochrone when it returns, or map with not pan to selected destination.


kshepard:
+1

PR #406, opened 2015-06-25
Only load jQuery once in production.

flibbertigibbet:
Fixes issue with jQuery/JotForms prototype conflict in production.
jQuery was loaded twice, once from bower, once from CDN.


kshepard:
+1

Issue #417, opened 2015-06-29
Featured locations highlighted as in error


flibbertigibbet:
Selecting a featured location from the typeahead results in the input being marked as in error. Doing so from the map page will result in both correct action (directions or isochrone) as well as error highlight.

PR #418, opened 2015-06-29
Set last selected typeahead value for featured locations.

### LINKED ISSUE #413: Destination directions are routing from City Hall
When using the 'Explore Nearby Locations' from the hompenage and selecting one of the GPG destinations when selecting in the "from" area, the map is defaulting to city hall as the starting point. In both instances, 'Bartram's Garden' and 'Tulpehakin Nature Center' were selected from one of the populated items that appears as you type.

flibbertigibbet:
Fixes #413 and #417.


kshepard:
+1

PR #425, opened 2015-06-30
Use Azavea fork of OTP.

flibbertigibbet:
Has shapefile isochrone endpoint removed, to fix itermittent deploy issues with isochrone endpoint.


kshepard:
+1

PR #434, opened 2015-07-02
Fix destination unset.

### LINKED ISSUE #422: Address not working
image
Not an intersection. This address was selected from the list!

### LINKED ISSUE #424: Previous destination still highlighted
Created new directions (to Palmyra Cove, but I just typed in Palmyra cove instead of selecting it on the map) and my previous destination, Bartram's Garden is still in the highlighted green.
image

flibbertigibbet:
Destination unset was actually setting destination to the string "null" (introduced by #362).

Fixes #422 and fixes #424.


kshepard:
Ah, good catch. +1

PR #435, opened 2015-07-02
Highlight share buttons on interaction.

### LINKED ISSUE #426: Create active state for sharing buttons
Is there any way to create an active state for the sharing buttons that is more obvious? Since the mail is opening in a new tab for some users, having a change in color to the icon when selected may help them notice that their selection has been made and the current light dark shade is not that noticeable.
150630-activestate-sharingbuttons

flibbertigibbet:
Closes #426.


flibbertigibbet:
Trivial, merging

PR #446, opened 2015-09-21
Set and retrieve origin/destination text for shared directions.

### LINKED ISSUE #444: Starting location wrong in shared directions
I used the share function to get a link for the directions I created. It works fine on my computer, but on anyone else's (I attached my phone as an example) it shows it starting at City Hall, which is wrong, followed by correct directions. 

Sidenote, I had mentioned this previously but that "osmnode" thing is another weird part of the shared directions that happens frequently.

Correct: 
image

Incorrect:
image

flibbertigibbet:
Fixes #444.


kshepard:
+1

PR #447, opened 2015-12-07
Fix bike share data parsing.

flibbertigibbet:
Enpoint now has proper content type set for JSON response.


flibbertigibbet:
Trivial, merging.

PR #448, opened 2015-12-29
Use CORS proxy with new bike share endpoint

flibbertigibbet:
City API endpoint now redirects to Indego site, which disallows CORS.

http://crossorigin.me does not work with the new endpoint due to [this issue](https://github.com/technoboy10/crossorigin.me/issues/26).


CloudNiner:
+1

PR #449, opened 2016-01-14
Remove CORS proxy on bikeshare endpoint.

flibbertigibbet:
Indego has enabled CORS, and response type is JSON again.


flibbertigibbet:
Trivial, merging.

PR #450, opened 2016-01-15
Use new role to download OTP jar

flibbertigibbet:


flibbertigibbet:
Pretty trivial and tested locally; merging for deploy.

kshepard:
+1

PR #456, opened 2016-03-16
Publish and unpublish articles by setting publication date.

### LINKED ISSUE #455: Unable to unpublish articles
Logged in as an admin, we are getting error messages when trying to unpublish old articles.

flibbertigibbet:
Fixes #455.

`published` is a property method on the article model that returns true if the publication date is in the past.


ddohler:
+1

Issue #459, opened 2016-05-23
Uwishunu events no longer showing up on map


nickrogers:
The Uwishunu event feed is no longer populating on the map.

jbranigan:
It looks like the feed is currently empty. We'll dig into this a bit more. The feed is currently at:
http://www.uwishunu.com/feed/google (which downloads a basic XML RSS file, with no events)

PR #460, opened 2016-06-01
Remove nearby events layer from map.

flibbertigibbet:
To be added back when Uwishunu feed with event coordinates returns; see #459.


flibbertigibbet:
Trivial; merging for deploy.

PR #461, opened 2016-06-01
Feature/role updates

flibbertigibbet:


flibbertigibbet:
Trivial, merging for deploy.

PR #462, opened 2016-06-20
Feature/fix dev js build

flibbertigibbet:
Fixes the development JS build by setting the `debug` variable on the Django template context.

This variable should be set by the template context processor, but isn't, and perhaps never was, as the original view also explicitly set `debug` on the context.

Also updates JS build library minor versions.


KlaasH:
My box provisioned and came up fine and I can confirm that both the `gulp-production` and `gulp-development` tasks are working.

What other tasks are we expecting to be working? For me right now, `gulp-lint` fails with actual linting errors, `gulp-test` fails with this:
```
PhantomJS 1.9.8 (Linux 0.0.0) ERROR
  TypeError: 'undefined' is not an object (evaluating 'L.AwesomeMarkers.icon')
  at /opt/app/src/app/scripts/cac/map/cac-map-control.js:9
```
And `gulp-test-dev` just sits there, seemingly waiting for something that never happens.

flibbertigibbet:
This is just to get `gulp-development` working again so we can deal with the other tasks this sprint without having to run `gulp-production`.

The linter issues mysteriously appeared on extant code and pinned JS library versions a month or two ago; the AwesomeMarkers issue seems to have surfaced more recently. `gulp-test-dev` has not worked in quite some time, and it doesn't run on Jenkins.

We will have tasks later for upgrading libraries and other technical-debt related things.

KlaasH:
Cool. :+1:

PR #464, opened 2016-06-29
URL for map directions

KlaasH:
Adds a URL router that only does one thing: reads and writes directions URLs in the map view. Since all of the interaction is built around local storage, this also works through local storage--updating stored values from the URL when a matching URL is loaded and keeping the URL in sync by updating it from local storage when the parameters change.

I'm not really clear on how these URLs are to be used and how this interacts with the existing share button.  The share button produces a (printable) directions page for a specific itinerary; the URLs I've added point to the map view with a specific set of options selected but no itinerary chosen.  So it's less specific but it enables exploration and modification whereas the existing directions link doesn't.

In the course of looking at map-view directions and share-link directions I ran into some weird behavior, hence the second commit making itineraries fully (rather than just almost-fully) independent of local storage.

Note: the routing package I found, [Navigo](https://github.com/krasimir/navigo), is small but seems useful.  Unfortunately bower versioning doesn't work because the author hasn't seen fit to create any tags in the repo, even though he's got a version number in the `package.json`.  Not sure what the best way to deal with that is.  It's active, so we could just ask him to start tagging versions...

### DIFF
--- a/src/app/scripts/cac/control/cac-control-sidebar-explore.js
+++ b/src/app/scripts/cac/control/cac-control-sidebar-explore.js
@@ -330,8 +330,7 @@ CAC.Control.SidebarExplore = (function (_, $, BikeModeOptions, Geocoder, MapTemp
 
         var dest = [destination.point.coordinates[1], destination.point.coordinates[0]];
 
-        Routing.planTrip(exploreLatLng, dest, date, otpOptions)
-            .then(function (itineraries) {
+        Routing.planTrip(exploreLatLng, dest, date, otpOptions).then(function (itineraries) {
             if (itineraries.length) {
                 var distance = itineraries[0].formattedDuration;
                 destination.formattedDuration = distance;
--- a/python/cac_tripplanner/templates/base.html
+++ b/python/cac_tripplanner/templates/base.html
@@ -97,6 +97,8 @@
     <script src="{% static 'scripts/vendor/moment-duration-format.js' %}"></script>
     <script src="{% static 'scripts/vendor/bootstrap-datetimepicker.min.js' %}"></script>
     <script src="{% static 'scripts/vendor/bootstrap-select.js' %}"></script>
+    <script src="{% static 'scripts/vendor/navigo.js' %}"></script>
+
 
     <script src="{% static 'scripts/main/cac/cac.js' %}"></script>
     <script src="{% static 'scripts/main/cac/utils.js' %}"></script>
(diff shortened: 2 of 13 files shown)


src/app/scripts/cac/control/cac-control-sidebar-explore.js
```diff
@@ -332,12 +332,12 @@ CAC.Control.SidebarExplore = (function (_, $, BikeModeOptions, Geocoder, MapTemp
 
         Routing.planTrip(exploreLatLng, dest, date, otpOptions)
             .then(function (itineraries) {
-            if (itineraries.length) {
-                var distance = itineraries[0].formattedDuration;
-                destination.formattedDuration = distance;
-                setDestinationMinutesText(distance);
-            }
-        });
+                if (itineraries.length) {
```
flibbertigibbet: Doubled indentation here not consistent with our usage.

flibbertigibbet:
The live-updating URL router solution wasn't what I was expecting, but it seems to work quite smoothly. Nice!

:+1:

flibbertigibbet:
Oh wait, I'm seeing a blank directions page from a shared URL.

flibbertigibbet:
Oh wait, it seems there's an issue with the production gulp build.

flibbertigibbet:
Nevermind, production JS build is fine.

KlaasH:
So yeah, I think I was approaching this as "adding some routing" so I went looking for a router.  Then I ended up using its API to do a couple things that aren't really routing.  I could rip it out and put the new behaviors (load URL params into local storage, compose URL from local storage, sync the displayed URL*) in the page and sidebar objects.

I don't think it would be hard, and it would reduce the number of moving pieces.  I guess I'm not totally sure how to weigh the pros and cons.

* Or not sync the displayed URL.  A share button is possibly cleaner and less error prone.

flibbertigibbet:
Yeah, I was initially thinking of this as something where we would only generate the full URL when needed for sharing, then just parse out any URL parameters on load, strip them out, and use their to update local storage. That would mean not having to live-update the URL or needing to keep both local storage and the URL in sync, but also that the user couldn't share a permalink by simply copying from the URL bar.

For URL sharing purposes, we probably won't want to populate all options. For example, we will want the time to default to now, and at least sometimes default to leaving the origin blank. Not sure if that influences the utility of the router.

I think this is great for now, and if we have reason to revisit using the router later, it seems that it should be easy enough to take it out.

PR #465, opened 2016-06-30
URLs for "Discover Places" list and locations

KlaasH:
Gives the "Discover Places" (aka "explore") tab the same treatment as the directions tab got in PR #464: parameterized URLs pointing to `map/places?` will now load a location list or a specific location, and changing your choices in the tab updates the URL.

Also some incidental fixes and little refactors that came up in the course of getting it working.

The questions of what parameters to include (all probably isn't the right answer. There might not be a single right answer for all users/use cases) and whether to always show it in the address bar (nice to have a bookmarkable/shareable link without having to find the right button, but then again the full URL is so long as to be not really shareable) are related, and also related to the question of how the directions URLs interact with the social media sharing feature on directions.

My current hypothesis is that there should be one social media sharing button that's available on directions, location list, and individual locations and that if we want to give users choices about what parameters to encode in the URL (where omitted parameters would fall back to defaults or local storage for the person who follows the link) the modal that the social media button produces should provide for those options.

### DIFF
--- a/python/cac_tripplanner/templates/home.html
+++ b/python/cac_tripplanner/templates/home.html
@@ -95,51 +95,33 @@
                 <div class="sk-bounce3"></div>
             </div>
             <div class="row destinations">
+                {% for destination in destinations|slice:":2" %}
+                    <div class="col-sm-6 col-md-4">
+                        <a class="block block-destination" data-destination-id="{{ destination.id }}" href="#">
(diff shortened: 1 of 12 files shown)


flibbertigibbet:
:+1: 

Might want to look into how values not set explicitly are handled in the next task, as with the `undefined` `placeID` URL parameter when viewing the list (it still behaves as expected).

PR #466, opened 2016-07-05
Feature/blank origin or destination links

KlaasH:
Main item: showing map markers for one-sided destinations.  Formerly it was both-or-nothing, so loading a link to directions with only an origin or a destination would show a blank map.  Now it will draw the marker it has and zoom to it.

[Example](http://localhost:8024/map/directions?mode=TRANSIT%2CWALK&destination=39.9611421%2C-75.1546293&destinationText=N%2010th%20St%20%26%20Nectarine%20St%2C%20Philadelphia%2C%20Pennsylvania%2C%2019123)

Other changes made on the way:
- Changed how the map and map page initialize so that it only triggers full initialization for the sidebar that's active, rather than doing the whole process for both.  This fixed some unwelcome interactions.
- Changed the URL encoding to not include undefined local preferences in the URL and to not cause local preferences to be initialized to default values if they're not set.
- Removed some parameters from the URL, on the theory that shorter URLs might be better and on the off-chance that someone has overridden their own bike triangle or max walking distance, it wouldn't necessarily be polite for shared URLs to force it to a different value.  (Not totally sure what's best for this situation. Nor sure it matters too much. "Include in URL if different from default value" would be another possible approach, maybe the optimal one.)

### DIFF
--- a/src/app/scripts/cac/control/cac-control-sidebar-directions.js
+++ b/src/app/scripts/cac/control/cac-control-sidebar-directions.js
@@ -59,8 +59,6 @@ CAC.Control.SidebarDirections = (function (_, $, Control, BikeModeOptions, Geoco
     var typeaheadDest = null;
     var typeaheadOrigin = null;
 
-    var initialLoad = true;
-
     function SidebarDirectionsControl(params) {
         options = $.extend({}, defaults, params);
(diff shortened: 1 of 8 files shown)


flibbertigibbet:
:+1:

PR #471, opened 2016-07-07
Fix second-try geocoding when no address comes back

KlaasH:
If the geocoder search doesn't get a location with a street address on
the first try, it splits on commas to see if there's a place name
preceding a street address, and tries the geocode again if it looks
like there is.  If it gets a location with no street address, we want
to return that anyway, which it was doing properly for the first try
but not for the retry.  There was also an edge case in which the search
could return nothing.  This rearranges it to do the same thing but always
return a location if we got one (and always return null if we didn't).

You can try it out with "Wells Fargo Center, 3601 S Broad St Philadelphia, PA 19148"


flibbertigibbet:
:+1: nice

PR #473, opened 2016-07-08
Override user storage with blank values in URLs

KlaasH:
The URL reading and encoding functions have been deferring to local
storage more than we want them to. This makes it so that:
- all fields given to buildUrlParamsFromPrefs() will be included in the
  URL, with empty values if the local storage value is empty.
- setPrefsFromUrl(), upon receiving an empty origin or destination, will
  set the corresponding local storage value to undefined (rather than
  blindly trying to build a feature from it and causing an error...)

These should work now, and show the same thing whether you have prior local storage values or not:
[Origin-only directions from Wells Fargo Center](http://localhost:8024/map/directions?origin=39.9003861%2C-75.1721269&originText=Wells%20Fargo%20Center&mode=TRANSIT%2CWALK&destination=&destinationText=)

[Destination-only directions to Wells Fargo Center](http://localhost:8024/map/directions?origin=&originText=&mode=TRANSIT%2CWALK&destination=39.9003861%2C-75.1721269&destinationText=Wells%20Fargo%20Center)


flibbertigibbet:
:+1:

PR #474, opened 2016-07-08
Feature/fix getbounds error

### LINKED ISSUE #470: Zooming to marker for one-sided directions fails in production
This link, for instance: http://gophillygo.org/map/directions?origin=39.9052591%2C-75.1735046&originText=3601%20S%20Broad%20St%2C%20Philadelphia%2C%20Pennsylvania%2C%2019148&mode=TRANSIT%2CWALK
loads the directions page with the address for the Wells Fargo Center filled in and the marker shown on the map.  What it also does in development but fails to do in production is zoom and center on the marker.  Which is unfortunate, since the marker is outside the default map view.  The error is "vendor.js:27 Uncaught TypeError: e.getBounds is not a function".

It's coming from https://github.com/azavea/cac-tripplanner/blob/develop/src/app/scripts/cac/map/cac-map-control.js#L533, but since I can't reproduce it in development it's hard to track down exactly what 'e' is and why it's being asked for bounds when it doesn't know how to get them.

flibbertigibbet:
Fixes #470 by only attempting to zoom to fit all markers if more than one marker loaded. Also sorts out build issues with loading cartodb.js and properly excluding Leaflet (which is included in cartodb.js).

Also adds more graceful handling in case of the isochrone geojson not loading properly by allowing the locations list to populate.

### DIFF
--- a/python/cac_tripplanner/templates/base.html
+++ b/python/cac_tripplanner/templates/base.html
@@ -30,7 +30,7 @@
 
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css">
     <link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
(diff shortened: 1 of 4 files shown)


KlaasH:
Nice. :+1:

PR #475, opened 2016-07-08
Modify SEPTA rail warning message

flibbertigibbet:
Rail GTFS has been updated, so remove warning part about departure times in app.


flibbertigibbet:
Trivial, merging.

PR #476, opened 2016-07-11
Install wpadmin manually to work around pip issue

KlaasH:
Ansible's pip module seems to be ignoring the 'editable' option, and
setting it to True by default for packages installed from git links.
Which puts a .egg-link file in /usr/local/lib/python2.7/dist-packages and
installs the actual package in /tmp.  Which means it's gone the next time
you reload the machine.

This works around the issue by just directly running a pip command that
does the right thing.


flibbertigibbet:
:+1:

PR #478, opened 2016-07-13
Upgrade OTP to 0.20.0

KlaasH:
Upgrades OpenTripPlanner to version 0.20.0, which required upgrading Java to major version 8, which required upgrading the VM image to Ubuntu 16.04, which required switching from upstart to systemd.

Goes with a forthcoming PR for https://github.com/azavea/ansible-opentripplanner/tree/feature/upgrade-to-OTP-0.20.0, which you'll probably need to check out and copy over top of `deployment/ansible/roles/azavea.opentripplanner/` (and hand-edit `deployment/ansible/roles/azavea.opentripplanner/meta/.galaxy_install_info`) to use.  At least, that's what I've been doing.

PR #480, opened 2016-07-15
Upgrade OTP to 0.20.0 but stick with Trusty

KlaasH:
Retreat!

The move to Xenial in PR #478 is causing all sorts of annoying problems, and as @hectcastro pointed out on https://github.com/azavea/ansible-opentripplanner/pull/15 and proved in https://github.com/azavea/ansible-opentripplanner/tree/feature/hmc/upgrade-otp, OTP 0.20.0 can be run on Trusty using Oracle Java 8.

So this is PR #478 without all the Xenial-related changes (which was most of the substance, but there are still small improvements) and it seems to be fully working, for me, with Hector's branch of the OTP role.  I think we should do this instead.

### DIFF
--- a/deployment/ansible/roles/cac-tripplanner.app/meta/main.yml
+++ b/deployment/ansible/roles/cac-tripplanner.app/meta/main.yml
@@ -5,5 +5,4 @@
     - { role: "azavea.nodejs", nodejs_npm_version: "2.1.5" }
     - { role: "azavea.packer" }
     - { role: "cac-tripplanner.papertrail", when: production }
-    - { role: "azavea.ruby", ruby_development: True }
     - { role: "azavea.virtualenv" }
--- a/deployment/ansible/roles/cac-tripplanner.transitfeed/tasks/main.yml
+++ b/deployment/ansible/roles/cac-tripplanner.transitfeed/tasks/main.yml
@@ -1,4 +1,6 @@
 ---
-- name: Install transitfeed
+- name: Install pytz
   pip: name=pytz
+
+- name: Install transitfeed
   pip: name='git+https://github.com/google/transitfeed@1.2.15#egg=transitfeed'
--- a/.gitignore
+++ b/.gitignore
@@ -47,5 +47,8 @@ nosetests.xml
 # Ansible roles
 deployment/ansible/roles/azavea.*
 
+# Ansible retry files
+*.retry
+
 # AWS stack launching parameters
 deployment/default.yaml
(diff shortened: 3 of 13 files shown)


deployment/ansible/roles/cac-tripplanner.app/meta/main.yml
```diff
@@ -3,7 +3,6 @@
     - { role: "azavea.git" }
     - { role: "azavea.nginx" }
     - { role: "azavea.nodejs", nodejs_npm_version: "2.1.5" }
-    - { role: "azavea.packer" }
+    - { role: "azavea.packer", packer_version: "0.7.5" }
```
hectcastro: It can sometimes be harder to find these when things go wrong. I like the style of this better, but in practice it had led to longer debugging cycles.

deployment/ansible/roles/cac-tripplanner.transitfeed/tasks/main.yml
```diff
@@ -1,4 +1,3 @@
 ---
 - name: Install transitfeed
-  pip: name=pytz
```
flibbertigibbet: This is an important optional dependency for some aspects of feed validation to work.

flibbertigibbet:
Fresh provision from scratch worked, isochrones are still generated fine.

Looks like we will need to fix up how the start point displays; it's coming through as a number now.
image

Apart from that the upgraded OTP seems to be working fine with our existing use of the API.
:+1:

deployment/ansible/roles/cac-tripplanner.transitfeed/tasks/main.yml
KlaasH: One task with two 'pip' keys only installs the second (at least in Ansible 2.1. I guess that could have changed?), so I figured if it hasn't been actually doing anything it wasn't needed.  I'll put it back in its own task.  Do we want to peg the version?  Looks like the transitfeed project doesn't.

flibbertigibbet: Okay, good catch on the second `pip`. For timezone data it's important to always use the latest, so not pegging the version seems appropriate. The transit feeds are currently also being validated as part of the download process with the feed fetcher library, but it seems useful for the app to enforce validation as well.

KlaasH:
The change to ansible-opentripplanner is merged and I did another run of this with the merged version.  Everything appears to be working fine.  Merging at last.

PR #490, opened 2016-08-19
Override URL for openlayers.js

### LINKED ISSUE #489: Map not appearing on back end
I logged in the back end of GoPhillyGo to update a featured destination Tookany/Tacony-Frankford Watershed Partnership's location. However, the map where you can drag the pin is not appearing. Even though I can type in a new address for the location, the location of the destination on the front end of the website does not move. 

backend map not working

flibbertigibbet:
Fixes #489. The URL for openlayers.js in GeoDjango defaults to HTTP, but is configurable.


KlaasH:
:+1:

PR #492, opened 2016-08-26
Modify STS header

flibbertigibbet:
Add includeSubDomains to allow site to go on HSTS preload list:
https://hstspreload.appspot.com/?domain=gophillygo.org


hectcastro:
I initially omitted this because I wasn't sure if all of the subdomains associated with GPG were covered by HTTPS (OTP, different stack color combinations, etc.). If they aren't, this change could cause complications with modern browsers trying to interact with those domains.

flibbertigibbet:
Requests to https://otp.gophillygo.org time out. Also I don't know if our certificate would cover subdomains.

PR #493, opened 2016-08-26
Set memory in Vagrantfile when testing

sharph:
chmod 755 scripts/vagrant-halt.sh


flibbertigibbet:
+1

sharph:
@flibbertigibbet Finally got it!

flibbertigibbet:
Nice! What was the cause? A vagrant version change perhaps? Wondering why it's been working locally.

sharph:
See the diff. I suspect a local version may have actually been using one value for all three VMs. The `vagrant-up.sh` just loads the VMs in a different order.

flibbertigibbet:
The diff just changes `config` to the VM name, though `config` is still referenced elsewhere. The per-VM memory has been getting set appropriately locally.

flibbertigibbet:
Not sure how it ever worked, or why the CI setup refused to set it to anything at all instead of setting each to the same value. LGTM

PR #496, opened 2016-09-12
Feature/draggable routes

flibbertigibbet:
Make routes editable using Leaflet Draw. Does not yet re-route with the modified points, which is a separate task.

The points used for the editable route are the turn points (the start points for each step, plus the itinerary end point.) This greatly simplifies the linestring, compared to the encoded polyline used to display the route; doing so both reduces the number of drag points displayed by Leaflet Draw to a reasonable number, and better prompts the user with key points in the journey.

To start editing a route, click the pencil icon at the head of the directions detail list:
image

To end editing, click `save` or `cancel` by the Leaflet Draw icon in the top left of the map:
image

On edit save, any changed or added points are output to the console. The next task is to pass those points as waypoints in the trip plan.

Also upgrades lodash to v4 to make use of handy new diffing functions. To pick up lodash package change and added Leaflet Draw dependency, run `npm run bower-install` from `/opt/app/src` inside the `app` VM.


KlaasH:
While in edit mode, clicking anywhere but an edit point or a marker produces `cartodb.js:7 Uncaught TypeError: Cannot read property 'call' of undefined` in the console.  Is it maybe trying to remove a point regardless of whether there is one?

flibbertigibbet:
That appears to be [this issue](https://github.com/Leaflet/Leaflet.draw/issues/555) with Leaflet Draw.

src/gulpfile.js
```diff
@@ -89,7 +89,12 @@ gulp.task('minify:scripts', function() {
 });
 
 gulp.task('minify:vendor-scripts', function() {
-    return copyBowerFiles(['**/*.js', '!**/leaflet.js', '!**/leaflet-src.js', '!**/jquery.js', '!**/jquery.min.js'], [])
+    return copyBowerFiles(['**/*.js',
+                          // Exclude minified vendor scripts that also have a non-minified version.
+                          // We run our own minifier, and want to include each script only once.
+                          '!**/leaflet.draw.js', '!**/lodash.min.js', '!**/bootstrap-datetimepicker.min.js',
+                          // exclude leaflet and jquery (loaded over CDN)
+                          '!**/leaflet.js', '!**/leaflet-src.js', '!**/jquery.js', '!**/jquery.min.js'], [])
```
KlaasH: Hooray for de-obfuscation!

KlaasH:
It's hard coming to a different draggable-route implementation when one is accustomed to the fanciness of Google's.  This seems pretty good to me, though, once you get used to thinking of it more as placing waypoints than as directly dragging the route.

Some usability thoughts:
- Any way to add snapping or some kind of feedback about whether you're over an actual road when dragging a point?
- The pointer change from drag-hand to point-hand is pretty subtle (on my machine, anyway), and not super intuitive given that both primary actions are dragging.  Could we grow the pointer when over a draggable point instead?
- The tooltip is terrible--confusing, ungrammatical, and in the wrong order (the primary task is listed second).  Can we override it?

edit_tooltip

Not sure whether these things are ripe for improvement or if it makes more sense to get a basic thing working and come back around and make it nice once the pieces all exist.  The code and behavior of what's here looks good to me, so :+1:  if the latter.

flibbertigibbet:
Points 2 and 3 relate to the innards of Leaflet Draw. I think it would be worthwhile for us to fork it and make our our customized draw library down the road, once we've got some  kind of proof-of-concept working. As to the first point, if we could snap to the road, we would already have live route editing functioning, so I don't see a straightforward way to address it.

CloudNiner:
Leaflet.draw supports changing the text of any of the UI items:
https://github.com/Leaflet/Leaflet.draw#customizing-language-and-text-in-leafletdraw

Issue #497, opened 2016-09-13
Irrelevant parameters passed to OTP


flibbertigibbet:
Routing requests to OTP are also passing `fromText` and `toText`, which are only used by our front end, not by OTP. This perhaps started with the addition of JS routing. Although OTP will ignore excess parameters, they consume some of the limited space available in the URL. The link shortener will error if the URL length is more than 512 characters, which can happen with long descriptive from/to text and intermediate places added.

flibbertigibbet:
Addressed by #506.

PR #498, opened 2016-09-13
Feature/intermediate places

### LINKED ISSUE #495: Draggable Routes: Add new/changed points in route as waypoints in OTP query

flibbertigibbet:
Closes #495.

Finish implementing route editing by re-querying with waypoints when user saves edits to a route. Follow-up to #496. After editing a route and clicking 'save', the app re-queries. The new route that then displays should include the added waypoints.

Required updating to OTP 1.0, as waypoint (`intermediatePlaces`) queries are broken on 0.20 (see opentripplanner/OpenTripPlanner#1784.)

Also customized the tooltip text shown in edit mode.

As `intermediatePlaces` are passed by repeating the parameter, multiple waypoints could not be passed with our existing dictionary of parameters, so the waypoint parameter substring gets built separately.k

Note that if the `toText` and `fromText` extraneous parameters passed for URL routing are too long, the link shortener may error when `intermediatePlaces` are added if the resultant URL is more than 512 characters. (Created issue #497 for that.)

### DIFF
--- a/src/app/scripts/cac/control/cac-control-sidebar-directions.js
+++ b/src/app/scripts/cac/control/cac-control-sidebar-directions.js
@@ -58,6 +58,7 @@ CAC.Control.SidebarDirections = (function (_, $, Control, BikeModeOptions, Geoco
     var itineraryListControl = null;
     var typeaheadDest = null;
     var typeaheadOrigin = null;
+    var waypoints = null;
 
     function SidebarDirectionsControl(params) {
         options = $.extend({}, defaults, params);
@@ -99,6 +100,8 @@ CAC.Control.SidebarDirections = (function (_, $, Control, BikeModeOptions, Geoco
                                        onItineraryClicked);
         itineraryListControl.events.on(itineraryListControl.eventNames.itineraryHover,
                                        onItineraryHover);
+
+        mapControl.events.on(mapControl.eventNames.waypointsSet, queryWithWaypoints);
 
         typeaheadDest = new Typeahead(options.selectors.typeaheadDest);
         typeaheadDest.events.on(typeaheadDest.eventNames.selected, onTypeaheadSelected);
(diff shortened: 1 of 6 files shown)


flibbertigibbet:
The responses when `arriveBy` is true appear to be broken, whatever the order of the waypoints passed; the origin point is always omitted. I suspect that is an issue with OTP, but would have to look into it some more.

src/app/scripts/cac/control/cac-control-sidebar-directions.js
```diff
@@ -168,6 +171,24 @@ CAC.Control.SidebarDirections = (function (_, $, Control, BikeModeOptions, Geoco
             arriveBy: arriveBy
         };
 
+        // add intermediatePlaces if user edited route
+        if (waypoints) {
+            // intermediatePlaces are ordered. see comment on `getGraphPathsConsideringIntermediates`:
+            // https://github.com/opentripplanner/OpenTripPlanner/blob/master/src/main/java/org/opentripplanner/routing/impl/GraphPathFinder.java#L241
+            // if `arriveBy` is `true`, places are expected to be ordered right-to-left.
+            if (arriveBy) {
+                waypoints.reverse();
```
KlaasH: I read that comment to mean that they're going to do the calculation in reverse internally but not that the waypoint arguments should be fed in in reverse.  And it looks like what the code does is build the same array either way then loop through it forward for other cases or backward for arriveBy.

Also when I add wacky waypoints to a route they appear to be traversed in the order I expect when I comment out this `waypoints.reverse()`.

KlaasH:
Bugs/needed changes:
- The missing origin in "arrive by" directions is a problem.  I seem to get the bug for walking-only or biking-only directions, not multi-mode, even if the multi-mode directions don't actually include a transit leg.
- Re-editing:
  - If I enter edit mode and only remove waypoints, it ignores me and the route goes back to the previous edited result.
  - If I move or add waypoints, it uses those but erases any previously-defined ones.  I.e. it treats the previously-edited route as though it were a fully-generated route and uses only the waypoints I change/set during the current editing session.
- Clicking the edit button doesn't disable it, so you can get lots of route editors going at once. Interesting looking but not ideal.
- Trivial, but in Firefox my cursor goes to the text-bar over the edit button rather than the click-hand.


Things I think should be done to improve usability.  I missed the UI review meeting for this, so maybe these are redundant or contradictory with what was said there, but here are my thoughts.
- Enter edit mode by clicking on the line (should probably also create a waypoint at the click that you can drag)
- Improve cursor visibility/differentiation over waypoints
- Update route on release.  Since we don't have snapping or dynamic routing, it seems to me like making multiple waypoints before seeing what each one does to the route isn't that useful. Recalculating after every change would be slower, but not as slow as making the user click 'save' and then recalculating after every change, which seems to me how this will play out in practice.  The 'Cancel' button would not be relevant in this setup, but we would need a 'Reset' button (which might be useful anyway).  (This assumes we add re-editing capability.)
- Distinguish between user-created waypoints and the ones auto-generated from turns.  The line-midpoint ones are already slightly different, but if I want to change an already-edited route I need to know which points are definitely goign to get routed through and which are just showing up for convenience. (This also assumes re-editing.)

KlaasH:
Ok, I wasn't clear on the proof-of-concept status of this.  Sounds reasonable.  I would stay the multiple-concurrent-editors bug (which probably arrived in #496 but I didn't notice before) is worth ironing out now.  And if the cursor thing is an easy class or HTML change...

Other than that, +1.

flibbertigibbet:
The editor-click issue (nice find) should go away once we ditch displaying the editor buttons altogether as part of the upcoming changes, so I don't think it would be worth it.

Issue #502, opened 2016-09-19
Additive waypoint editing


flibbertigibbet:
Depends on #499.

Rather than passing all turn points in route to Leaflet Draw on entering edit mode, initially only display journey start and end point (connected by a straight line). When user releases mouse after dragging route line, store the added waypoint and immediately re-query for the edited route. Entering edit mode again from a route with one or more waypoints will pass the previously added waypoints to Leaflet Draw, in addition to the start and end points.

Do not display the Leaflet Draw edit toolbar (edit pencil button in upper left of map, with 'save' and 'cancel' buttons to its right).

flibbertigibbet:
Closing in favor of #514.

PR #507, opened 2016-09-27
Feature/store waypoints

### LINKED ISSUE #499: Store waypoints
When user edits route, internally maintain a list of waypoints (all turn points added or modified by the user.)

flibbertigibbet:
Closes #499.

Stores user-added waypoints (modified places on an edited route) in local storage.
Puts waypoints in the URL also. Cleared on origin or destination change,
or other events that cause all itineraries to be cleared from the map.

To test, edit a route. New `waypoints` should be set in local storage and in URL; edited route should be preserved on page refresh. Shortened link URL should contain `intermediatePlaces` parameter.

Does not yet preserve previously edited waypoints on subsequent route edits, as we will need to implement #502 to be able to easily get the ordering right.


sharph:
build again

KlaasH:
Looks good :+1:

PR #511, opened 2016-09-29
Additive waypoint editing

flibbertigibbet:
Edit route by starting with straight line from origin to destination, then add in waypoints to it, using previously added waypoints on successive edits.

Decided to break up work on #502 into two parts, since the remaining UI aspects I think will require forking Leaflet Draw. (Do not display edit toolbar and redraw on mouse release.)

### DIFF
--- a/src/app/scripts/cac/map/cac-map-control.js
+++ b/src/app/scripts/cac/map/cac-map-control.js
@@ -1,4 +1,4 @@
-CAC.Map.Control = (function ($, Handlebars, cartodb, L, _) {
+CAC.Map.Control = (function ($, Handlebars, cartodb, L, _, UserPreferences) {
     'use strict';
 
     var defaults = {
@@ -449,11 +449,15 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, _) {
     function editItinerary(itinerary) {
         editingItinerary = itinerary;
 
-        // edit a simplified shape that just has the turn points
+        // edit a simplified shape that just has the start, end, and any previously added waypoints
+        var waypoints = UserPreferences.getPreference('waypoints');
+        var coordinates = _.concat([[itinerary.from.lon, itinerary.from.lat]],
+                                   waypoints,
+                                   [[itinerary.to.lon, itinerary.to.lat]]);
         map.removeLayer(itinerary.geojson);
         editLayer = cartodb.L.geoJson({
             type: 'LineString',
-            coordinates: itinerary.getTurnPoints()
+            coordinates: coordinates
         });
 
         editLayer.setStyle(itinerary.getStyle(true, true));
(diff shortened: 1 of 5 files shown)


src/app/scripts/cac/urlrouting/cac-urlrouting.js
```diff
@@ -14,8 +14,8 @@ CAC.UrlRouting.UrlRouter = (function (_, $, UserPreferences, Utils, Navigo) {
     var SHARED_READ = ['maxWalk', 'wheelchair', 'bikeTriangle'];
     var EXPLORE_ENCODE = SHARED_ENCODE.concat(['placeId', 'exploreTime']);
     var EXPLORE_READ = EXPLORE_ENCODE.concat(SHARED_READ);
-    var DIRECTIONS_ENCODE = SHARED_ENCODE.concat(['destination', 'destinationText', 'waypoints']);
-    var DIRECTIONS_READ = DIRECTIONS_ENCODE.concat(SHARED_READ).concat(['arriveBy']);
+    var DIRECTIONS_ENCODE = SHARED_ENCODE.concat(['destination', 'destinationText']);
+    var DIRECTIONS_READ = DIRECTIONS_ENCODE.concat(SHARED_READ).concat(['arriveBy', 'waypoints']);
```
flibbertigibbet: Modified this because of the strange way OTP send waypoints by repeating the `intermediatePlaces` parameter. Updates to user preferences are already handled in code. Unsure what this would mean for someone viewing a shared link that has waypoints from a browser that already has waypoints in local cache.

src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -492,31 +496,15 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, _) {
             // get the points from the edited linestring
             var modified = editLayer.toGeoJSON().features[0].geometry.coordinates;
             cleanUpItineraryEditEnd(false);
-            endItineraryEdit(itinerary.getTurnPoints(), modified);
+            // remove start and end points
+            modified.shift();
+            modified.pop();
```
KlaasH: I seem to have an aversion to `shift` and `pop` in null context.  Not sure if it's rational or not.  Lodash `slice` supports negative indexes, though, so this could be simplified to
```
// get the points from the edited linestring, excluding start and end
var modified = _.slice(editLayer.toGeoJSON().features[0].geometry.coordinates, 1, -1);`
```

src/app/scripts/cac/urlrouting/cac-urlrouting.js
KlaasH: So this means the address bar URL omits waypoints, right?  So to share an edited route you'd need to use the share button?  Right now the shared itinerary ignores waypoints, I think because this is looking for `waypoints` but the parameter being included in the URL is `intermediatePlaces`.

But yeah, also this means that if you make an edited itinerary such that you have waypoints in your local storage then you follow a link copy/pasted from the address bar, you end up with the start/end/etc parameters encoded in the link but routing through the waypoints you defined for your old route.  If we can't include waypoints in the address bar URL, we need to at least figure out how to clear them when appropriate.

flibbertigibbet: Oh, I expected this to include the parameter in the URL, but not then save it (incorrectly) into local storage from the URL. I think that should be an appropriate option for settings in a shared trip link that should override but not reset the user's local storage.

KlaasH: I agree that that sounds like a sensible and desirable thing to do, but I don't think we're set up for it.  I would have to dig in a little more to be more specific, but my memory of how this is all set up is that local storage is authoritative because there are lots of components that interact with it and aren't set up to take their parameters from arguments, so the way the app loads directions from a URL is to save all the URL params to local storage then kick off the directions generation function which reads from local storage to do its thing.

So I'm for the existence of a category of operational parameters distinct from the parameters in storage, but my understanding is that we're a non-trivial refactor away from having that.

flibbertigibbet:
Closing in favor of #514.

PR #512, opened 2016-10-03
Remove contingency alert

flibbertigibbet:
Remove contingency plan alert for SEPTA rail, now that regular service has resumed.

http://www.septa.org/service/contingency.html


flibbertigibbet:
Closing in favor of #513.

Issue #501, opened 2016-09-19
Special highlighting for waypoints on map


flibbertigibbet:
Depends on #499.

Differentiate map dot highlighting turn points that are also waypoints on direction list hover.

(Perhaps make a different color that non-waypoint turn points.)

PR #514, opened 2016-10-04
Feature/live update editing

### LINKED ISSUE #502: Additive waypoint editing

### LINKED ISSUE #503: Enter edit mode by clicking route
Probably best to do this after #502.

Instead of entering route edit mode by clicking pencil button at top of route directions, enter edit mode by clicking route. (Enable interaction for route polyline and remove pencil button from directions list.)

When entering edit mode, add clicked point on route to list of points for Leaflet Draw, unless user clicked on the circle for an existing waypoint.

flibbertigibbet:
Supersedes #511 and is an alternate approach to close #502. Also closes #503.

Instead of editing the route linestring in Leaflet Draw, show a draggable marker on route hover. When dragged marker is released, add the point as a new waypoint between the two nearest existing waypoints, or start or end point.

Removes Leaflet Draw and the route edit button and adds the turf.js `nearest` function. As the full turf.js is quite large (521K, minified), use browserify to build just the `nearest` module and the supporting helper functions for creation of point and feature collection objects.

The draggable marker displayed will need modification; circle markers in Leaflet are not draggable, so the easiest solution may be a custom marker. Also this does not address putting the waypoints in the URL, as the URL scheme needs some other related rework regarding storage and override.


flibbertigibbet:
Travis build failing apparently due to [this issue](https://github.com/MrBoolean/gulp-run/issues/31) with the gulp task that runs `collectstatic`. Considering removing it from gulp, as we've also had an issue with its output being too large, and it's easy enough to run separately.

src/gulpfile.js
```diff
@@ -107,7 +143,13 @@ gulp.task('copy:scripts', function() {
 });
 
 gulp.task('copy:vendor-css', function() {
-    return copyBowerFiles('**/*.css', [bootstrapSelectRoot + 'css/bootstrap-select.css'])
+    return copyBowerFiles('**/*.css',
+                          // FIXME: excluding minified CSS results in console error
+                          // about missing font awesome webfonts
+                          //'!**/*.min.css',
```
KlaasH: Should we make an issue for this?

KlaasH:
`vagrant provision app` didn't load the new npm modules properly for me.  Or at least, it left me with `npm run gulp-development` complaining that aliasify couldn't be found.  I got it working by blowing away my `node_modules` and running `npm install` again.

src/gulpfile.js
flibbertigibbet: Since we've apparently always been compiling CSS this way and it hasn't been an issue, I'm not too worried about it, but yes, let's make it an issue. I imagine with the redesign our font libraries and external CSS will change, so we might as well put off dealing with it until later.

flibbertigibbet:
1. Yes, #501 is to add markers highlighting previously added waypoints.

2. In our design discussions we talked about wanting to make it feel like draggable rerouting and hide the fact that we're using waypoints to accomplish that. Since to the user it's supposed to look like they're just dragging to change the route and not like they are adding waypoints, calculating where in the sequence to place the new waypoint based on the end drop seems like more natural-feeling behavior to me than requiring it stay between the two nearest existing waypoints to the initial line click, which was what Leaflet Draw was essentially doing for us.

flibbertigibbet:
I destroyed the app VM on Jenkins so it could do a fresh build based on this branch.

KlaasH:
Sounds reasonable.  :+1:

flibbertigibbet:
Going to hold off merging this PR until I figure out why the gulp production build isn't quite working.

flibbertigibbet:
Update: demoed to John. He does want the behavior of keeping new points between the two nearest existing points on drag start.

jbranigan:
@flibbertigibbet Let's run it by Nick and Scott to get their read before making a final call on the waypoint ordering

flibbertigibbet:
rebuild

flibbertigibbet:
build again

src/gulpfile.js
flibbertigibbet: #515

flibbertigibbet:
Modified to add new waypoint between nearest points to drag start, rather than drag end, per discussion in status meeting.

PR #525, opened 2016-10-10
Feature/waypoints urls

flibbertigibbet:
Retain waypoints in URL and handle parsing them. Closes #516.

To test, create a trip with waypoints, click to share link, and open the shortened shareable URL in another browser or incognito tab (that does not have site settings in local storage). Trip directions should load with waypoints showing, and from/to text correctly displayed.

Issue #524, opened 2016-10-06
Waypoint directions popovers


flibbertigibbet:
Display instructions to user to drag a point to edit route, or to click to delete a waypoint.

Issue #522, opened 2016-10-06
Custom waypoint marker


flibbertigibbet:
Either use a custom marker icon that is circular with a modified anchor point, or use a circle marker modified to be draggable. Custom marker option is preferable, as it would allow us to reuse existing functionality.

Issue #521, opened 2016-10-06
Waypoint editing


flibbertigibbet:
Make markers for existing waypoints draggable, and update the waypoint coordinates on drag end.

Depends on #501.

flibbertigibbet:
Closed in #526.

PR #526, opened 2016-10-11
Feature/waypoint editing

### LINKED ISSUE #501: Special highlighting for waypoints on map

flibbertigibbet:
Display draggable markers for waypoints. Delete waypoint on marker click; move waypoint when marker dragged.

Closes #501, #521, and #520. Re-uses destination marker icon for now, until #522 can be completed.

Adds tooltips to the markers that may be sufficient for #524, unless we don't want to make the user wait for the tooltip hover delay to find out what to do.

Also fixes issue where all itineraries would be editable if multiple shown; limits itinerary editing to when only one itinerary is shown, either because only option one was returned from the trip planner, or because the user has selected the itinerary from the list. Note that if a plan query includes waypoints, only one itinerary will be returned.

To test, try adding, moving, and deleting waypoints, and clicking in and out of the directions list to verify expected behavior.

Developed this on a branch based on #525, flibbertigibbet:feature/waypoint-markers.

### DIFF
--- a/src/app/scripts/cac/map/cac-map-control.js
+++ b/src/app/scripts/cac/map/cac-map-control.js
@@ -40,7 +40,6 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
     var itineraryHoverListener = null;
     var isochroneLayer = null;
     var waypointsLayer = null;
-    var waypointsMarkers = {};
     var tabControl = null;
 
     var destinationIcon = L.AwesomeMarkers.icon({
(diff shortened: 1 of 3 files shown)


flibbertigibbet:
build again

KlaasH:
Seems like the waypoint marker layer is sometimes not getting cleared.  If I click the itinerary after setting a waypoint then click over to the Explore tab, the green markers stick around, apparently forever.

stale markers

KlaasH:
Not necessarily specific to this, but related to the previous comment: is it intended behavior that switching to Discover Places and back re-runs your directions and gets rid of your waypoints?

src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -425,7 +439,7 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _, UserPreferences
                     }).on('dragend', function(e) {
                         dragging = false;
                         var coords = e.target.getLatLng();
-                        addWaypoint(itinerary, [coords.lng, coords.lat],
+                        addWaypoint(itinerary, [coords.lat, coords.lng],
```
KlaasH: Maybe add a note in the comment on `addWaypoint` making it clear that the waypoint argument is [lat, lon] and the startDragPoint argument is [lon, lat].

flibbertigibbet:
Good catch. I set it up to clear things under other cases, but not for tab switch.

KlaasH:
Though the draggable markers no longer show up when there are multiple itineraries, the cursor still changes when it passes over the line.  It would be extra cool if in multi-itinerary mode clicking the line selected the itinerary.  Next best would be remaining unchanged to make it clear that the line is not currently interactive.

KlaasH:
:+1: aside from the minor notes above.

I took the branch name in the initial comment (flibbertigibbet:feature/waypoint-markers) to be a typo.  To test this I rebased onto flibbertigibbet:feature/waypoint-urls, which is the branch for PR #525.

flibbertigibbet:
Branch name not a typo; it's [here](https://github.com/flibbertigibbet/cac-tripplanner/tree/feature/waypoint-markers). Just rebased this onto develop, which now has #525.

flibbertigibbet:
build again

flibbertigibbet:
I haven't been able to reproduce waypoint markers lingering after switching to the explore tab; the layer gets removed. The marker icon being used currently for waypoints happens to be the same as that for the events shown on the explore tab; were those perhaps event markers (have a popup with event info if so)?

flibbertigibbet:
Sorted issue with waypoint marker layer not always going away; it would sometimes get added multiple times if user repeatedly goes into and out of the directions detail list.

flibbertigibbet:
Regarding cursor hover on the line: Leaflet does not currently allow toggling interactivity on and off for a layer (issue [here](https://github.com/Leaflet/Leaflet/issues/2669)).

PR #528, opened 2016-10-12
Don't show waypoints in directions list

### LINKED ISSUE #500: Do not show waypoints as intermediate destinations in directions list
Conflate multiple trips to look like one

KlaasH:
Waypoints always cause OTP to make a new leg, but we don't want them to
show up as separate "To Destination" steps in the directions list.
So this processes the legs array to conflate the steps on either side
of a waypoint.

**Notes**
- This post-hoc combination of legs isn't ideal, and could be susceptible to subtle bugs from changes in the attributes of the leg objects returned by OTP, but I'm not seeing a cleaner way to do it.
- Putting a waypoint in the middle of a transit leg always makes you stop there.  I.e. if you want to force your directions to take the 23 bus from center city to Germantown, dragging a waypoint onto a spot along the 23's route might put you on the 23 the whole way, but the resulting directions will have you get off at the spot and then get back on.  I don't see any way to get around that.  I note also that Goog the Great and Powerful doesn't allow route customization in transit mode.

**To test**
Drag a waypoint that's not in the middle of a transit leg (either in single-mode directions or a walking/biking leg of multi-mode directions) and click on the itinerary to show the directions list. The step you modified should still be all one step, just with more turns, and there should be no "to Destination" steps except the last one.

merged_across_waypoints

Closes #500.

### DIFF
--- a/src/app/scripts/cac/routing/cac-routing-itinerary.js
+++ b/src/app/scripts/cac/routing/cac-routing-itinerary.js
@@ -203,8 +203,8 @@ CAC.Routing.Itinerary = (function ($, cartodb, L, _, moment, Geocoder, Utils) {
         return newLegs;
     }
 
-    /* Waypoints always result in a step break, which ends up producing useless "To Destination"
-     * steps in the itinerary details.
+    /* Waypoints always result in a step break, which ends up producing intermediate
+     * "to Destination" steps that we don't want to show in the itinerary details.
      * See https://github.com/opentripplanner/OpenTripPlanner/blob/otp-1.0.0/src/main/java/org/opentripplanner/routing/impl/GraphPathFinder.java#L305
      * and https://github.com/opentripplanner/OpenTripPlanner/blob/otp-1.0.0/src/main/java/org/opentripplanner/api/resource/GraphPathToTripPlanConverter.java#L212
      * There's no configuration option to make OTP not do that, so instead this munges the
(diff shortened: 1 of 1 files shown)


src/app/scripts/cac/routing/cac-routing-itinerary.js
```diff
@@ -195,6 +197,48 @@ CAC.Routing.Itinerary = (function ($, cartodb, L, _, moment, Geocoder, Utils) {
...
+     * Note that this makes no attempt to merge the `legGeometry` attributes so it's important that
+     * `getFeatures`, which gets the itinerary's geometry into Leaflet, gets called on the original
+     * legs array rather than the munged one.
+     */
+    function mergeLegsAcrossWaypoints(legs) {
+        if (legs.length === 1) {
+            return legs;
+        }
+        var index = 0;
+        while(index < legs.length - 1) {
```
flibbertigibbet: Running a `while` loop over a mutating collection makes me a little nervous, but I can't think of any cases where this might get stuck in an infinite loop or throw an out-of-range exception.

src/app/scripts/cac/routing/cac-routing-itinerary.js
```diff
@@ -195,6 +197,48 @@ CAC.Routing.Itinerary = (function ($, cartodb, L, _, moment, Geocoder, Utils) {
...
             return leg;
         });
+        if (hasWaypoints) {
+            newLegs = mergeLegsAcrossWaypoints(newLegs);
+        }
+        return newLegs;
+    }
+
+    /* Waypoints always result in a step break, which ends up producing useless "To Destination"
+     * steps in the itinerary details.
```
flibbertigibbet: Regarding the comments on OTP's behavior: their treatment of waypoints as intermediate stops is intentional; the feature is there with the intent of the user adding extra stops to a trip (i.e. stop for gas), so the directions returned with extra explicit destinations make sense for that use case. What we're doing is routing via waypoints in order to create the appearance of live-editing a route, which is not an intended use for intermediate stops, though it serves our purpose well enough.

KlaasH:
Good point about merging steps as well.  Added (4afc007).

src/app/scripts/cac/routing/cac-routing-itinerary.js
flibbertigibbet: Agreed. It's fine the way it is.

PR #530, opened 2016-10-18
GoPhillyGo v2 Prototype

lederer:
This is the prototype of GoPhillyGo v2.

### Installation and use
Run `npm install` in the project directory to install required packages.

Run `grunt` to watch for file changes and automatically open your browser to localhost:3000.

### Responsive
The entire app should be responsive to different screen sizes via css media queries.

### Information architecture
Much of the app lives in `app/index.html`. That's where the home view and map view live. Switch between them by changing `body.body-home` to `body.body-map`.

The prototype contains most of the components statically within `app/index.html` and toggles their visibility depending on the state of the app. Replace this approach with whatever rendering mojo you see fit (Angular, I assume).

### Map view
CSS flexbox is used for arranging the map and sidebar into 2-columns (desktop) or 1-column (mobile). 

When in map view, there should always be exactly 5 or 6 top-level sibling elements, in order: 

1. `.the-map`
1. `.app-header`
1. `.directions-form-element`
1. `.sidebar-banner` iff (1) options are set, or (2) biking directions are shown and indego banner hasn't been displayed for this user (or client) before
1. Some element containing the main content of the sidebar. Its SCSS must`@import` the `sidebar-main` mixin.
1. `.app-footer`

The `sidebar-main` mixin ensures the cumulative height of all these elements _except the map_ is 100vh, forcing the map to wrap into the second column. Note, for this to work correctly, whenever `.sidebar-banner` is shown, ensure `<body>` has the `.body-has-sidebar-banner` class. 

Feel free to change how all this works if you have a better way, though let me know so I can sanity check and/or learn.

### Info pages
All other pages (`app/learn.html`, `app/learn-detail.html`, `app/explore-detail.html`, `app/about.html`) depend on `body.body-info`. 

### Prototype-only files
`app/extras.html` contains various components missing from the prototype but that will need to be in the app. If anything is missing and isn't trivial to make, let me know and I'll make it for you.

`app/sass/pages/_PROTOTYPE.scss` contains minimal styles to make `app/extras.html` work.

`app/js/prototype.js` makes the prototype work, mostly via toggling classnames on `<body>` or other elements.

You can remove all these files from the production app, although you'll probably need to retain or adapt the last function in `prototype.js`. It's a hack that makes the _latest blog post_ card on the home page work (it has a `<button>` nested inside an `<a>`).


flibbertigibbet:
build again

PR #531, opened 2016-10-18
Use circles for waypoint markers

### LINKED ISSUE #522: Custom waypoint marker

KlaasH:
Leaflet has `Marker`, on which you can very conveniently set `draggable: true`, and it has `CircleMarker`, which very conveniently provides circular markers, but ne'er the twain shall meet.

This converts the waypoint markers to CircleMarker then reimplements dragging using mousedown/mousemove/mouseup events.

I see now that the issue says "Custom marker option is preferable".  Oops.  I don't disagree, and am open to throwing this away and making a custom icon instead.  That path was proving a bit less straightforward than I expected so I switched to this way to see how it would go and ended up getting caught up in trying to make it work even though it's likely to be more brittle and less usable/maintainable than we'd like.

The styles for new and existing markers are the same, but I didn't use a shared options object because I figured there's a good chance we'll want them to not look identical.

Also the green I chose is nice and Christmasy and won't be confused with the turn highlighting circles, but possibly a different color scheme is preferable.

Resolves #522.

merry_christmas

### DIFF
--- a/src/app/scripts/cac/map/cac-map-control.js
+++ b/src/app/scripts/cac/map/cac-map-control.js
@@ -433,48 +433,71 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
                 var startDragPoint = null;
                 // Use a timeout when closing the popup so it doesn't strobe on extraneous mouseouts
                 var popupTimeout;
-                lastItineraryHoverMarker = new cartodb.L.Marker(e.latlng, {
-                        draggable: true,
-                        icon: highlightIcon,
-                    }).on('dragstart', function(e) {
-                        dragging = true;
-                        startDragPoint = e.target.getLatLng();
-                    }).on('dragend', function(e) {
-                        dragging = false;
-                        var coords = e.target.getLatLng();
-                        addWaypoint(itinerary, [coords.lat, coords.lng],
-                                    [startDragPoint.lng, startDragPoint.lat]);
-                        startDragPoint = null;
-                    });
+                // The marker closes with a timeout as well, and we want to hold on to it so we can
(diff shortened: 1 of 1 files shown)


src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -433,49 +433,72 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
...
-                                map.removeLayer(lastItineraryHoverMarker);
-                                lastItineraryHoverMarker = null;
-                                dragging = false;
-                                startDragPoint = null;
-                            }
-                        }, 3000);
                     });
 
+                // circleMarkers aren't draggable, so this implments dragging using mouse
+                // down/move/up events. What could go wrong?
```
flibbertigibbet: :laughing: 

The reason why I'd suggested a preference toward making a circle icon for the already-draggable marker is that we wouldn't have to deal with anything broken with a custom draggable implementation, or any weird browser edge cases.

flibbertigibbet:
Glad you were able to implement dragging without too much trouble.

This will conflict significantly with #532, so let's wait until that's in to revisit marker changes, as rebasing a smaller thing onto a bigger thing should be much easier.

KlaasH:
After much disappointment, Google finally comes through: https://groups.google.com/forum/#!topic/leaflet-js/GSisdUm5rEc
So much cleaner and easier.

I'll put up a new PR with that approach once #532 is in.

PR #532, opened 2016-10-18
Feature/live updating reroute

### LINKED ISSUE #523: Live update route during edit
As user is dragging waypoint, live update route path (debounced).

Do not update step-by-step directions until route edit ends.

### LINKED ISSUE #527: Add console polyfill
Consider adding a polyfill such as [console-polyfill](https://github.com/paulmillr/console-polyfill) to the project.

We have been vigilant about keeping console logs from getting in, but we do have a number of `console.error` statements that might break things more than they were already if hit in IE.

flibbertigibbet:
Show re-routing previews as the user drags a waypoint marker. Closes #523.

Also add a `console` polyfill so that if any of our error logs happen on a browser without `console` defined, it won't break things more. (Close #527.)

Would like a refactor on the map controller after this goes in, as it has slowly grown to nearly 1,000 lines and could be broken up into several files.

### DIFF
--- a/src/app/scripts/cac/map/cac-map-control.js
+++ b/src/app/scripts/cac/map/cac-map-control.js
@@ -135,6 +135,7 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
     MapControl.prototype.clearDirectionsMarker = clearDirectionsMarker;
     MapControl.prototype.highlightDestination = highlightDestination;
     MapControl.prototype.displayPoint = displayPoint;
+    MapControl.prototype.updateItineraryLayer = updateItineraryLayer;
 
     return MapControl;
 
@@ -663,6 +664,11 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
 
         // requery with the changed points as waypoints
         events.trigger(eventNames.waypointsSet, {waypoints: coordinates});
+    }
+
+    function updateItineraryLayer(oldLayer, newLayer) {
+        map.removeLayer(oldLayer);
+        newLayer.addTo(map);
     }
 
     function clearItineraries() {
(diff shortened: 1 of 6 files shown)


src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -644,6 +748,21 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
...
+     * which mean itinerary layers cannot be reliably removed by using the dictionary of
+     * itineraries stored in the map control during or on end of waypoint drag.
+     */
+    function removeItineraryLayers() {
+        map.eachLayer(function(layer) {
+            if (layer.feature && layer.feature.properties && layer.feature.properties.from) {
+                map.removeLayer(layer);
+            }
+        });
+    }
```
KlaasH: What about maintaining an `itineraryLayerList` array that you push layers itinerary layers onto whenever you add them to the map?  Then this could loop over that and remove each layer then set the array back to `[]`.  Leaflet doesn't appear to mind if you call removeLayer on a layer that was already removed.

Pro: none of this "if it has a `from` attribute, ice it" business.
Con: have to be sure to push itinerary layers to the array when adding them to the map (currently done in 2 places). Possibly a wrapper function would be prudent.  Also probably want to remove them from the array in `updateItineraryLayer`.

src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -471,10 +501,15 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
...
-                                startDragPoint = null;
+                                newWaypointIndex = null;
                             }
                         }, 3000);
-                    });
+                    }).on('drag', function(event) {
+                        // get itinerary from collection to pick up newer version with
+                        // layer modified by live dragging updates
+                        redrawWaypointDrag(event, newWaypointIndex, true);
+                    }).bindPopup('Drag marker to change route',
```
flibbertigibbet: No reason, apart from it having different syntax than the event bindings above. I'll move it above the event bindings.

src/app/scripts/cac/map/cac-map-control.js
KlaasH: Ah, I was thinking the problem was that we sometimes clobbered a layer reference before removing it.  But yeah, if the map has lost track and can't remove it by reference then keeping a copy isn't much help.

PR #533, opened 2016-10-19
Circle waypoint markers

### LINKED ISSUE #522: Custom waypoint marker

KlaasH:
Converts the marker icon for waypoints to a circle.  Replaces PR #531 and resolves #522.

Also fixes a small issue I came across while testing by adding a `clearTimeout` to stop the add-waypoint marker from being deleted if you moused out but then came back to it.

circles_again

### DIFF
--- a/src/app/scripts/cac/map/cac-map-control.js
+++ b/src/app/scripts/cac/map/cac-map-control.js
@@ -55,6 +55,18 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
         iconColor: 'black',
         markerColor: 'lightgreen'
     });
+    var waypointRadius = 6;
+    var waypointColor = '#444';
+    var waypointCircle = '<svg xmlns="http://www.w3.org/2000/svg" version="1.1" ' +
+        'width="' + waypointRadius * 2 + '" height="' + waypointRadius * 2 + '">' +
+        '<circle cx="' + waypointRadius + '" cy="' + waypointRadius +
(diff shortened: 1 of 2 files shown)


KlaasH:
build again

flibbertigibbet:
Not sure if I'm missing something, but the waypoint markers here aren't dragging for me (just pans map).

rmartz:
Can't rule out that I'm doing it wrong, but my experience mirror's @flibbertigibbet's... I see the dots and they follow the mouse, but trying to drag one pans the map.

flibbertigibbet:
:+1: nice find!

To correct the comment on the last commit, cartodb.js does not depend on leaflet; it includes its own, modified version of leaflet. The Leaflet.encoded polyline decoder library we're using pulls in leaflet, which needs to be there to build, but which we then exclude from our minified/exported scripts, in favor of the cartodb.js (included) leaflet.

flibbertigibbet:
Despite having completely destroyed and rebuilt my app VM, I found it necessary to run `npm run bower-clean` to clear the bower cache, which apparently must be kept on the host. This would explain why we were experiencing different behavior.

KlaasH:
I was unclear on the exact relationship of cartodb.js to Leaflet, because I thought when I cleared everything but cartodb.js from the bower.json it still ended up installing Leaflet.  But I just tried it again and it seems I was mistaken.  Thanks for the clarification.

PR #550, opened 2016-10-21
Feature/redesign django template

### LINKED ISSUE #535: Create Django template for redesign
Make a Django HTML template to load the prototype home page. Doesn't need to load the map in background or actual place data in the cards list.

flibbertigibbet:
Load prototype home page. Closes #535.

Modifies base and home page templates to look like the prototype. No functionality yet.

Removes a number of dependencies and build steps we will no longer be using on the frontend, including bootstrap, FontAwesome, and Material Design fonts.

Adds build step for handling new Fontello fonts and icons.

To test, load home page. It should look like the one under `src/prototype` (which may be run separately via grunt outside the VM).

This breaks most things :smile: 
Here's a partial list of next steps; some of these may be better suited to go in separate PRs/tasks.

## Remaining tasks:
- [x] fix JavaScript tests

- [x] fix Python tests

- [x] update jQuery version (was held back due to date picker widget, now removed)

- [x]  move map initialization to JS (rework home page JS) (created issue #553)

- [x] incorporate `prototype.js` click handlers (depends on #553, can happen with other tasks)

- [x] Wire up email subscription form to JotForms (and handle jQuery conflict) (created issue #551)

- [x] Add back Facebook and Google Tag Manager script loading

- [x] find better approach to body/top level div class toggle, with design help (currently doesn't play nice with templating)

- [x] remove `old_images` from images directory (keeping for now, since we may want some marker icons) (created #552 as a reminder)


python/cac_tripplanner/templates/home.html
```diff
@@ -304,12 +304,30 @@
...
+    attribution: '&copy; <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a> &copy; <a href="http://cartodb.com/attributions">CartoDB</a>',
+    subdomains: 'abcd',
+    maxZoom: 19
+    }).addTo(map);
+
+    // set class on body
+    // FIXME: avoid having to do this for templating to work.
+    // also had to wrap contents here in a div with the same class
+    // leave out class on div or class on body, and either have no map,
+    // or map and no content overlay
```
KlaasH: I'm not clear on what this comment is saying.  Where's the extra div?

flibbertigibbet: The top-level div in home.html: `<div class="body-home">`.

In the prototype, that div does not exist and that class exists only on the `body`. I had to add the extra div and put the class on both.

flibbertigibbet: Update: sorted! With design help.

python/cac_tripplanner/templates/home.html
```diff
@@ -320,14 +320,6 @@
...
     maxZoom: 19
     }).addTo(map);
-
-    // set class on body
-    // FIXME: avoid having to do this for templating to work.
-    // also had to wrap contents here in a div with the same class
-    // leave out class on div or class on body, and either have no map,
-    // or map and no content overlay
-    $('body').removeClass();
-    $('body').addClass('body-home');
```
KlaasH: This gets rid of this class manipulation but still leaves the new div.  Is the plan to use the `body-home` and `body-map` classes only on the div, not on `<body>`?

KlaasH:
[This comment] (https://github.com/azavea/cac-tripplanner/pull/550/commits/8136d1b7ad20ec582021e4bc7ebc00a36e5d0fb5#r84741066) showed up as outdated as soon as I wrote it, I guess because I made a line comment on a line that was added then removed in this PR.  

Anyway, this looks great.  :+1:

flibbertigibbet:
Oh yeah, I didn't see that message come through. But yeah, plan is now to set the `body-home` or `body-map` class on the added `div` and not manipulate the `body` classes. Which means that the class names are something of a misnomer now, but structurally I like it this way.

Issue #536, opened 2016-10-20
Redesign for Handlebars places template


flibbertigibbet:
Apply redesign prototype styles to handlebars template for places cards.

flibbertigibbet:
Awaiting new style for place in the sidebar.

lederer:
Place cards in the sidebar are the same as in the Home view. If an origin is set, they have an green "time · distance · mode" component.

https://marvelapp.com/30a0g54/screen/16895442

PR #558, opened 2016-10-26
Feature/load places homepage

flibbertigibbet:
Load four destinations via template on homepage cards. Issue intended for the card was #536, though I will leave that issue open, since there is a handlebars template for places, but it's used in the sidebar in the 'explore' tab and not the homepage, which uses Django templating. To test, home page should load four different locations instead of the placeholder Heinz refuge for all cards.

Also fixes font usage to bring back awesome-markers, which fixes the map controller, which fixes the frontend tests.

Brings back the home page controller JS. Note the tab control will need significant rework. Map controller uses a new `homepage` flag to load but hide hide map controls (layer switcher, zoom control, and layers). We might want to revisit how this is managed.


flibbertigibbet:
The app prototype loads some custom images that have been resized from what we have in the app, so the sizing looks off from the design.

@lederer our user-uploaded images are at a predefined aspect ratio. We have two images for each destination or article, one expected to display at 400x400, the other, wider image at 400x200. Can we work with that, or will we need to modify the existing images in our CMS?

flibbertigibbet:
Ah okay, by using the wider destination image, it comes out to display the same as the prototype, 310x155.

src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -174,16 +186,28 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
         overlays['Bike Share Locations'] = overlaysControl.bikeShareOverlay();
         overlays['Bike Routes'] = overlaysControl.bikeRoutesOverlay(map);
 
+        // TODO: handle hiding layers on home view more cleanly.
+        // Would be better to initialize but not add to map, to avoid flashing,
+        // although it probably won't be visible due to centering of home page text.
+        if (homepage) {
+            overlays['Bike Share Locations'].eachLayer(function(layer) { layer.hide(); });
+            overlays['Bike Routes'].eachLayer(function(layer) { layer.hide(); });
+        }
```
flibbertigibbet: Good point. If someone had them selected and went back to the home view, I suppose we would need to deal with that.

Issue #504, opened 2016-09-19
Waypoints with 'arrive by' broken


flibbertigibbet:
This is an issue with OpenTripPlanner, but opening against this repo for tracking.

Currently, routing with waypoints when `arriveBy=true` returns incorrect results (origin is missing); there is an open issue to fix routing with waypoints here: opentripplanner/OpenTripPlanner#2345.

Should investigate and report to upstream with examples, and fix if we can.

PR #560, opened 2016-10-27
Disable draggable itinerary when arriveBy is true

flibbertigibbet:
Do not allow user to reroute by dragging itinerary for trips planned with arrive-by at time set.
Workaround for #504 (directions with waypoints broken for non-transit trips with arrive-by set.)

### DIFF
--- a/src/app/scripts/cac/control/cac-control-sidebar-directions.js
+++ b/src/app/scripts/cac/control/cac-control-sidebar-directions.js
@@ -157,10 +157,7 @@ CAC.Control.SidebarDirections = (function (_, $, Control, BikeModeOptions, Geoco
         var date = picker.date() || moment();
 
         var mode = bikeModeOptions.getMode(options.selectors.modeSelectors);
-        var arriveBy = false; // depart at time by default
-        if ($(options.selectors.departAtSelect).val() === 'arriveBy') {
-            arriveBy = true;
-        }
+        var arriveBy = isArriveBy(); // depart at time by default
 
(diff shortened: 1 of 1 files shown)


KlaasH:
Bummer to be having to disable features due to upstream bugs, but this does the job.  :+1: 

In probably-unrelated news, changing this all-walking itinerary from "Depart at" to "Arrive by" changes the route, regardless of what time I use.  I know arrive-by directions are generated backwards, but that still seems a little odd.  Unless it's a case of a perfect tie between segment weights and they end up getting returned in opposite order so the time breaking choice switches...

flibbertigibbet:
I see that too. There may be valid reasons to route walking differently on changed direction, say, due to OSM sidewalk and crosswalk tags.

KlaasH:
Oops, I said "this itinerary" but didn't include a link.  [Here's one] (https://gophillygo.org/map/directions?origin=39.9616133%2C-75.1541914&originText=990%20Spring%20Garden%20St%2C%20Philadelphia%2C%20Pennsylvania%2C%2019123&mode=WALK&destination=39.952748%2C-75.1651571&destinationText=Dilworth%20Park).

I'm not flipping the direction, though, just the timing.  So what it's saying is that I should take a different route if I leave at 3:00 to walk to City Hall than if I leave at 2:37 to arrive at City Hall by 3:00.  Or, if I change the times to describe an exactly identical trip, it still gives different routes.

Not important, but a little odd.

PR #561, opened 2016-10-27
Copy media assets on dev VM

### LINKED ISSUE #534: Default media not copying on dev VM
The images in `default_media` testing images should be copied by Django into `/media/cac`. This does not affect production.

flibbertigibbet:
Fixes #534.

When building VMs during development, test media assets will be copied to the expected directory. Fixes destination images not displaying after fresh provision.


KlaasH:
:+1:

PR #562, opened 2016-10-27
Merge fixes from develop into redesign

KlaasH:
With the stuff merged today, draggable routing is good to go and develop is going to be pretty quiet for a while, right?  Seems like a good time to pull those fixes into the redesign branch.


flibbertigibbet:
There's a fix for packer not downloading on fresh provision I'm looking at right now.

If the redeploy works with the latest version, I'll merge in the version change. Might as well hold off on updating until then.

flibbertigibbet:
Looks like the packer download issue was transient, and our outdated version is still available. Also, our scripts don't work with newer versions.

flibbertigibbet:
Since we're going to merge this branch back in to develop eventually, shouldn't we rebase instead of merge here?

flibbertigibbet:
Actually merging will be great, since we can keep history that way. Was thinking it might cause issues when we go to merge the other way later.

flibbertigibbet:
Will need to bump java version to fix fresh OTP provision; will have PR out on development soon for that.

KlaasH:
With the fixes in and the launch successful, this is good to go.  Pushing the button.

PR #563, opened 2016-10-28
Bring selected itinerary layer to front

flibbertigibbet:
To fix interactivity issues on Firefox.


flibbertigibbet:
Trivial, merging.

Issue #537, opened 2016-10-20
Redesign template for articles


flibbertigibbet:
Redesign Handlebars template for articles cards

flibbertigibbet:
Closed in #566

PR #566, opened 2016-10-28
Feature/learn tab

### LINKED ISSUE #543: Implement article list in Learn tab
Display cards for all articles when user clicks 'Learn' from directions page.

KlaasH:
Creates the django views and templates for the article list and detail pages, i.e. the Learn tab, plus the article card in the home page.

Also replaces the placeholder article card on the home view with a random article.  Since there's no longer a "View all" expand-in-place link on the home view, and that was the only place that the Handlebars article template was used, I removed it and the related javascript functions.

Note/question, probably for @lederer : do the pieces of Article need to change?  In the old version we had title, 'teaser', and 'content'.  The teaser was shown on the card view and content on the detail view.  Now we have 'aside' and the card views in the prototype show the first few lines of the content.  Do we need to get rid of 'teaser' and add 'aside' to the articles model?  And how do we know when to cut of the content shown on the card view?

Also on the subject of articles, the example article detail view has an image in the middle of the text, which also isn't possible without adding something to the model to specify more structure.

I can make an issue for these if the answers aren't obvious.

List:
articles_list
Detail:
article_detail

Resolves #543, #547, and #537.


lederer:
Thanks @KlaasH.

The new `aside` is basically the old `pull-quote` ([example](https://www.gophillygo.org/tips-and-tricks/winterbikecommute2/)).

And the new inline image is just placeholder for any sort of inline ~embed. For example the [video in this one](https://www.gophillygo.org/community-profile/connor-barwin-gophillygo/).

IIUC the teaser seems mostly redundant with the title, so I reckon we can drop that. We can use the title and truncated content for the card. Truncate conservatively — ie, leave plenty leftover and just use `text-overflow: ellipsis;` etc to make it look okay. I can fine tune that once it’s working

flibbertigibbet:
build again

python/cac_tripplanner/cms/models.py
```diff
@@ -27,12 +27,7 @@ def random(self):
         """Returns a randomized article"""
         # Need to use the full object, because there is a magic transformation of the URL
         # at some point which is needed for assembling the s3 url.
-        randomized = self.published().order_by('?')[:1]
-
-        if randomized:
-            return randomized[0]
-        else:
-            None
+        return self.published().order_by('?').first()
```
flibbertigibbet: :sunglasses:

python/cac_tripplanner/cms/views.py
```diff
@@ -11,67 +11,54 @@
...
-                                          slug=slug)
-    context = {'article': community_profile, 'debug': DEBUG}
-    return render(request, 'community-profile-detail.html', context=context)
-
-
-def tips_and_tricks_detail(request, slug):
-    """Tips and tricks detail view
+def learn_detail(request, slug):
+    article = get_object_or_404(Article.objects.published(), slug=slug)
+    more_articles = Article.objects.published().order_by('-publish_date').exclude(pk=article.pk)[:3]
```
flibbertigibbet: Probably need to guard here against not having any articles published.

python/cac_tripplanner/templates/partials/article-detail.html
```diff
@@ -0,0 +1,19 @@
+<article class="info-article">
+    <header class="info-article-header">
+        <div class="info-article-image-hero"style="background-image: url('{{ article.wide_image.url }}');"></div>
```
flibbertigibbet: The wide image at the top of the detail view gets cut off, only showing the left half or so. Setting the height to 200px instead of 400px seems to do the trick.

flibbertigibbet:
Tests pass locally. Jenkins should be happy again once the fixes on develop are merged back in #562.

KlaasH:
Made the tweaks mentioned above, rebased to get the stuff merged into redesign from develop, and punted on article card layout (https://github.com/azavea/cac-tripplanner/issues/570).  Good to go?

PR #569, opened 2016-10-31
Break up map controller

KlaasH:
Work-alike refactor to move the isochrone and itinerary functionality of the map into their own controllers, to make the whole setup easier to understand and work with.  They're pretty well self-contained conceptually and the seams in the code were not hard to separate.

This is hard to test on `redesign`, since none of this behavior is wired up yet.  I did the work on `develop` where I could try it out then cherry-picked the finished changes to `redesign` and there was only one conflict (the addition, in `redesign` of `var layerControl = null;` right below some lines I had deleted.  So it might cherry-pick back to `develop` cleanly at the moment.  Or I could push my `develop` branch for reference.

### DIFF
--- a/src/app/scripts/cac/map/cac-map-control.js
+++ b/src/app/scripts/cac/map/cac-map-control.js
@@ -6,9 +6,6 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
         homepage: true,
         center: [39.95, -75.1667],
         zoom: 14,
-        selectors: {
-            destinationPopup: '.destination-directions-link'
-        }
     };
 
     var map = null;
@@ -24,7 +21,6 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
 
     var events = $({});
     var eventNames = {
-        destinationPopupClick: 'cac:map:control:destinationpopup',
         currentLocationClick: 'cac:map:control:currentlocation',
         originMoved: 'cac:map:control:originmoved',
         destinationMoved: 'cac:map:control:destinationmoved',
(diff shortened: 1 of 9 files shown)


src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -84,25 +45,13 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
...
-    var pendingIsochroneRequest = null;
-
     function MapControl(options) {
         this.events = events;
         this.eventNames = eventNames;
         this.options = $.extend({}, defaults, options);
         overlaysControl = new CAC.Map.OverlaysControl();
         map = new cartodb.L.map(this.options.id, { zoomControl: false })
-            .setView(this.options.center, this.options.zoom);
+                           .setView(this.options.center, this.options.zoom);
```
flibbertigibbet: spacing

KlaasH: Are you saying this should be further to the right so it aligns under the paren after `map` (as [here] (https://github.com/azavea/cac-tripplanner/commit/ad89bca928bbe6ecabbed8ab2cba7bb622e6bc9e#diff-bfd83b9efe49a865ad1b7a51733c0b66R334) or that it should be back under the `=` where I moved it from?  My understanding is that when breaking lines on a chained method call it's good to align the dots, though in this case `setView` is the first method on `map`, so it's true that lining it up under the `L.map` dot isn't quite correct.

flibbertigibbet: Oh, I thought maybe you'd accidentally put in an extra tab there.

For parentheses and brackets we'll regularly align with the line above, but our handling of dot methods varies. For several lines of them, I think indenting them all by some regular spacing multiple to be the easiest to read and maintain. This is fine.

Issue #571, opened 2016-10-31
Style typeahead input


flibbertigibbet:
On adding back our autocompleting Typeahead inputs to the home page form, the width of the input box shortens and there's no solid background for the dropdown of suggestions.

![image](https://cloud.githubusercontent.com/assets/960264/19864091/2125cf4a-9f6d-11e6-85c2-13015e23dbdd.png)

KlaasH:
Also the smaller version, with its different color configuration (light text against the blue, currently white on white).

![small_directions_form_invisible_text](https://cloud.githubusercontent.com/assets/6598836/19930829/8555f82c-a0e0-11e6-9797-422495c042a9.png)

flibbertigibbet:
Yes, I'm aware. I created #571 for the typeahead styling, which will require design support. Currently working on a branch with fixes enough to make the sidebar functional and the dropdown readable.

PR #573, opened 2016-10-31
Feature/directions form

flibbertigibbet:
Wires up the directions form controls on the home page:
  - Adds back address autocomplete (needs styling; see #571)
  - Wires up new geolocation button
  - Updates handing of mode selection and toggling classes on the new mode buttons
  - Transitions to the map view when both 'from' and 'to' fields completed (does not query for directions)

### DIFF
--- a/src/app/scripts/cac/control/cac-control-bike-mode-options.js
+++ b/src/app/scripts/cac/control/cac-control-bike-mode-options.js
@@ -88,7 +88,7 @@ CAC.Control.BikeModeOptions = (function ($) {
         });
 
         // slice off trailing comma
-        mode = mode ? mode.substr(0, mode.length-1) : options.defaultMode;
+        mode = mode ? mode.substr(0, mode.length - 1) : options.defaultMode;
         return mode;
     }
 
--- a/src/app/scripts/cac/control/cac-control-bike-mode-options.js
+++ b/src/app/scripts/cac/control/cac-control-bike-mode-options.js
@@ -25,10 +25,12 @@ CAC.Control.BikeModeOptions = (function ($) {
                 triangleTimeFactor: 0.17
             }
         },
+        defaultMode: 'WALK,TRANSIT',
+        // map button class names to OpenTripPlanner mode parameters
         modes: {
-            // used to suffix to mode inputs selector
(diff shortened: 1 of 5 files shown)


src/app/scripts/cac/control/cac-control-bike-mode-options.js
```diff
@@ -67,12 +69,26 @@ CAC.Control.BikeModeOptions = (function ($) {
...
      *
-     * * @param modeSelectors {String} jQuery selector like '#someId input'
+     * @param modeSelectors {String} jQuery selector like '#someId input'
+     * @returns {String} comma-separated list of OpenTripPlanner mode parameters
      */
     function getMode(modeSelectors) {
-        var mode = $(modeSelectors + options.modes.walkBike + ':checked').val() || 'WALK';
-        var transit = $(modeSelectors + options.modes.transit).prop('checked');
-        mode = transit ? 'TRANSIT,' + mode : mode;
+        var mode = '';
```
KlaasH: A more heavily lodashified version of this function:
```
var $selected = $(modeSelectors);
if (!$selected) {
    console.error('no mode controls found');
    return options.defaultMode;
}
var mode = _(options.modes).filter(function(val, key) { return $selected.hasClass(key); }).join(',');
return mode || options.defaultMode;
```

flibbertigibbet:
updated

src/app/scripts/cac/pages/cac-pages-home.js
```diff
@@ -158,7 +205,7 @@ CAC.Pages.Home = (function ($, BikeModeOptions,  MapControl, Templates, UserPref
         event.preventDefault();
 
         // hide existing destinations list and show loading spinner
-        $(options.selectors.destinationsContainer).addClass('hidden');
+        $(options.selectors.placeList).addClass('hidden');
```
KlaasH: There's no "View all destinations" link on the redesigned home page.  Is there a reason to keep this function (and the related selector and listener) around?

flibbertigibbet: There is a bunch of old code that has been retained. I've been holding off on purging things generally until some more of the new functionality is stable, as some of it we do want or will break something else if removed immediately.

PR #577, opened 2016-11-01
Feature/place detail

### LINKED ISSUE #544: Implement 'More Info' for location
When user clicks 'More Info' button on a location in the home page list, open detail view. Update Handlebars template for details.

CloudNiner:
## Overview

Implements the 'place-detail' view, as part of the explore section. Adds relevant links to home page.

## Demo

cac-place-detail

Closes #544

### DIFF
--- a/python/cac_tripplanner/templates/partials/destination-card.html
+++ b/python/cac_tripplanner/templates/partials/destination-card.html
@@ -1,0 +1,9 @@
+<a href="{% url 'place-detail' pk=destination.pk %}" class="preview-card-link">
+    <div class="preview-card">
+        <img class="preview-photo" src="{{ destination.wide_image.url }}" width="308" height="204" alt="{{ destination.name }}" />
+        <div class="preview-details">
+            <h2>{{ destination.name }}</h2>
+            <p class="lede">{{ destination.description|striptags|truncatewords:20 }}</p>
+        </div>
+    </div>
+</a>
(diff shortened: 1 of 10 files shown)


python/cac_tripplanner/templates/partials/destination-card.html
```diff
@@ -0,0 +1,9 @@
+<a href="{% url 'place-detail' pk=destination.pk %}" class="preview-card-link">
+    <div class="preview-card">
+        <img class="preview-photo" src="{{ destination.wide_image.url }}" width="308" height="204" alt="{{ destination.name }}" />
+        <div class="preview-details">
+            <h2>{{ destination.name }}</h2>
+            <p class="lede">{{ destination.description|striptags|truncatewords:20 }}</p>
```
flibbertigibbet: Yeah, I like this. If we need to do anything more complex in future I think it would make sense to move it out.

CloudNiner:
Updated with tests, refactored to move place_detail to destination app, think it makes more sense there since its displaying Destination objects.

CloudNiner:
Yes, it is cut off a bit. I left this as is because I wasn't sure exactly what the plan was for images moving forward. If we update the image dimensions for users to upload, and leave the styling as is in prototype, we make the old images somewhat backwards-incompatible with the redesign of the site. If we don't make any changes to the db model, then we should probably consider changing the size of these hero images across the board. Figure this could be addressed later in a styling round when we have more direction.

Or I can hack something in now. Doesn't matter to me.

flibbertigibbet:
We've modified image sizes that were specifically set in the prototype in a couple of other places to correct display for our existing image sizes, and it seems to work pretty well for their relative ratios. Here it looks like the image size wasn't explicitly set already in the prototype, so it's fine with me if you don't wan to mess with it.

CloudNiner:
Ok, makes sense. I'll go ahead and modify the place detail view to make the hero image 400px high then so it looks nicer and aligns with current images.

Issue #539, opened 2016-10-20
Wire up map page directions list


flibbertigibbet:
Handle user-entered addresses. Do not wire up options hamburger menu, but do include mode options selectors. Update Handlebars template for itinerary options, but not the step-by-step directions.

flibbertigibbet:
No longer relevant, since map page form is the same as the home page form.

PR #580, opened 2016-11-02
Feature/sidebar directions

### LINKED ISSUE #538: Wire up home page from/to form
And transition to map page prototype page

flibbertigibbet:
Closes #538.

Mostly this PR consists of refactoring:
  - Move directions form logic from home page to the directions control
  - Rename sidebar directions controller to just directions control (not always in a sidebar)
  - Rename bike mode options controller to just mode options control (handles all mode options)

Fixes typeahead styling enough to be usable.

Implements origin/destination reversal button.

Fixes the mode toggle buttons (would toggle away if user clicked current selection).

Temporarily disables URL route updating and loading saved options from local storage. URL routing uses the `map` page, which we are no longer using but haven't removed yet.  Since the home page directions form has no 'go' button, as soon as an origin and destination fields are completed, the app transitions to the map page, which has no way to go back to the home page, so if origin and destination are loaded from storage, it causes the app to jump immediately to the map page.

Gets far enough to request directions and load itineraries successfully, but much work remains to be done around directions. This PR should set things up well to proceed with #539.

### DIFF
--- a/src/app/scripts/cac/control/cac-control-directions.js
+++ b/src/app/scripts/cac/control/cac-control-directions.js
@@ -17,13 +17,18 @@ CAC.Control.Directions = (function (_, $, Control, ModeOptions, Geocoder, Routin
 
     var defaults = {
         selectors: {
+            // directions form selectors
+            directionsForm: '.directions-form-element',
+            directionsFrom: '.directions-from',
+            directionsTo: '.directions-to',
 
-            selectedModes: '.mode-option.on',
-            modes: '.mode-option',
+            // typeahead
+            typeaheadFrom: '#input-directions-from',
+            typeaheadTo: '#input-directions-to',
 
-            typeaheadDest: '#input-directions-to',
-            typeaheadOrigin: '#input-directions-from',
-
+            // top-level classes
+            homePageClass: 'body-home',
+            mapPageClasses: 'body-map body-has-sidebar-banner',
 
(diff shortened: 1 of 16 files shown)


src/app/scripts/cac/control/cac-control-directions.js
```diff
@@ -386,6 +405,25 @@ CAC.Control.SidebarDirections = (function (_, $, Control, BikeModeOptions, Geoco
...
+        if (key === 'origin') {
+            prefKey = 'origin';
+            $input = $(options.selectors.directionsFrom);
+        } else if (key === 'destination') {
+            prefKey = 'destination';
+            $input = $(options.selectors.typeaheadTo);
+        } else {
+            console.error('unrecognized typeahead key ' + key);
+            return;
+        }
```
KlaasH: What is this for?  Since neither `$input` nor `prefKey` are used below, it seems like it's not doing anything except validating that the key is one of `origin` or `destination`.

flibbertigibbet: That is correct, it's just for sanity checking the typeahead keys set in the HTML.

flibbertigibbet:
Yes, the directions fetch needs some work. Once default preference loading is re-enabled, it should set some default options for bike directions, so that can work before we get the custom controls in place.

PR #582, opened 2016-11-03
Closes #547, Closes #548: Implement article card buttons

### LINKED ISSUE #547: Click article on home page to view details

### LINKED ISSUE #548: Click button on article to view all articles
Go to list of all articles from button within article on home page.

CloudNiner:
## Overview

Nested link tags were causing trouble with button clicks. Reworked the HTML structure a bit for preview cards after some discussion with @lederer. See commit log for details, this version is no JS and HTML5 compliant.

Also makes some minor tweaks to the preview card, including showing the teaser html inline (not wrapped in a `<p>` tag) and only show the teaser, not the entire content.

## Demo

cac-home-link-learn-detail
cac-home-link-learn-large
cac-home-link-learn-list

### DIFF
--- a/src/app/styles/components/_preview-card.scss
+++ b/src/app/styles/components/_preview-card.scss
@@ -1,6 +1,6 @@
-.preview-card-link {
-    @include delinkify;
+.preview-card {
     display: flex;
+    position: relative;
     flex: 1 0 auto;
     flex-flow: column nowrap;
     align-items: stretch;
@@ -16,7 +16,8 @@
     }
 }
 
-.preview-card {
+.preview-card-link {
+    @include delinkify;
     display: flex;
     flex: 1 0 auto;
     flex-flow: row nowrap;
@@ -55,6 +56,7 @@
         flex-flow: column nowrap;
         align-items: stretch;
         justify-content: space-between;
+        margin-bottom: 5rem;
     }
 
     .category-heading {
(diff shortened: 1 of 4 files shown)


src/app/styles/components/_preview-card.scss
```diff
@@ -55,6 +56,7 @@
         flex-flow: column nowrap;
         align-items: stretch;
         justify-content: space-between;
+        margin-bottom: 5rem;
```
lederer: Since the More link is absolutely positioned instead of inline, I can't think of any way to coax the browser into calculating the layout we want. But we can do it in SASS using a variable.

Looks like the height of `.goto-more-posts` is currently `line-height: 2.75` ✖️ `font-size: 1.4rem` = 3.85rem. We could hardcode a `$goto-more-posts-btn-height` variable to 3.85rem (or round to 4 for some slack), then set the card's `margin-bottom` to, say, `$goto-more-posts-btn-height + 2*$home-section-padding`

lederer:
Nice work.

One nit: The More link is too narrow in the wider viewport example you included (included below). It should be the full width of the textual content area. I reckon this might get tricky due to the absolute positioning of the button. Happy to pair up to see if we can pull it off.

cac-home-link-learn-large

src/app/styles/components/_preview-card.scss
CloudNiner: Makes sense to me. Unfortunately `$home-section-padding` is in px and that value is used in a bunch of calculations elsewhere as px. So I guess this stays as is for now.

lederer: Not sure how SASS handles mixed units, but you can totally drop those mixed units in a CSS `calc()` function and it should _just work_. 

Something like:
`margin-bottom: calc(#{oto-more-posts-btn-height} + 2 * #{$home-section-padding});`

CloudNiner: Yeah, knew about the unit mixing, but didn't know that about calc. Can give that a shot.

CloudNiner:
Updated to address full width button:

101315_nlds_schwarber_homerun_statcast_medres_bq4qks7r

Oh sorry, thats the wrong screenshot. Here's the right one:
cac-home-link-full-width-button

lederer:
🔥

PR #596, opened 2016-11-07
Closes #551: Reimplement JotForms subscribe form in footer

### LINKED ISSUE #551: wire up email subscription form
Re-implement JotForms email subscription form.

Note this will require using `jquery.noConflict()`, as JotForms conflicts with jQuery.

CloudNiner:
The First/Last name fields were left in, but hidden, to remain
compatible with existing form.

Added email validation on keyup so that the subscribe button
changes state while typing. It felt weird to have to click
out of the input box to enable the subscribe button.

Seemingly no issues with re-enabling jQuery.noConflict().

Added some basic styling for subscribe button disabled state.

## Demo

cac-jotform-1
cac-jotform-2

### DIFF
--- a/python/cac_tripplanner/templates/home.html
+++ b/python/cac_tripplanner/templates/home.html
@@ -230,7 +230,7 @@
 
 {%block jspage %}
 <script type="text/javascript">
-//jQuery.noConflict(); // because JotForms.
+jQuery.noConflict(); // because JotForms.
 jQuery(document).ready(function ($) {
     CAC.Settings.fbAppId = '{{ fb_app_id }}';
     CAC.Settings.routingUrl = '{{ routing_url }}';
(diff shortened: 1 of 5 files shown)


python/cac_tripplanner/templates/home.html
```diff
@@ -230,7 +230,7 @@
 
 {%block jspage %}
 <script type="text/javascript">
-//jQuery.noConflict(); // because JotForms.
+jQuery.noConflict(); // because JotForms.
```
KlaasH: This takes care of the home view, but the footer is used on other views (i.e. Learn list and detail), so they're getting console errors "prototype.forms.js:591 Uncaught TypeError: element.dispatchEvent is not a function" and "prototype.forms.js:575 Uncaught TypeError: element.attachEvent is not a function".

CloudNiner:
RE: the flashing subscribe button. I went for the debounce, as it should generally solve the issue and keeps us from eliminating technically correct addresses.

PR updated.

KlaasH:
Looks good, though this project has lodash, so the hand-crafted debounce isn't strictly necessary.

CloudNiner:
Whoops. Updated to `_.debounce`.

PR #597, opened 2016-11-07
Closes #541: Show modal for directions options

### LINKED ISSUE #541: Show modal for directions option
When hamburger icon clicked. Prototype will be updated, so just display dummy modal.

CloudNiner:
Left the Modal control very generic for now. Added
a few source TODOs for things to implement once additional
logic/wireframes are available.

## Demo

cac-options-modal

### DIFF
--- a/python/cac_tripplanner/templates/base.html
+++ b/python/cac_tripplanner/templates/base.html
@@ -118,6 +118,7 @@
     <script src="{% static 'scripts/main/cac/map/cac-map-overlays.js' %}"></script>
     <script src="{% static 'scripts/main/cac/map/cac-map-templates.js' %}"></script>
     <script src="{% static 'scripts/main/cac/control/cac-control-mode-options.js' %}"></script>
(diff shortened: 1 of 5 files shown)


KlaasH:
Kind of a bummer that this is necessary, but I believe adding the new file to `src/karma/karma-dev.conf.js` and `src/karma/karma-coverage.conf.js` will placate Jenkins.

There are more pieces of the modals at https://github.com/azavea/cac-tripplanner/blob/redesign/src/prototype/app/extras.html (so http://localhost:3000/extras.html when running grunt to serve the prototype).  But issue #541 just calls for a dummy modal, which this accomplishes just fine.

:+1: with addition to karma files and assuming further modal functionality is for later.

CloudNiner:
Ah. I didn't see those additional modal mockups. After looking at those, this should be good for most of them. Any modals with more advanced functionality can just extend the base `Modal` control. 

Saw the jenkins failure, thanks for the hint, that seemed to do it.

Issue #599, opened 2016-11-09
Change itinerary card sidebar on hover


flibbertigibbet:
Prototype shows a different SVG on the left of the itinerary summary card for the highlighted itinerary.

Issue #598, opened 2016-11-09
Add missing mode icons


flibbertigibbet:
New fontello font icons are missing icons for several transportation modes.

Needed are:

 - bus
- subway
- car (should not be visible, but good to have to debug if we ever got car directions)
- rail (not the same as 'train', which we have)
- tram
- ferry

flibbertigibbet:
Also missing some directions icons:

 - continue (was using an up-pointing arrow)
 - circle clockwise or counterclockwise (was using a `repeat` icon)
 - elevator (was using a cloud-upload icon, hopefully we can find something more appropriate)
 - default icon (was the glyphicon `remove-circle`, not sure what that looks like)

PR #600, opened 2016-11-09
Feature/sidebar directions

flibbertigibbet:
Implements the itinerary summary list in the sidebar.

The summary cards will need a little styling help, as the text in practice overflows the expected one line of text. The formatted duration string isn't split into numeric and unit portion, so the additional formatting for the unit is applied to the whole string. Using moment's `humanize` function now, which can return numberless strings such as "an hour".

Created #598 for the missing mode icons (just use a train for now).

Created #599 for card highlight interactivity (perhaps can be handled via CSS).

image

Also updates some date/time and duration formatting, adds icons for origin and destination markers, fixes issues with itinerary map plotting, and shows map controls once directions return (map controls are hidden on home view).

### DIFF
--- a/src/app/scripts/cac/map/cac-map-templates.js
+++ b/src/app/scripts/cac/map/cac-map-templates.js
@@ -201,23 +201,32 @@ CAC.Map.Templates = (function (Handlebars, moment, Utils) {
 
     // Template for itinerary summaries
     function itineraryList(itineraries) {
-        var source = '{{#each itineraries}}' +
-                '<div class="block block-itinerary" data-itinerary="{{this.id}}">' +
-                    '<div class="trip-numbers">' +
-                        '<div class="trip-duration"> {{this.formattedDuration}}</div>' +
-                        '<div class="trip-distance"> {{this.distanceMiles}} mi</div>' +
-                    '</div>' +
-                    '<div class="trip-details">' +
-                        '{{#each this.modes}}' +
-                            '<div class="direction-icon">' +
-                                ' {{modeIcon this}}' +
-                            '</div>' +
-                        '{{/each}}' +
-                        '<span class="short-description"> via {{this.via}}</span>' +
-                        '<a class="itinerary" data-itinerary="{{this.id}}"> View Directions</a>' +
-                    '</div>' +
-                '</div>' +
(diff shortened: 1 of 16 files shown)


src/app/scripts/cac/map/cac-map-templates.js
```diff
@@ -320,7 +325,9 @@ CAC.Map.Templates = (function (Handlebars, moment, Utils) {
         });
 
         Handlebars.registerHelper('datetime', function(dateTime) {
-            return new Handlebars.SafeString(new Date(dateTime).toLocaleTimeString());
+            // round to the nearest minute
```
CloudNiner: Is this comment correct? Isn't the string below essentially taking the ceiling of the current minute?

flibbertigibbet: This was the [most upvoted approach for moment.js rounding](http://stackoverflow.com/a/28561371), although apparently it has a problematic edge case when it should roll over. You'd think this would be made easy. I'll dig some more for another approach.

flibbertigibbet:
If you have no transit times, my guess is you probably have an outdated GTFS.

src/app/scripts/cac/map/cac-map-templates.js
CloudNiner: Ah ok. Maybe the comment should just be adjusted to `// round up to the nearest minute` to be a bit clearer.

Providing an upper bound on the estimate makes sense to me, as its better to overshoot than undershoot. If the estimate is 12.0 minutes and that estimate gets rounded up in the edge case to 13 minutes, is that a big deal? Doesn't seem like it to me?

flibbertigibbet: I think the issue is that 60+ minutes would get rounded up to 61 minutes instead of rolling over to an hour. I'm going to give [moment-round](https://github.com/WebDevTmas/moment-round) a shot.

flibbertigibbet: Rounding library proved problematic, so modified to just round the Unix timestamp before conversion.

CloudNiner:
👍

Issue #606, opened 2016-11-11
Missing departure and arrival address text


flibbertigibbet:
In the itinerary data passed to the step-by-step directions template, the origin and destination address text has gone missing.

Possibly related to user preferences having been disabled.

flibbertigibbet:
This got fixed at some point, probably with updates to `UserPreferences` usage.

PR #607, opened 2016-11-11
Feature/step by step

### LINKED ISSUE #540: Step by step directions
Update Handlebars template for step by step directions.

Update iconography used for turn directions and reformat time/distance rounding.

flibbertigibbet:
Implement step-by-step directions for redesign. Closes #540. On clicking an itinerary in the sidebar, displays step-by-step directions. Back button returns to the itinerary list.

Note that several of the mode and turn direction icons aren't in yet (#598).

Created #606 for the missing departure and arrival addresses.

image

### DIFF
--- a/python/cac_tripplanner/templates/home.html
+++ b/python/cac_tripplanner/templates/home.html
@@ -91,7 +91,8 @@
 
     <div class="directions-step-by-step">
         <header class="step-by-step-header">
-            <button name="back-to-directions-results" class="back-to-directions-results"><i class="icon-left-big"></i></button>
+            <button name="back-to-directions-results" class="back-to-directions-results">
+            <i class="icon-left-big"></i></button>
             <h1>Directions</h1>
             <button name="share-directions" class="share-directions"><i class="icon-share"></i></button>
         </header>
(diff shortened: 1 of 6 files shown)


KlaasH:
Hard to look at template changes for too long without it all becoming a blur, but what I managed to absorb of the code changes look good.  A few functionality notes/questions:

- Is it supposed to be possible to pick which itinerary to follow on mobile?  All three lines show up on the map but when the list becomes a bar across the bottom only the first shows up and I don't see a way to select a different one.  Or is the plan for the lines on the map to be clickable and activate the corresponding itinerary (or bring its bar to the top so it can be activated)?

- I imagine this should be a separate issue, but I'll note it here: when I work around #608 and get bike directions, the "Check agency bike policy" note looks like this:
bike_policy_note

- After I select an itinerary then click 'Back', the draggable interactivity is still active on the route I had selected.

flibbertigibbet:
- There has been some discussion with design around how to change itineraries on mobile. We don't have anything set on that yet. I think the idea is that the user can swipe right/left to change, though that functionality isn't obvious, so we should have some other way to change the selected itinerary as well. I think we can just punt on that for now.

 - Our existing bike policy/other trip message modal is going away, but we don't have the replacement banner HTML snippet yet.

 - I've also noticed some strange behavior around draggable itineraries, including the old itinerary remaining visible after drag start. Haven't made issues for any of it yet, as it hasn't been obvious exactly what's going on.

PR #615, opened 2016-11-14
Feature/fix bike route layer

### LINKED ISSUE #611: Fix bike routes layer
Throws console error on attempting to add to map.

flibbertigibbet:
Fixes #611.

cartodb.js uses jQuery, which we run in `noConflict` mode for JotForms, which is necessary because JotForms uses prototype.js. Consequently, references to `$.` for jQuery in third-party libraries such as CartoDB will fail.

This fixes the specific case of references to `$.ajax`, as a global search-and-replace of `$.` for `jQuery.` caused other issues.

### DIFF
--- a/python/cac_tripplanner/templates/base.html
+++ b/python/cac_tripplanner/templates/base.html
@@ -92,7 +92,7 @@
     }(document, 'script', 'facebook-jssdk'));</script>
     {% endif %}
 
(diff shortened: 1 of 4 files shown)


CloudNiner:
+1

If we have to fix anything else related to this it might be worth looking at replacing JotForms with some other tool.

PR #616, opened 2016-11-14
Feature/mobile map load

### LINKED ISSUE #587: Do not load map initially in mobile view
When first loading home page, map is hidden on mobile view.

Postpone initializing map and loading tiles until map becomes visible for the first time.

flibbertigibbet:
Closes #587.

Delay loading map tiles in mobile view until map becomes visible, either by window resize, or by going to directions view.

### DIFF
--- a/src/app/scripts/cac/map/cac-map-control.js
+++ b/src/app/scripts/cac/map/cac-map-control.js
@@ -3,6 +3,7 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
 
     var defaults = {
         id: 'map',
+        selector: '#map',
         homepage: true,
         center: [39.95, -75.1667],
         zoom: 14,
@@ -32,7 +33,8 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
     var tabControl = null;
     var zoomControl = null;
 
-    var homepage = true; // whether currently displaying home page view TODO: rework
+    var homepage = true; // whether currently displaying home page view
+    var loaded = false; // whether map tiles loaded yet (delay on mobile until in view)
 
     var esriSatelliteAttribution = [
         '&copy; <a href="http://www.esri.com/">Esri</a> ',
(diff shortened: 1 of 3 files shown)


src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -49,29 +51,17 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
...
-        initializeBasemaps();
-
-        if (!homepage) {
-            // hide zoom control on home page view
-            zoomControl.addTo(map);
-            // delay loading overlays and layer switcher
-            initializeOverlays();
-            initializeLayerControl();
+        // only load map tiles if map visible
+        if (!homepage || $(this.options.selector).is(':visible')) {
```
CloudNiner: It _may_ be better for this condition to check page width rather than visibility of the map div? I could see the styles changing in the future and this silently breaking, whereas if we check for page width it would always work.

src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -94,6 +88,49 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
...
+    function isLoaded() {
+        return loaded;
+    }
+
+    /**
+     * Load base map tiles and set map on associated controls.
+     *
+     * @param {Object} ctl Reference to this control (useful if calling from elsewhere)
+     */
+    function loadMap(ctl) {
```
CloudNiner: Could consider just calling this method using `apply()` instead of explicitly requiring the `ctl` argument. 

It feels a bit odd to have a prototype method require a reference to itself as the first argument. But that's as much of a way these files are setup as anything.

src/app/scripts/cac/map/cac-map-control.js
flibbertigibbet: This seemed the most resilient option to me, as it's independent of where the page width breaks happen. If the map selector changes, it would be pretty obvious that we'd need to update that selector in our code.

src/app/scripts/cac/map/cac-map-control.js
flibbertigibbet: I'm not a fan of passing an argument to `this` either, but yay JavaScript classes. I'll try making it cleaner with `apply`.

flibbertigibbet:
I could take a stab at delaying the cartodb.js script fetch in a separate PR.

CloudNiner:
Up to you. I just remember a conversation about that and wasn't sure if it was supposed to be part of this task or not. It may make sense to defer working on the script load in order to get the remaining map interactions working again?

flibbertigibbet:
We did talk about it in carding, but I hadn't mentioned the script load delay on the issue, so forgot about it. If it's not easy, I'll just make another issue for that.

PR #617, opened 2016-11-15
Closes #609: Move modal selectors to options

### LINKED ISSUE #609: Move selectors out of modal control
All jQuery selectors used in this project are to be declared in the `options.selectors` property at the head of the module. Currently `CAC.Control.Modal` does not do this.

CloudNiner:


CloudNiner:
As noted by the comment for `modalClass`, it is a required argument in order for the modal control to work as expected, since it has to know what HTML to bind to. It's separate from the selectors object to differentiate this. In all other cases that I see, the selectors object seems to be more of a way to keep track of all the internal selectors that a control uses. 

The remaining concatenation of classes is an artifact of the way the CSS classes are setup. I don't think it makes sense to set those as selector options, since that concatenation is more internal view logic than a setting.

I did move the body class concatenation to a function so its easier to see whats going on there and modify later if necessary.

Issue #585, opened 2016-11-03
Load all places on home page


flibbertigibbet:
Instead of random four

flibbertigibbet:
Closed in #619

jbranigan:
It looks like a number of these might not be marked as active in the back end, and result on 404's when they're clicked on. Filtering to active - reopening issue.

Issue #589, opened 2016-11-03
Update stored user preferences


flibbertigibbet:
Related to #588, which syncs with local storage.

With redesign, do not store/use origin or destination in local storage.

PR #619, opened 2016-11-15
Restore url routing for directions

### LINKED ISSUE #574: Remove City Hall as default origin
Just leave blank.

KlaasH:
Re-enables reading and updating of URL parameters for interactive directions.

Also converts the UserPreferences component to just store values in a variable rather than in user storage, as a very rough first pass at issue #589.  Maybe this will be fine, maybe we still want to put some things in local storage, maybe things break, or maybe we want to go further and not use UserPreferences as an interface for sharing parameters between components.  Anyway, doing this was enough to get URL routing working.

I also moved the `home` view from the `cms` app to the `destinations` app because it seemed to fit in better there, and spruced up the shared view logic in `destinations/views.py` a bit.

And since I was in UserPreferences, I removed the City Hall default (resolves #574).

**To test**
- Following a link to directions should take you to the directions, e.g. http://localhost:8024/?origin=39.9616133%2C-75.1541914&originText=990%20Spring%20Garden%20St%2C%20Philadelphia%2C%20Pennsylvania%2C%2019123&mode=WALK%2CTRANSIT&destination=39.96581%2C-75.183494&destinationText=Fairmount%20Waterworks%20Interpretive%20Center
- Following the same link in the old format (with `map/directions` at the beginning) should redirect you to `/` and work, e.g. http://localhost:8024/map/directions?origin=39.9616133%2C-75.1541914&originText=990%20Spring%20Garden%20St%2C%20Philadelphia%2C%20Pennsylvania%2C%2019123&mode=WALK%2CTRANSIT&destination=39.96581%2C-75.183494&destinationText=Fairmount%20Waterworks%20Interpretive%20Center
- Since there's no more local storage, clearing the URL (i.e. going to http://localhost:8024/) should get you back to the home view with empty origin/destination and the default mode options.

### DIFF
--- a/python/cac_tripplanner/destinations/views.py
+++ b/python/cac_tripplanner/destinations/views.py
@@ -1,54 +1,37 @@
 from datetime import datetime
-import json
-
 from pytz import timezone
+import json
+import requests
 
+from django.conf import settings
 from django.contrib.gis.geos import GEOSGeometry, Point
 from django.forms.models import model_to_dict
 from django.http import HttpResponse
-from django.shortcuts import get_object_or_404, render, render_to_response
-from django.template import RequestContext
+from django.shortcuts import get_object_or_404, render
 from django.views.generic import View
-from cac_tripplanner.settings import DEBUG
 
-import requests
-
-from cac_tripplanner.settings import FB_APP_ID, HOMEPAGE_RESULTS_LIMIT, OTP_URL, DEBUG
 from .models import Destination, FeedEvent
-
-ROUTING_URL = OTP_URL.format(router='default') + 'plan'
 
 
 DEFAULT_CONTEXT = {
-    'debug': DEBUG,
-    'fb_app_id': FB_APP_ID,
-    'routing_url': ROUTING_URL
+    'debug': settings.DEBUG,
+    'fb_app_id': settings.FB_APP_ID,
+    'routing_url': settings.ROUTING_URL
 }
 
 
-def base_otp_view(request, page):
+def base_view(request, page, context):
     """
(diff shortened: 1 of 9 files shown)


KlaasH:
Build failure was from an insufficiently escaped regex. I rebased the fix in.

python/cac_tripplanner/destinations/views.py
```diff
@@ -1,54 +1,59 @@
...
+        context = {'tab': 'map'}
+    elif request.GET.get('origin') is not None:
+        # If there's no destination but there is an origin, go to Explore
+        context = {'tab': 'explore'}
+    else:
+        # Otherwise show the home view
+        # Show one random article
+        article = Article.objects.random()
+        # Show all destinations, in random order
+        destinations = Destination.objects.order_by('?').all()
```
KlaasH: The note on `place_detail` reminded me: this resolves issue #585.

python/cac_tripplanner/destinations/views.py
```diff
@@ -58,7 +63,15 @@ def directions(request):
     :param request: Request object
     :returns: A rendered response
     """
-    return base_otp_view(request, 'directions.html')
+    return base_view(request, 'directions.html')
+
+
+def place_detail(request, pk):
+    destination = get_object_or_404(Destination.objects.published(), pk=pk)
+    more_destinations = Destination.objects.published().order_by('?').exclude(pk=destination.pk)[:3]
```
KlaasH: Not something I remember off the top of my head, but I just tried in shell_plus and learned that, while index lookups get grumpy when they're out of range, slices are forgiving.

flibbertigibbet:
build again

Issue #586, opened 2016-11-03
Implement icon link back to home view from map


flibbertigibbet:
Handle state transition and hide map components.

flibbertigibbet:
Closed in #626

Issue #556, opened 2016-10-26
Re-implement tab control


flibbertigibbet:
Currently disabled as part of redesign.

Should set and manage tabs from home and map views.

Now have three tabs instead of two: have added 'learn' tab.

PR #624, opened 2016-11-15
Back to home via full page refresh

CloudNiner:
This feels ridiculously trivial, but now that the routing stuff is (mostly) setup, this is all that's necessary. If we wanted to trigger a back operation without a full page refresh I think we're in for a much larger world of hurt. 

That would involve the following I think, if I'm following what's going on correctly:
- Resetting the homepage flag in map control
- Destroying the map elements manually
- Removing any added event handlers
- Triggering a change to the urlRouter
That doesn't feel very easy, since we can't hook into any destroy hooks throughout the application -- most of the jQuery elements on the page aren't destroyed in transition, they're just hidden and shown. 

Thoughts?

### DIFF
--- a/python/cac_tripplanner/templates/partials/header.html
+++ b/python/cac_tripplanner/templates/partials/header.html
@@ -3,7 +3,7 @@
         <div class="slogan-start">
             Walk. Ride. Pedal. Discover.
         </div>
-        <div class="logo"><a href="#"></a></div>
+        <div class="logo"><a href="{% url 'home' %}"></a></div>
         <div class="slogan-end">
             Next time bring your bike onboard.
         </div>


KlaasH:
I have a memory of a redesign presentation at some point in which the idea was mentioned of having animation so that the home view and the directions view appeared not so much as two separate pages but as a form and destinations layer that starts out above the map then moves to the side and becomes more compact when you start interacting with things.  @flibbertigibbet and @lederer, is that a real memory?  Is that behavior a current goal?

Now that I look at this, I think anything but this super-simple version depends on issue #556 (reimplement tabControl).  There's a bunch of old tabControl stuff lying around (e.g. [Home.setTab](https://github.com/azavea/cac-tripplanner/blob/redesign/src/app/scripts/cac/pages/cac-pages-home.js#L158)) and bits of new code doing that stuff but in an ad hoc way (e.g. [inside planTrip](https://github.com/azavea/cac-tripplanner/blob/redesign/src/app/scripts/cac/control/cac-control-directions.js#L218)).

There's no harm in having that logo do something rather than nothing, and it's possible that this will be good enough permanently, so you might as well merge this, but there might be more needed to fully resolve issue #586.

To your points,
- _Resetting the homepage flag in map control_: I hope that and tasks like it will get consolidated into a new tabControl setup that makes it fairly smooth and simple.
- _Destroying the map elements manually_: If the idea is that the home page panels are supposed to look like they're floating over the map, I was figuring most of the stuff on the map could stay there even if the home view is reactivated.
- _Removing any added event handlers_: Yup, could be a thing.
- _Triggering a change to the urlRouter_: This shouldn't be bad, though it's true that it's another thing to keep track of.

lederer:
@KlaasH Your memory is intact. That was the idea. Technically speaking the prototype did indeed "transition" between those two states of a singular page, but instantaneously so, since I had declared few if any CSS `transition` properties. I figured I'd add those in later.

I think the fluid transition should be considered "quite nice to have". It shouldn't be blocking, as long as the current (lack of) transition isn't weird or off-putting, and it should probably be skipped if it will end up taking significant effort to implement. (FWIW I've never worked on a complex transition that didn't take a lot of effort to get right. But it sure would be nice.)

CloudNiner:
Closing, superseded by #626

PR #625, opened 2016-11-15
Feature/add back layer switch minimizer

### LINKED ISSUE #612: Fix layer switcher minimize button
Since FontAwesome icons have been removed, there are no icons for minimize/expand.

flibbertigibbet:
Re-implements the layer switcher minimize button and shuffles around some selectors.

Fixes #612.

@lederer: this could use some styling help. Since we no longer have a minus icon, I just went with the layer icon for both minimize and maximize. Tried just using a textual minus sign in a span, but it was too small to be easily seen. Also it doesn't pull right any more in the maximized layer switcher.

image

### DIFF
--- a/src/app/scripts/cac/map/cac-map-control.js
+++ b/src/app/scripts/cac/map/cac-map-control.js
@@ -3,7 +3,13 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
 
     var defaults = {
         id: 'map',
-        selector: '#map',
+        selectors: {
+            id: '#map',
+            leafletMinimizer: '.leaflet-minimize',
+            leafletLayerList: '.leaflet-control-layers-list',
+            leafletLayerControl: '.leaflet-control-layers',
+            layerListMinimizedClass: 'minimized'
+        },
         center: [39.95, -75.1667],
         zoom: 14,
     };
@@ -31,6 +37,7 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
     var layerControl = null;
     var tabControl = null;
     var zoomControl = null;
+
 
     var loaded = false; // whether map tiles loaded yet (delay on mobile until in view)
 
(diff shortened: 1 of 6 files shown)


src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -3,7 +3,13 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
 
     var defaults = {
         id: 'map',
-        selector: '#map',
+        selectors: {
+            id: '#map',
+            leafletMinimizer: '.leaflet-minimize',
+            leafletLayerList: '.leaflet-control-layers-list',
+            leafletLayerControl: '.leaflet-control-layers',
+            layerListMinimizedClass: 'minimized'
```
KlaasH: Since there was no layer picker in the prototype, these classes don't exist in the redesigned styles.  Grabbing them from [the old version](https://github.com/azavea/cac-tripplanner/blob/develop/src/app/styles/templates/_app.scss#L25) makes this look a lot better.

layer_picker layer_picker_minimized

Specifically that's with
```
    .leaflet-control-layers {
        margin-right: 30px;
        margin-bottom: 30px;
        position: relative;
        padding: 15px 20px 18px;
        border-radius: 0;
        box-shadow: 0 0 12px rgba(0,0,0,0.35);

        .leaflet-control-layers-separator {
            margin: 15px -20px 15px -20px;
        }
        .leaflet-control-layers-selector {
            margin-right: 8px;
            top: 0;
        }
    }

    .leaflet-minimize {
        color: #707070;
        cursor: pointer;
        position: absolute;
        border-radius: 3px;
        line-height: 10px;
        padding: 13px 15px;
        bottom: 0;
        z-index: 99999;
        right: 0;

        &.minimized {
            height: 38px;
            line-height: 0;
            background: white;
        }
    }
```
added to `src/app/styles/components/_map.scss`, which is almost the same as the old styles but with the font override removed and the height adjusted in `.minimized`.

lederer:
Let's use the layers button for both states. No minus needed.

@KlaasH's suggestion SGTM. Once this is merged, just assign an issue to me to tweak the styles and I'll polish it up.

flibbertigibbet:
@KlaasH for future reference, we will need styling help for any modifications to our font selection; we can't just add an icon with Fontello, as the whole font set must be regenerated.

lederer:
Yeah. It's not too complicated to do (if the desired icon is already part of the fontello library) once you get the hang of it. I'm happy to do it anytime.

PR #628, opened 2016-11-17
Share modal redesign

KlaasH:
Implements the share modal, which replaces the share links popup.  The logic was partly in CAC.Share.Social and partly in CAC.Control.DirectionsList, now it's almost all in CAC.Share.Modal.

Makes some changes to CAC.Control.Modal to make it create independent instances, and also adds onOpen and onClose callback options.

Most of the items pop up the same 3rd-party windows they created before, but the direct link option shows in the modal itself.

share_modal

get_link

### DIFF
--- a/python/cac_tripplanner/templates/home.html
+++ b/python/cac_tripplanner/templates/home.html
@@ -87,39 +87,7 @@
 
     <div class="directions-step-by-step"></div>
 
-
-    <div class="modal-overlay">
-        <div class="modal-panel modal-options">
-            <header class="modal-header">
-                Bike options
-            </header>
-            <ul class="modal-list modal-contents">
-                <li class="modal-list-indego">Indego bike sharing</li>
-                <li class="modal-list-timing">Arrive/depart at…</li>
-                <li class="modal-list-ride">Ride quality</li>
-            </ul>
-            <footer class="modal-footer">
-                <button name="close-modal" class="btn-close-modal">Close</button>
-            </footer>
-        </div>
-
-        <div class="modal-panel modal-share">
-            <header class="modal-header">
-                <i class="icon-share"></i>
-                Share
-            </header>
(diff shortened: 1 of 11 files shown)


python/cac_tripplanner/templates/home.html
```diff
@@ -87,39 +87,7 @@
...
-                <li class="modal-list-facebook">Facebook</li>
-                <li class="modal-list-google">Google+</li>
-                <li class="modal-list-email">Email</li>
-            </ul>
-            <footer class="modal-footer">
-                <button name="close-modal" class="btn-close-modal">Close</button>
-            </footer>
-        </div>
-    </div>
+    {% include "partials/modals.html" %}
```
CloudNiner: 👍

KlaasH:
build again

KlaasH:
Looking at this just now, I remembered there's a design issue I let slide: adding `<span>` elements inside `.modal-header` caused the flex layout options to stop being applied, meaning it started looking like the above (icon next to text) rather than like the prototype.

prototype_link_modal

@lederer, any tips?  I tried a couple things to fix it but they didn't work.  I could make copies of the whole `<header>` tag instead of making `<span>`s inside if that's necessary, but less duplication between the variations would be preferable.

KlaasH:
Thanks, @lederer, for the flex-styling guidance!  1d45147aa5 adds a `modal-header-contents` that allows multiple switchable versions within the `<header>` (with `default` shown initially and anything with `secondary` hidden initially), but leaves the styles on `modal-header` as well so that modals with no such complications can just put their stuff in `<header>`.  Possibly maintaining both options will end up being problematic, and it's certainly not worth going to any trouble for it, but right now it seems to be working fine.

In other news, I've run all the tests on my instance and I'm pretty sure Jenkins will be fine with this once he gets over his grumpy mood, so I'm going to merge.

PR #640, opened 2016-11-18
Feature/fix bike mode icon

flibbertigibbet:
Fix bike icon not displaying within step-by-step directions list and temporarily disables bike triangle options (to get bike mode working again).

### DIFF
--- a/src/app/scripts/cac/control/cac-control-directions.js
+++ b/src/app/scripts/cac/control/cac-control-directions.js
@@ -173,11 +173,13 @@ CAC.Control.Directions = (function (_, $, Control, Geocoder, Routing, Typeahead,
         }
 
         if (mode.indexOf('BICYCLE') > -1) {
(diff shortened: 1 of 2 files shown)


KlaasH:
:+1: Cool.

I noticed one situation where it's not showing up.  The first itinerary for http://localhost:8024/?origin=39.965811%2C-75.183494&originText=Fairmount%20Waterworks%20Interpretive%20Center&mode=BICYCLE%2CTRANSIT&destination=39.9616133%2C-75.1541914&destinationText=990%20Spring%20Garden%20St%2C%20Philadelphia%2C%20Pennsylvania%2C%2019123 never shows a bike icon.  After "arrive path" there should be a step saying to set out by bike, but it goes right to a turn step.

Probably a separate issue, unless you have a good idea of why that would be happening.

no_bike_icon

flibbertigibbet:
Yeah, that sounds like a separate issue; worth creating an issue for that.

PR #646, opened 2016-11-21
Feature/fix typeahead opening

### LINKED ISSUE #601: Typeahead dropdown sometimes opens on marker drag
Sometimes when the destination address updates after destination marker drag, typeahead selection drop-down opens on the destination form field.

flibbertigibbet:
Fixes #601.

Also tidies some typeahead and origin/destination marker related things for the redesign.

Issue can be reproduced (either on current production/`develop` branch or `redesign` branch) by clicking either origin or destination text field on map view to give it focus, then dragging the related origin/destination marker on the map; typeahead autocomplete list will open on the text field.

This hasn't been noticeable on production as the origin/destination fields do not have focus by default on map page load, but with the redesign, whichever got selected last to trigger map page load will still have the focus.


flibbertigibbet:
Just fixed up the test/production build issues introduced in #628.

Note that if modules get renamed, moved, or if their interdependencies change, the file names/orders in the gulp file and/or the karma configurations may need updating, in addition to the development script loading in the `base.html` template.

flibbertigibbet:
build again

flibbertigibbet:
I'm guessing you ran just the development tests then. The JS test run command used by Jenkins is `npm run gulp-test`, which runs the production suite, development suite, and coverage check. There's also a linter at `npm run lint`.

PR #642, opened 2016-11-18
Configure OTP for bike share

flibbertigibbet:
If bcycle_api_key ansible parameter is defined and non-empty in the group vars,
add router configuration to OpenTripPlanner server to support bike share routing
via the BICYCLE_RENT mode.

Closes #627.

To test, set the `bcycle_api_key` in your development group_vars. See #627 for how to query; something like [this](http://192.168.8.26/otp/routers/default/plan?fromPlace=39.9611421%2C-75.1546293&toPlace=39.9535847%2C-75.1735277&time=02%3A45pm&date=11-18-2016&fromText=N%2010th%20St%20%26%20Nectarine%20St%2C%20Philadelphia%2C%20Pennsylvania%2C%2019123&toText=2020%20Market%20St%2C%20Philadelphia%2C%20Pennsylvania%2C%2019103%2F&mode=BICYCLE_RENT&arriveBy=false&maxWalkDistance=482802) should work.

Issue #630, opened 2016-11-17
Wire up bike share option in UI


flibbertigibbet:
Wire up UI components for selecting and routing via Indego bike share.

Depends on #627 and #629.

lederer:
This should be in `src/prototype/extras.html` (demoable at http://localhost:3000/extras.html if you run `grunt` in the `prototype` directory; scroll to the bottom of the page). 

Holler if that's insufficient or if I misunderstand the issue.

flibbertigibbet:
Closed in #662

PR #651, opened 2016-11-23
Add bike share arrivals, use mode icon for first step

### LINKED ISSUE #631: Bike share specific directions
Modify itinerary summary and step-by-step directions results for bike share to use Indego icon and include information on picking up and dropping off bike.
Depends on #630.

### LINKED ISSUE #645: Missing bike icon in steps list when transitioning from walking
The first itinerary [biking from Fairmount Waterworks](http://localhost:8024/?origin=39.965811%2C-75.183494&originText=Fairmount%20Waterworks%20Interpretive%20Center&mode=BICYCLE%2CTRANSIT&destination=39.9616133%2C-75.1541914&destinationText=990%20Spring%20Garden%20St%2C%20Philadelphia%2C%20Pennsylvania%2C%2019123) never shows a bike icon.  After "arrive path" there should be a step saying to set out by bike, but it goes right to a turn step.

no_bike_icon

KlaasH:
Adds the bike share icon and the text "Indego station" to the steps list
at the end of any leg that ends at a bike share station. Currently
doesn't distinguish between picking up a bike and dropping one off.

Resolves #631.

Also resolves #645 by changing the first step of each non-transit leg to
use a mode icon and the absolute direction (e.g. 'Bike west on ...')
rather than a turn icon and the relative direction.

**To Test**
To enable bike share directions before issue #630 is done, assuming your OTP instance was provisioned with bike share (PR #642), change `'BICYCLE'` to `'BICYCLE_RENT'` in the `modes` object in [`cac-control-mode-options.js`](https://github.com/azavea/cac-tripplanner/blob/feature/bike-share-steps/src/app/scripts/cac/control/cac-control-mode-options.js#L32).

What that done, get bike directions, e.g. http://localhost:8024/?origin=39.9616133%2C-75.1541914&originText=990%20Spring%20Garden%20St%2C%20Philadelphia%2C%20Pennsylvania%2C%2019123&mode=BICYCLE_RENT&destination=39.965811%2C-75.183494&destinationText=Fairmount%20Waterworks%20Interpretive%20Center, and you should see Indego icons and text at the point where you pick up and drop off the bike and also mode icons at the beginning of each leg.

share_steps1
share_steps2
share_steps3

### DIFF
--- a/src/app/scripts/cac/map/cac-map-templates.js
+++ b/src/app/scripts/cac/map/cac-map-templates.js
@@ -235,6 +235,15 @@ CAC.Map.Templates = (function (Handlebars, moment, Utils) {
     }
 
     function itinerary(templateData) {
+        // set a flag on the last leg, so we can avoid diplaying arriving there right above
+        // also arriving at the final destination
+        templateData.legs[templateData.legs.length - 1].lastLeg = true;
+        // And set a flag on legs that end at bike share stations (whether on a bike or walking
+        // to a station), so we can show the icon
+        _.forEach(templateData.legs, function (leg) {
+            leg.toBikeShareStation = leg.to.vertexType === 'BIKESHARE';
+        });
+
         // The &nbsp;'s are used instead of 'hide' classes because of some styling-related issues
         var source = [
             '<header class="step-by-step-header">',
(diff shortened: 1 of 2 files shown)


flibbertigibbet:
Not in scope for this task, but I just noticed that we probably should change the walk instruction text from 'walk' to 'walk the bike' when telling the user they should dismount for a section, but expect them to still have the bike. From above:

image

src/app/scripts/cac/map/cac-map-templates.js
```diff
@@ -258,36 +267,41 @@ CAC.Map.Templates = (function (Handlebars, moment, Utils) {
...
-                        '<div class="directions-instruction">Disembark <strong>',
-                            '{{this.to.name}}</strong></div>',
-                    '</div>',
-                    '{{else}}',
-                    // non-tranist step directions
-                    '{{#each steps}}',
-                        '<div class="directions-step {{directionClass this.relativeDirection ../this.mode}}" ',
-                            'data-lat="{{ lat }}" data-lon="{{ lon }}">',
-                            '<div class="directions-instruction">{{directionText}}</div>',
+                        // transit step directions
```
flibbertigibbet: It's unclear to me why the transit step directions here have been moved outside of the `step` block of the Handlebars template. References to `this` here will now refer to the `leg` instead of the `step`. It seems that it's fine though in practice, and [here](https://github.com/opentripplanner/OpenTripPlanner/blob/f60440f12f9465390bc39dea7ee8e120d382eb3f/src/main/java/org/opentripplanner/api/model/Leg.java#L222) it looks like OTP will only return a single step for transit.

My concern is that the info may be different for the leg, like departure/arrival times not matching... seems fine though.

src/app/scripts/cac/map/cac-map-templates.js
```diff
@@ -258,36 +267,41 @@ CAC.Map.Templates = (function (Handlebars, moment, Utils) {
             '</div>',
             '{{#each data.legs}}',
                 '<div class="directions-leg" ',
-                    'data-lat="{{this.from.lat}}" data-lon="{{this.from.lon}}">',
-                    // transit step directions
+                'data-lat="{{this.from.lat}}" data-lon="{{this.from.lon}}">',
```
KlaasH: Yup

src/app/scripts/cac/map/cac-map-templates.js
KlaasH: Actually OTP doesn't return any steps for transit ("used for walking, biking and driving" in that comment isn't entirely explicit about what happens for other modes, but seemingly there's an implied "only").  I only changed the indentation of the transit block (which I should have done in a separate commit...) but yeah, this setup seems reasonable to me.

KlaasH:
Adding "walk the bike" was simple, so I did it.

walk_the_bike

flibbertigibbet:
Nice. Will still need a separate issue to deal with walking the bike when it isn't rented, but this is a nice simple solution for the rented bike case.

PR #659, opened 2016-11-30
UI Enhancements

lederer:
Various UI tweaks, including:

- Replace (red logo) favicon and apple-touch-icons with blue logo
- Style error state of directions input fields.
- Style itinerary alert.
- Style typeahead

Some notes on typeahead:

### Tabindex
Not a typeahead issue per se, but while I was in there I added tabindexes to the origin, destination, geocode, and reverse buttons.

### Reverse button visibility
I removed the logic that toggles the visibility of the "reverse route" button in map view. There were cases where it would show or hide when it shouldn't. In attempting to improve it, I determined there are two states when it should be hidden:

1. Both origin and destination fields are empty. Swapping two empty fields is a noop.
2. The origin typeahead menu is visible, overlapping the reverse button. Cosmetically it's… not great.

The value of fixing 1 is negligible, though, and not commensurate with the effort required to get the logic right.

And 2 is blocked by a [typeahead bug](https://github.com/twitter/typeahead.js/issues/1141). It's fixed in the latest version of typeahead, but it's a complete rewrite. I don't think it can just be dropped in.

### Hinting and Autoselecting
I've disabled typeahead `hint` and `autoselect`. I'm not 100% sure this is right, but want to sit with it for a bit. Rationale:

- When autoselect is enabled, pressing enter selects the top suggestion, even if you haven't cursored to it. But geocoder ranking is poor enough that it's unlikely the top suggestion is the one you want. The path of least resistance shouldn't lead you away from your intent, so… thanks but no thanks, autoselect.

- We disable hinting because it's confusing and misleading to have hinting when autoselect is disabled. Further, if your input is all lowercase, it looks strange to have the rest of the string (ie, the hint) displayed in sentence case (address case?).

### TODO
- We may need additional UX follow up to finesse the case where the user blurs but has not selected a typeahead suggestion, leaving the input with an empty or incomplete location. This may be as simple as adding the `.error` class.

- If the the destination field is empty when the user selects an origin suggestion, we should automatically focus it.


flibbertigibbet:
The behavior of hitting enter selecting the first option listed was something the client had asked us for specifically, as I recall.

flibbertigibbet:
Typeahead is no longer supported. I would like to ditch it altogether, but it's there and it works, more or less. After the rewrite, it was abandoned. There are a couple of forks that are supported, but they both only support the rewrite version, which we are not using.

flibbertigibbet:
Regarding the error class TODO: that should be the existing behavior. Input validation gets repeated on trip plan.

Issue #660, opened 2016-12-01
Sync controls with URL on page load


flibbertigibbet:
When loading a page with directions options in the URL, user controls for mode and trip options modals should sync with what got set in user preferences from the URL before display.

Also, the router should no longer distinguish between 'read' versus 'encode' settings (no longer using local storage, so nothing from the URL should be overridden).

flibbertigibbet:
The mode options controls will need some rework to handle `BICYCLE_RENT` as well as `BICYCLE` as a mode; default to just `BICYCLE`.

flibbertigibbet:
Closed in #693

PR #662, opened 2016-12-01
Feature/trip options

### LINKED ISSUE #589: Update stored user preferences

### LINKED ISSUE #629: Trip options modal
Wire up a trip options modal for when overflow icon clicked, including options such as departure/arrival time, bike optimization options, and the like. Should fix #608.

flibbertigibbet:
Adds and wires up the trip options modals. Also implements bike share option.

Closes #629 and #630.

Created #663 for displaying user option selection in top-level modal if non-default.

Removes local storage altogether and implements differentiation between default options and user-set options (per discussion with Scott), so closes #589. Created #660 for issues around page reload.

Also per Scott: when navigating back to home page, user options should be reset; created #660 for this.


flibbertigibbet:
For some reason, the 'safer' option isn't clickable in the bike triangle options.

python/cac_tripplanner/templates/partials/modals.html
```diff
@@ -13,6 +13,79 @@
...
+    <div class="modal-panel modal-options modal-sub-menu bike-triangle">
+        <header class="modal-header">
+            <i class="icon-road"></i>
+            What kind of ride?
+        </header>
+        <ul class="modal-list modal-contents modal-list-narrow">
+            <li id="bikeTriangleAny" class="modal-list-choice selected">Any</li>
+            <li id="bikeTriangleFast" class="modal-list-choice">Fast</li>
+            <li id="bikeTriangleFlat" class="modal-list-choice">Flat</li>
+            <li id="bikeTriangleSafe" lass="modal-list-choice">Safe</li>
```
KlaasH: I believe we can hold this `lass` responsible for issue #664.

flibbertigibbet: D'oh! Good catch.

src/app/scripts/cac/control/cac-control-directions.js
```diff
@@ -178,35 +149,30 @@ CAC.Control.Directions = (function (_, $, Control, Geocoder, Routing, Typeahead,
...
+                console.error('unrecognized bike triangle option ' + bikeTriangle);
             }
 
-            // true if box checked
-            var wheelchair = $('input', options.selectors.wheelchairDiv).prop('checked');
-            UserPreferences.setPreference('wheelchair', wheelchair);
-            $.extend(otpOptions, { wheelchair: wheelchair });
+            // check user preference for bike share here, and update query mode if so
+            if (UserPreferences.getPreference('bikeShare')) {
+                mode = mode.replace('BICYCLE', 'BICYCLE_RENT');
```
KlaasH: Once we get these preferences loading from the URL, is there a possibility that the mode will already include `BICYCLE_RENT` at this point?  If so this would turn it into `BICYCLE_RENT_RENT`.
```
mode = mode.split(',').map(function (m) {
    return m === 'BICYCLE' ? 'BICYCLE_RENT' : m;
}).join(',');
```
is a little wordier, but it's idempotent.

src/app/scripts/cac/user/cac-user-preferences.js
```diff
@@ -1,67 +1,75 @@
...
-    // TODO: figure out local storage strategy
-    // set up local storage
-    // var namespaceStorage = Storages.initNamespaceStorage('cac_otp');
-    // var storage = namespaceStorage.localStorage;
 
     // Initialize preference storage object.
-    // Currently it just uses an 'options' dictionary, so preferences lives only as long as
-    // the page for which this is initialized.
+    // All values in storages should be stringified first.
+    // Setting to a falsy value will remove the object from storages;
```
KlaasH: Unintended falsiness is such a troublesome source of bugs that I'm trying to think of whether this could be a little tighter without becoming annoying.  It doesn't seem impossible that there could come a time when a zero should be treated as a zero.  I guess treating empty string as "unset" makes sense since it could be set that way by a URL param.  What about `Setting to null, undefined, or empty string will remove the object from storages`? Then the condition below would be:
```
if (!_.isUndefined(val) && !_.isNull(val) && val !== 'undefined' && val !== '') {
```
It's wordier, but not terrible.
Or flip it and have the `delete` be the conditional case, with
```
if (_.isUndefined(val) || _.isNull(val) || val === 'undefined' || val === '') {
```
That's slightly easier to read and would also short-circuit, for what that's worth.

src/app/scripts/cac/user/cac-user-preferences.js
```diff
@@ -1,67 +1,75 @@
...
      *
      * @param {String} preference Name of setting to fetch
-     * @param {Boolean} [setDefault=true] If false, don't set the default value if no value is set
      * @return {Object} setting found in storage, or default if none found
      */
-    function getPreference(preference, setDefault) {
+    function getPreference(preference) {
         var val = storage.get(preference);
-        if (val) {
+        if (!val || val === '') {
```
KlaasH: Maybe move the logic defining "set-ness" up to `storage.get` so that it's all in one place?  Then `val` here would be guaranteed to be either meaningful or `undefined`.

src/app/scripts/cac/user/cac-user-preferences.js
```diff
@@ -76,6 +84,17 @@ CAC.User.Preferences = (function(Storages, _) {
...
 
     /**
+     * Check if value has been set by user, or is a default.
+     * Will return false if given preference does not exist.
+     *
+     * @param {String} preference Name of setting to check
+     * @return {Boolean} True if getPreference will return a default value
+     */
+    function isDefault(preference) {
+        return !_.has(options, preference) && _.has(defaults, preference);
```
KlaasH: Nice and tidy.

It's making me wonder, though, about what the right behavior is when someone sets an option to the default value.  This will be false.  Is that what we want?  If not we could address that either here or in `storage.set()`, by checking incoming values against the default and deleting them if they match.

But maybe it is intended.  If the idea is for people to see their selections in the modal, it might make sense to show them even if they pick default values.

flibbertigibbet: Yes, that is what we want. We need to distinguish between whether it is the default value because it is returning a default, or if it happens to be the default because the user explicitly set it to that value. In the latter case, the options dialog needs to show the user selection and not the general menu text.

This user preferences refactor is to distinguish between the two.

src/app/scripts/cac/user/cac-user-preferences.js
flibbertigibbet: All values passed in to this are stringified first (see comment; happens in public preference setter), so an empty string and undefined are the only possible falsy input, and an empty string always means "not set."

Zero can be stored fine because we're actually storing `'0'`.

We should keep the stringifying, if for no other reason, as it will let us switch back to using browser storage if we want.

src/app/scripts/cac/control/cac-control-directions.js
flibbertigibbet: Right, this needs a refactor to deal with URL sync. The mode button togging updates will also need modification for that.

flibbertigibbet:
Tests pass locally; that was a PhantomJS crash. Merging.

PR #666, opened 2016-12-02
Configure OTP for bike share

flibbertigibbet:
Cherry-pick back server configuration for bike share, so production and staging can share an OTP stack and staging can use bike share routing.


flibbertigibbet:
Cherry-pick with no conflicts, merging.

Issue #657, opened 2016-11-29
Add back loading spinners


flibbertigibbet:
Hide directions sidebar content on trip plan start to show a loading indicator until directions are ready to display.

Also should probably add back spinners anywhere else AJAX requests happen before display refresh.

flibbertigibbet:
@lederer are our old loading animations okay to use? We've got spinkit in the project.

lederer:
@flibbertigibbet Can you outline steps to repro a spinner in the current production site? I can't seem to trigger one, despite network throttling.

flibbertigibbet:
Apparently broken on current production (related CSS not loaded?), although there is still a div with the `sk-spinner` class you can see show/hide for loading 'view all' featured locations on home page or in sidebar during directions fetch.

flibbertigibbet:
Usage changed upstream. The classes should now be:
```
<div class="sk-three-bounce">
    <div class="sk-child sk-bounce1"></div>
    <div class="sk-child sk-bounce2"></div>
    <div class="sk-child sk-bounce3"></div>
</div>
```

lederer:
Yup that spinner is fine.

PR #668, opened 2016-12-02
Fix loading spinners

flibbertigibbet:
Usage of spinkit changed in package upgrade. See #657.


flibbertigibbet:
Fixes loading animation on production. Trivial; merging.

PR #671, opened 2016-12-05
Display user selections for trip options

### LINKED ISSUE #663: Display user-set trip options
In top-level trip options control, display user set options instead of default menu text if user has selected a preference for that option (user preference value is not returning a default value).

flibbertigibbet:
Closes #663.

Display user selections in top-level trip options modal if set; if using defaults,
use default menu option text.

After making a selection from a sub-menu, on returning to the main options menu or closing/re-opening the trip options will display the selection instead of the default menu text.

### DIFF
--- a/src/app/scripts/cac/control/cac-control-trip-options.js
+++ b/src/app/scripts/cac/control/cac-control-trip-options.js
@@ -6,6 +6,14 @@ CAC.Control.TripOptions = (function ($, Handlebars, moment, Modal, UserPreferenc
 
     var defaults = {
         currentMode: 'WALK',
+        // default text to display for top-level modal menu options
+        // when user has not set anything for it yet
+        defaultMenuText: {
+            bikeShare: 'Indego bike sharing',
+            timing: 'Arrive/depart at…',
+            bikeTriangle: 'Ride quality',
+            accessibility: 'Accessibility'
+        },
         selectors: {
             bodyModalClass: 'body-modal body-modal-options',
             selectedClass: 'selected', // used to mark selected list item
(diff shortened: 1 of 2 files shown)


src/app/scripts/cac/control/cac-control-trip-options.js
```diff
@@ -270,6 +285,108 @@ CAC.Control.TripOptions = (function ($, Handlebars, moment, Modal, UserPreferenc
...
+     */
+    function getTimingText() {
+        var timing = options.defaultMenuText.timing;
+        if (!UserPreferences.isDefault('dateTime')) {
+            var dateTime = moment.unix(UserPreferences.getPreference('dateTime'));
+            var arriveBy = UserPreferences.getPreference('arriveBy');
+            timing = arriveBy ? 'Arrive ' : 'Depart ';
+            // format as relative time
+            // TODO: something more exact
+            timing += dateTime.fromNow();
```
KlaasH: The prototype shows this as an absolute time ("Depart 12/14 11:00a"), i.e. `dateTime.format('M/YY h:mma')`

flibbertigibbet: Prototype shows a combination of relative date with exact time in at least one place ("Today at 3:30pm" or some such), which cannot be accomplished directly with moment.js. @lederer has suggested maybe using https://github.com/EvanHahn/HumanizeDuration.js for date/time formatting elsewhere.

lederer: I've got this on my list to follow up on.

PR #679, opened 2016-12-05
Add back directions loading spinner

### LINKED ISSUE #657: Add back loading spinners

flibbertigibbet:
Closes #657.

Display spinner while directions are loading and hide itinerary list.
Also remove old home template (replaced).


flibbertigibbet:
The old design also had a loading spinner for lazy-loading all locations/articles, but since we pass them all to the template now, there's no more ajax calls on the home page.

PR #673, opened 2016-12-05
Hide unpublished destinations

flibbertigibbet:
Do not show unpublished destinations on home page.
Closes #585 (again).

PR #681, opened 2016-12-06
Order destinations by priority

### LINKED ISSUE #592: Sort featured destinations by priority column

flibbertigibbet:
Add priority value to destinations; defaults to a high number (9999).
By default, order published destinations by priority.
Closes #592.

Involves a migration, so will need to run that to test.

### DIFF
--- a/python/cac_tripplanner/destinations/models.py
+++ b/python/cac_tripplanner/destinations/models.py
@@ -15,14 +15,14 @@ class DestinationManager(models.GeoManager):
     """Custom manager for Destinations allows filtering on published"""
 
     def published(self):
-        return self.get_queryset().filter(published=True).order_by('priority')
-
-    def get_queryset(self):
-        return super(DestinationManager, self).get_queryset()
+        return self.get_queryset().filter(published=True)
 
 
 class Destination(models.Model):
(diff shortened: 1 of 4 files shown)


flibbertigibbet:
Supersedes #673.

flibbertigibbet:
build again

python/cac_tripplanner/destinations/models.py
```diff
@@ -15,7 +15,7 @@ class DestinationManager(models.GeoManager):
     """Custom manager for Destinations allows filtering on published"""
 
     def published(self):
-        return self.get_queryset().filter(published=True)
+        return self.get_queryset().filter(published=True).order_by('priority')
```
KlaasH: It may be a distinction without a difference, but I'd be inclined to set this as the default ordering for the model, i.e.
```
class Meta:
    ordering = ['priority']
```
If we want to keep randomizing destinations with tied priorities, it could be `ordering = ['priority', '?']`

Also, not directly related but since we're right here, you could delete `get_queryset` since it's just calling super with no modification.

flibbertigibbet: Sure

PR #682, opened 2016-12-06
Humanize distances in yards/miles

### LINKED ISSUE #643: Display short distances in yards
Currently all distances are displayed in miles.

Display short distances (less than a quarter mile?) in yards.

flibbertigibbet:
Modify distance display for itinerary summary and step list to show distances less than
half a mile in yards. Closes #643.

### DIFF
--- a/src/app/scripts/cac/map/cac-map-templates.js
+++ b/src/app/scripts/cac/map/cac-map-templates.js
@@ -162,7 +162,7 @@ CAC.Map.Templates = (function (Handlebars, moment, Utils) {
             '<div class="block-detail">',
                 '<div class="trip-numbers">{{#if d.formattedDuration}}<div class="trip-duration"> ',
                 '{{ d.formattedDuration }}</div>{{/if}}<div class="trip-distance">',
-                '{{ d.distanceMiles }}</div></div>',
+                '{{ d.formattedDistance }}</div></div>',
                 '<h3>{{ d.name }}</h3>',
                 '<img class="explore-block" src="{{#if d.wide_image}}{{ d.wide_image }}',
                     '{{^}}https://placehold.it/300x150{{/if}}" />',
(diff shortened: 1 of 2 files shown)


src/app/scripts/cac/map/cac-map-templates.js
```diff
@@ -333,8 +333,14 @@ CAC.Map.Templates = (function (Handlebars, moment, Utils) {
             return new Handlebars.SafeString(dt.format('h:mma'));
         });
 
-        Handlebars.registerHelper('inMiles', function(meters) {
-            return new Handlebars.SafeString(Math.round(((meters / 1000) * 0.621371) * 100) / 100);
+        Handlebars.registerHelper('inMilesYards', function(meters) {
+            if (meters < 797) {
+                var yards = Math.round(meters * 1.09361);
+                return new Handlebars.SafeString(yards.toString() + ' yards');
```
KlaasH: Was there a discussion of what small unit to use and where to set the threshold for using it?  I remember a conversation in RRP which I thought was about feet and the threshold floating around was lower (.2 mi? or maybe it was just that we observed that Google uses .1), but I don't remember it reaching a definitive conclusion.  But for me, the smaller unit in which I think is feet, and I switch to fractional miles lower, probably around a quarter mile.

So I would vote for
- renaming this helper to `formatDistance`
- using feet, with a lower switch-over threshold
- setting this up using a config object, e.g.
```
smallDistanceUnit = {
    thresholdMiles: 0.25,
    unitName: 'feet',
    conversionFromMeters: 3.28084
};
```
(The threshold could be in meters, but it seems nicer to have the code convert it, i.e. `if (meters < smallDistanceUnit.thresholdMiles * 1609.34) {`.)
- put that somewhere where it can be shared between `cac-map-templates.js` and `cac-routing-itinerary.js`. I'm not sure where that would be.  Two identical copies of the config object would still be better than two similar-but-not-identical functions, but one copy would be best.

flibbertigibbet:
Reworked. Changed to use feet for distances under 0.2 miles, and ditched the Handlebars template formatter to use a consistent format for itineraries/legs/steps that is set on each as part of the itinerary parse, which should be more efficient and easier to maintain.

src/app/scripts/cac/routing/cac-routing-itinerary.js
```diff
@@ -105,16 +105,33 @@ CAC.Routing.Itinerary = (function ($, cartodb, L, _, moment, Geocoder, Utils) {
...
+    function getFormattedItineraryDistance(legs) {
         var distanceMeters = _.chain(legs).map('distance').reduce(function(sum, n) {
             return sum + n;
         });
-        return Math.round(((distanceMeters / 1000) * 0.621371) * 100) / 100;
+        return getFormattedDistance(distanceMeters);
+    }
+
+    /**
+     * Helper function to get formatted string in yards or miles for a given distance in meters
```
KlaasH: "yards" -> "feet"

PR #683, opened 2016-12-06
Fix missing event on typeahead blur

### LINKED ISSUE #670: Cannot read property currentTarget of undefined
Sometimes after a marker drag, receive JS error `Cannot read property currentTarget of undefined`. Cannot reproduce reliably.

Possibly related to typeahead input blur to prevent autocomplete from opening on marker drag.

flibbertigibbet:
Fixes #670.

Typeahead blur listener referenced event, but did not have event param; fixes that
and checks for currentTarget before referencing it.


flibbertigibbet:
To test when the `if` test is false, give origin or destination field focus, then drag the related (origin or destination) marker. To test when `if` case is true, edit origin or destination field text without selecting any option, leave focus in the field, then drag the related marker. Neither should throw a console error/result in blank itinerary results.

PR #684, opened 2016-12-06
Handle trip reversal with missing address

### LINKED ISSUE #677: Check for missing field on reverse
Check for one or both fields missing on trip reverse button click.

See #659.

flibbertigibbet:
Let reverse trip button work if origin, destination, or both are missing.
Fixes #677.

### DIFF
--- a/src/app/scripts/cac/control/cac-control-directions.js
+++ b/src/app/scripts/cac/control/cac-control-directions.js
@@ -317,12 +317,6 @@ CAC.Control.Directions = (function (_, $, moment, Control, Geocoder, Routing, Ty
         var destination = UserPreferences.getPreference('destination');
         var destinationText = UserPreferences.getPreference('destinationText');
 
-        if (!(directions.origin && directions.destination)) {
-            setDirectionsError('origin');
-            setDirectionsError('destination');
-            return;
-        }
-
         // update local storage
         UserPreferences.setPreference('origin', destination);
         UserPreferences.setPreference('originText', destinationText);
@@ -333,9 +327,12 @@ CAC.Control.Directions = (function (_, $, moment, Control, Geocoder, Routing, Ty
         typeaheadFrom.setValue(destinationText);
         typeaheadTo.setValue(originText);
 
-        // set on this object and validate
-        setDirections('origin', [destination.location.y, destination.location.x]);
-        setDirections('destination', [origin.location.y, origin.location.x]);
+        if (directions.origin) {
+            setDirections('destination', [origin.feature.geometry.y, origin.feature.geometry.x]);
+        }
+        if (directions.destination) {
+            setDirections('origin', [destination.feature.geometry.y, destination.feature.geometry.x]);
+        }
 
         // update the directions for the reverse trip
         planTrip();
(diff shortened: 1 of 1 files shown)


flibbertigibbet:
Might need some additional cleanup after #668 goes in, which makes the itineraries list show/hide again.

src/app/scripts/cac/control/cac-control-directions.js
```diff
@@ -332,11 +326,18 @@ CAC.Control.Directions = (function (_, $, moment, Control, Geocoder, Routing, Ty
...
+        // have both origin and destination
+        if (directions.origin && directions.destination) {
+            // set on this object and validate
+            setDirections('origin', [destination.feature.geometry.y, destination.feature.geometry.x]);
+            setDirections('destination', [origin.feature.geometry.y, origin.feature.geometry.x]);
+        } else if (directions.origin) {
+            setDirections('destination', [origin.feature.geometry.y, origin.feature.geometry.x]);
+        } else if (directions.destination) {
+            setDirections('origin', [destination.feature.geometry.y, destination.feature.geometry.x]);
+        }
```
KlaasH: These are two independent cases now, right?  I don't see anything special happening in the "both" case.

More importantly, the typeahead `setValue` calls above are triggering events that cause `directions.origin` or `directions.destination` to get cleared.  I think these should be using the local copies of `origin` and `destination` saved at the top.

src/app/scripts/cac/control/cac-control-directions.js
```diff
@@ -332,11 +326,18 @@ CAC.Control.Directions = (function (_, $, moment, Control, Geocoder, Routing, Ty
...
 
-        // update the directions for the reverse trip
+        // update the directions for the reverse trip, or clear itineraries
```
KlaasH: I'm not seeing itineraries get cleared.  Or rather, they're disappearing from the map but not from the sidebar.

flibbertigibbet: Should be fixed by #668; see note above. Code sets 'hidden' class on sidebar, but styling for 'hidden' got removed when we stripped out Bootstrap.

flibbertigibbet: Just pushed rebase on redesign, so itineraries hide now.

src/app/scripts/cac/control/cac-control-directions.js
flibbertigibbet: I originally wrote it to only call to `planTrip` in the 'both' case, but yes, now they're independent.

Pretty sure we do want/need to `setValue`. Clearing out the `error` class for an empty field should be a special case here.

KlaasH: Yeah, the `setValue` is doing necessary work, but it's also clearing the origin or destination from `directions` before we're done with it.  But the conditions can just use the local `origin` and `destination` variables, which is already what the `setDirections` calls are using.

flibbertigibbet:
Going to hold off on further rework on this until #689 is merged

flibbertigibbet:
Rebased and tidied a bit. This should be better behaved now. A field will go red on trip reverse if it was empty or incomplete; not sure if that's a bad thing or not.

KlaasH:
:+1:

PR #687, opened 2016-12-06
Prevent zoom on mobile safari

### LINKED ISSUE #654: Prevent page zooms
As of iOS 10, Mobile Safari ignores `user-scalable=no` in the `viewport` meta tag. This is generally a win for accessibility. However, mobile web apps like this one that employ large maps (or other controls that consume pinches and double-taps) aren't particularly resilient to this change. If the user scales the page via a gesture on a non-map control, they may be unable to scale back if the map is the only remaining touchable control in the viewport.

This code, or some variation on it, should mimic the effects of `user-scalable=no` when set as the `onTouchStart` event handler for the root container element.

```javascript
preventPageZoom(e) {
    if (e.touches.length > 1) {
        e.preventDefault(); // pinch - prevent zoom
        e.stopPropagation();
        return;
    }

    const t2 = e.timeStamp;
    const t1 = this.lastTouch || t2;
    const dt = t2 - t1;
    this.lastTouch = t2;

    if (dt && dt < 500) {
        e.preventDefault(); // double tap - prevent zoom
        e.stopPropagation();
    }
}
```

flibbertigibbet:
Detect if on Mobile Safari, and if so, disable zooming on home page, so user cannot
accidentally zoom in on map and not be able to zoom out.
Closes #654.

### DIFF
--- a/src/app/scripts/cac/pages/cac-pages-home.js
+++ b/src/app/scripts/cac/pages/cac-pages-home.js
@@ -155,9 +155,9 @@ CAC.Pages.Home = (function ($, ModeOptions,  MapControl, TripOptions, TabControl
                     return;
                 }
 
-                const t2 = e.timeStamp;
-                const t1 = this.lastTouch || t2;
-                const dt = t2 - t1;
+                var t2 = e.timeStamp;
+                var t1 = this.lastTouch || t2;
+                var dt = t2 - t1;
                 this.lastTouch = t2;
 
                 if (dt && dt < 500) {
--- a/src/app/scripts/cac/pages/cac-pages-home.js
+++ b/src/app/scripts/cac/pages/cac-pages-home.js
@@ -146,6 +146,27 @@ CAC.Pages.Home = (function ($, ModeOptions,  MapControl, TripOptions, TabControl
             tabControl.setTab(tabControl.TABS.HOME);
         });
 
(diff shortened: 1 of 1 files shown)


flibbertigibbet:
Was able to test the user agent string check with Chrome dev tools, but the dev tools keyboard shortcut for mobile zoom (shift + scroll) apparently doesn't trigger `touchstart`, so will need to test with an actual device.

flibbertigibbet:
build again

lederer:
Might be prudent to remove the check for mobile safari, and just let this apply to all browsers. Rationale:
1. I _think_ it wouldn't have any practical impact for this app in other browsers.
1. We'll be future-proofed against any other browsers following Apple's lead on this.

Thoughts?

flibbertigibbet:
I dunno; I'd prefer to limit this to where we know it's needed in order to reduce the amount of testing we should do for it to one browser/device, as it might result in weird behavior on some device/browser combination(s).

lederer:
That's fair. If it ain't broke yet, don't fix it yet.

lederer:
Will test on staging.

flibbertigibbet:
Merging so we can test after next staging push.

lederer:
Tested on staging. Seems to be working 💯

PR #693, opened 2016-12-09
Feature/sync on load

### LINKED ISSUE #660: Sync controls with URL on page load

### LINKED ISSUE #661: Reset options on navigation to home view
When navigating back to home view from map view, unset all user set trip options and revert to defaults.

Depends on doing #660 first.

flibbertigibbet:
Keep user preferences in sync with URL and the controls, and ensure router can manage reading/setting all the preferences. Clear all preferences, controls, and URL on navigation back to home view.

Closes #660 and closes #661.

* After page refresh, URL options should be reflected in trip options list pickers and top-level modal text
* Clicking logo button on map page to go home should reset all trip options


flibbertigibbet:
Closing in favor of #694.

PR #694, opened 2016-12-09
Feature/trip options banner

flibbertigibbet:
Based on #693. Could close #693 in favor of this PR, or close this and open a new PR once #693 is in.

Implements green trip options banner and "need wheels?" trip options prompt banner. Brings back local storage for the purpose of tracking whether or not the user has ever opened trip options.

Also modifies date/time string formatter to be more precise with the time.

### DIFF
--- a/src/app/scripts/cac/control/cac-control-directions.js
+++ b/src/app/scripts/cac/control/cac-control-directions.js
@@ -45,7 +45,6 @@ CAC.Control.Directions = (function (_, $, moment, Control, Geocoder, Routing, Ty
         destination: null
     };
 
-    var modeOptionsControl = null;
     var mapControl = null;
     var itineraryControl = null;
     var tabControl = null;
@@ -61,9 +60,6 @@ CAC.Control.Directions = (function (_, $, moment, Control, Geocoder, Routing, Ty
         tabControl = options.tabControl;
         itineraryControl = mapControl.itineraryControl;
         urlRouter = options.urlRouter;
-        modeOptionsControl = options.modeOptionsControl;
-        modeOptionsControl.events.on(modeOptionsControl.eventNames.toggle, planTrip);
-        modeOptionsControl.events.on(modeOptionsControl.eventNames.transitChanged, planTrip);
 
         $(options.selectors.reverseButton).click($.proxy(reverseOriginDestination, this));
 
@@ -132,7 +128,7 @@ CAC.Control.Directions = (function (_, $, moment, Control, Geocoder, Routing, Ty
         var date = UserPreferences.getPreference('dateTime');
         date = date ? moment.unix(date) : moment(); // default to now
 
-        var mode = modeOptionsControl.getMode();
+        var mode = UserPreferences.getPreference('mode');
         var arriveBy = UserPreferences.getPreference('arriveBy');
 
         // options to pass to OTP as-is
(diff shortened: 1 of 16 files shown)


flibbertigibbet:
Changes in #697 to date/time relative formatting and modal width to accommodate should override the similar work done in the last two commits (switch to `calendar` for formatting and widen modal to fit).

KlaasH:
Clicking the "x" to dismiss the "need wheels" banner does so, but also opens the options modal.  That's obviously not right, but is clicking the rest of the banner supposed to open the modal?  If so, it should be styled with `cursor: pointer`, but given that the text is telling you to click elsewhere to open the modal, maybe clicking the banner isn't supposed to do anything?

KlaasH:
Toggling between Walk and Bike makes directions refresh, but currently turning Transit on and off doesn't.

flibbertigibbet:
Yes, clicking the banner anywhere other than the 'x' is supposed to open the trip options modal, per @lederer, though I can't find where we had that discussion now.

src/app/scripts/cac/control/cac-control-directions.js
```diff
@@ -239,6 +226,18 @@ CAC.Control.Directions = (function (_, $, moment, Control, Geocoder, Routing, Ty
         directionsListControl.hide();
     }
 
+    /**
+     * Wipe out user-set trip options and clear controls
+     */
+    function clearUserSettings() {
```
KlaasH: I think this should be a function in the Home page controller that triggers on a `tabShown` event, but either setup will cause conflicts with #699, so we can deal with it then.

flibbertigibbet:
Clearing all user settings on navigation home was a decision by @lederer

src/app/scripts/cac/control/cac-control-directions.js
flibbertigibbet: The directions controller owns the typeahead controls cleared here; the home page doesn't have a way to refer to them directly.

flibbertigibbet:
Tests pass locally; merging for staging redeploy

KlaasH:
The transit button now triggers a refresh but doesn't actually affect what mode the directions use.  I think what happened was that the top half of the changes to `cac-control-mode-options.js` got lost when moving from PR #693 to this one.  I.e. https://github.com/azavea/cac-tripplanner/pull/693/files#diff-ba766172ae5d438df1309e2662dfb617R68 and above, which got rid of the separate transit toggle and made the listener on the transit button call `getMode()` (not to mention cleared out the old copy of `bikeTriangle`).

flibbertigibbet:
Good call. Was wondering what happened there. I'll patch it up in another PR.

lederer:
@KlaasH The main rationale for clearing state on returning Home is that there needs to be some easy way to reset the app. AFAICT there's been no clear way of doing that — the only way I've come up with is to manually trim the browser URL to just the domain and press Enter. Obviously that's not reasonable for our users, especially on mobile devices.

Clicking a logo is the de facto way to return Home on the web, and a common way to reset an app, hence this decision.

So what does it mean to reset this app? I think a reset:
- MUST reset the *Options*. Those are invisible on the Home screen and are naturally volatile (especially Timing). Always resetting them is predicable and clean.
- SHOULD reset the *Origin & Destination*. There will be cases when the user wants to reuse their most recent origin and/or destination, and cases when they won't. Without data to know the relative frequencies of these cases, I think we should favor a predictably clean slate. 
  - In fact a better way to make previous locations easily accessible is to remember all of the user's history and integrate it into the suggest list (obviously out of scope right now). 
  - A more subtle but critical observation: our form is designed to work without a submit button. When the fields are pre-filled, it's unclear how to re-submit the form as is – and awkward to actually do so.
- COULD reset the *Travel Modes*. It may be worth remembering these. As you point out, a user may want to default to a particular mode. But I think it's sensible to start with a clean slate and clear everything for now, then reevaluate freely over time. If it becomes clear through usage — whether next week or next year — that we should remember these, then we should.

KlaasH:
@lederer Thanks for laying that out.  I had been playing around with moving between views on my Explore branch which still had 'Directions' as a tab link, which provided a way to return to directions from a pre-filled Home view.  But still a little weird, since changing either field would bounce you back to directions.  So yeah, this all makes sense.

Issue #593, opened 2016-11-03
Add back explore map


flibbertigibbet:
Re-implement map with isochrone

flibbertigibbet:
closed in #699

PR #689, opened 2016-12-06
Use new ESRI geocoder endpoint

flibbertigibbet:
Replace deprecated `find` endpoint with new `findAddressCandidates` endpoint.
New endpoint returns information in a different structure.
New endpoint now supports using both search bounding box and bias location,
so add logic to bias typeahead suggestions to favor by proximity to current map center.

Closes #688.

Issue #557, opened 2016-10-26
Implement places minutes from origin


flibbertigibbet:
On homepage, places list includes text for "N minutes from origin."

Depends on #585.

When the origin field is populated, requery for places to get them in order of travel time by selected mode from origin, then refresh the list of places to reorder them closest first (by travel time) and also display the travel time for each in the places card.

flibbertigibbet:
When the origin field is populated, requery for places to get them in order of travel time by selected mode from origin, then refresh the list of places to reorder them closest first (by travel time) and also display the travel time for each in the places card.

PR #699, opened 2016-12-12
Explore tab

KlaasH:
Implements the redesigned Explore tab.  Get there by clicking Explore in the home view.

explore

Since a lot of what Directions and Explore were doing involved interacting with the form, and doing it from both controllers would be a disaster when both are loaded at once, I broke all of that behavior out into a new controller, `DirectionsFormControl` (notes on naming welcome, since it's also the form for Explore, but I couldn't think of anything un-clunky).  It handles the form and UserPreferences interactions then fires events (`selected`, `cleared`, `reversed`, and `geocodeerror`) that the Directions and Explore controllers can listen for and do the stuff that's specific to them.

Still to do, as separate tasks:
    - Activate the link to Explore at the bottom of the Directions sidebar, and vice versa
    - Limit settings available in the trip options modal to ones that are relevant
    - Trigger updates on changes to settings
    - Make the places list change in response to changes in the origin or settings. I think that's the same feature as issue #557. Or it's a feature for which that is a special case, since the sidebar list should show times corresponding to the selected settings. And probably filter out places that are out of range. I left a bunch of commented-out stuff in the Explore controller related to the old version of this.

This has conflicts from PR #689 and probably from other recent changes as well.  Even though the commits are already too big, I may end up squashing the whole thing together, since my one attempt so far indicated that rebasing the series of commits onto `redesign` is troublesome.

Resolves issue #593..


python/cac_tripplanner/destinations/views.py
```diff
@@ -47,10 +47,10 @@ def home(request):
     }
     if request.GET.get('destination') is not None:
         # If there's a destination in the URL, go right to directions
-        context['tab'] = 'map'
+        context['body_class'] = 'body-map body-map-dir body-has-sidebar-banner'
```
flibbertigibbet: The Python template context is not a good place to put classes, which will change frequently and which we will expect to find set via JavaScript manipulation and not go looking here. I think it might be good to take the class out of the base template altogether.

KlaasH: The tricky thing is that if the page class in the template is 'home', the Home view mostly renders by the time the javascript runs and switches it to the desired view, which looks pretty bad.  And the rendering of "no body class" isn't really defined, so that doesn't look good either.

If every cold load was supposed to start on Home and then animate into the target view, that would be another story.  But even if we add nice animations, I think loading a direct link by animating to it would look weird.

I can poke at it some more and see if I can figure out a better approach.  Thoughts and suggestions welcome :smile:

src/app/scripts/cac/pages/cac-pages-home.js
```diff
@@ -75,13 +77,25 @@ CAC.Pages.Home = (function ($, ModeOptions,  MapControl, TripOptions, SearchPara
...
         directionsControl = new CAC.Control.Directions({
             mapControl: mapControl,
+            directionsFormControl: directionsFormControl,
+            tabControl: tabControl,
+            urlRouter: urlRouter
+        });
+
+        exploreControl = new CAC.Control.Explore({
+            mapControl: mapControl,
+            modeOptionsControl: modeOptionsControl,
```
flibbertigibbet: I think we should be able to leave the `modeOptionsControl` out of the `exploreControl`, as it has been removed from the `directionsControl`, and instead modify the mode toggle listener on this home page module to check the tab and call appropriately into either explore or directions control.

flibbertigibbet:
Looks good! A few observations:

* 'explore' tab on home page goes straight to map page; should there be another view there that goes to the map once an origin entered?
* clicking 'more info' on a place in the sidebar goes to a view that looks to be missing some formatting
* clicking 'learn' tab from 'home' tab now broken as 'explore' tab once was (floating input forms on a map)
* clicking 'home' tab link does nothing
* toggling to bike share mode doesn't update the isochrone; not sure at what levels that might need change

flibbertigibbet:
To note something else (doesn't need to be fixed in this PR):
The buttons on the bottom of the sidebar in the 'explore' view list aren't fully implemented. The 'explore' button should probably switch be a  'directions' button when in the 'explore' view. Right now it just does a full page refresh from both map views.

python/cac_tripplanner/destinations/views.py
KlaasH: 5139467 changes the view back to just picking a tab name and moves the logic for setting the initial body classes into `base.html`.  Still not ideal, but so much of the styling hangs on body class that it's not clear to me how to move it further down in the templates or purely into javascript.

Also note that using both `.body-map` and `.body-map-{tab}` is convenient since there's a lot of styling that's common to both.  `.body-map` without one of the subclasses isn't actually a valid thing to do.  Maybe there's an easy way to switch the styles to not need the shared class, but if not it doesn't seem worth the trouble.

KlaasH:
Still to do on this PR:
- [x] clicking 'learn' tab from 'home' tab now broken as 'explore' tab once was (floating input forms on a map)
- [ ] modeOptionsControl out of the exploreControl

Follow-issues:
- [ ] clicking 'more info' on a place in the sidebar goes to a view that looks to be missing some formatting
- [ ] clicking 'home' tab link does nothing
- [ ] toggling to bike share mode doesn't update the isochrone; not sure at what levels that might need change
- [ ] sidebar buttons to toggle between Explore and Directions

Open questions:
- [ ] 'explore' tab on home page goes straight to map page; should there be another view there that goes to the map once an origin entered?

KlaasH:
After many changes and some additions, I've gotten the "do on this PR" boxes checked in the comment above. I'll open issues for the others.

flibbertigibbet:
Looks great. One observation: from the 'learn' tab, neither 'home' nor 'explore' tab buttons do anything.

python/cac_tripplanner/destinations/views.py
```diff
@@ -47,10 +47,10 @@ def home(request):
     }
     if request.GET.get('destination') is not None:
         # If there's a destination in the URL, go right to directions
-        context['tab'] = 'map'
+        context['tab'] = 'map-dir'
```
flibbertigibbet: Might consider renaming to `map-directions`. `dir` makes me think 'directory'.

python/cac_tripplanner/templates/home.html
```diff
@@ -54,46 +54,45 @@
...
         {% include "partials/article-card.html" with show_link=True %}
     {% endif %}
 
     <div class="directions-results">
-        <div class="sk-spinner sk-three-bounce hidden">
-            <div class="sk-child sk-bounce1"></div>
-            <div class="sk-child sk-bounce2"></div>
-            <div class="sk-child sk-bounce3"></div>
-        </div>
+        {% include "partials/spinner.html" %}
```
flibbertigibbet: :sunglasses:

python/cac_tripplanner/templates/partials/footer.html
```diff
@@ -19,8 +19,9 @@
             </p>
         </a>
     </div>
-    <div class="footer-compact">
-        <a href="{% url 'home' %}">Explore</a>
+    <div class="footer-compact tab-control">
+        <a class="nav-item explore-link" href="#" data-tab-id="EXPLORE"">Explore</a>
+        <a class="nav-item directions-link" href="#" data-tab-id="DIRECTIONS"">Directions</a>
```
flibbertigibbet: double quotation marks at the end of both `data-tab-id`s here

KlaasH:
Turns out the Home tab link issue can be fixed by just making it a regular link to root and setting the right tab ID.  Now it'll actually go to Home from the Learn or Place Detail views because there's nothing listening for tab clicks so it just follows the link.  When the Home page controller is loaded, the click will be intercepted and do nothing, which I think makes sense.  People can hit the logo or refresh the page to clear everything, I think having the Home tab do it would be more of an unpleasant surprise than a useful feature.

The Explore tab link is a little more complicated, so I'm going to open an issue about that.

src/app/scripts/cac/pages/cac-pages-home.js
```diff
@@ -201,7 +214,20 @@ CAC.Pages.Home = (function ($, ModeOptions,  MapControl, TripOptions, SearchPara
         }
 
         // load directions if origin and destination set
-        $(document).ready(directionsControl.setFromUserPreferences());
+        $(document).ready(loadInitialMethod);
+    }
+
+    function loadInitialMethod() {
+        if (!UserPreferences.isDefault('method')) {
+            var method = UserPreferences.getPreference('method');
```
flibbertigibbet: Just noting for future reference: I suspect we can ditch the `method` user preference altogether, since it's not state that survives refresh.

KlaasH: Just made reference to this bit of code in issue #715, but if it doesn't turn out to be useful for that I think you're right that it can go.

PR #700, opened 2016-12-12
Feature/fix timing browser bug

### LINKED ISSUE #698: Show 24h of depart/arrive timing options when not "Today"
Time menu only showing "Now" to 11:45 pm even if I change to a future date.

Repro:
1. Get any directions.
1. Options > Arrive/depart at…
1. Change from "Today" to a future date.
1. Time menu says "Now".
1. Open time menu. Can't choose a time before Now.

Proposed behavior:
- When user changes date from "Today" to a future date, the selected time option should change from "Now" to the next 15-min interval closest to now (eg, 1:15 pm). If the user opens the time menu, it should list all 15-min intervals from 12:00 am to 11:45 pm.

- When the user changes the date from a future date to "Today", the selected time option should remain at whatever it's currently set to — unless it's earlier than now, in which case set it to Now. If the user opens the time menu, it should only list options from Now to 11:45 pm.

flibbertigibbet:
Fixes #698.

Also fixes issue where times before 'now' would still be visible if a future date was selected and then 'clear' button is clicked.

### DIFF
--- a/src/app/scripts/cac/control/cac-control-trip-options.js
+++ b/src/app/scripts/cac/control/cac-control-trip-options.js
@@ -487,11 +487,12 @@ CAC.Control.TripOptions = (function ($, Handlebars, moment, Modal, UserPreferenc
      * Helper to listen for change events on the day drop-down created by timingModalOptions
      */
     function addDayDropDownListener() {
-        $('#' + options.selectors.dayOptionsId).on('change', function(e) {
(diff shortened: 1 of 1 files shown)


flibbertigibbet:
Merging for redeploy.

This is a pretty simple fix, just changing the event to listen to and moving some logic into a new method.

PR #701, opened 2016-12-13
Feature/fixup transit toggle

flibbertigibbet:
Fixes up some transit-toggle related refactor work from #693 that was missing in #694.

Also bumps `pytz` timezone library version to latest release.


flibbertigibbet:
Merging for staging redeploy

PR #702, opened 2016-12-13
Add migration for destinations meta

flibbertigibbet:
Add missing migration for change to destinations model meta default sort.

Meta change made in 649465b25d7fe82891e93d512657c23137702f4b and apparently requires a database migration.


flibbertigibbet:
trivial, merging

PR #704, opened 2016-12-14
UI polish

lederer:
- Added "Map View" button to home page.
- Styled waypoints and waypoint popup.
- Lighter background color for selected route in sidebar.
- Focus on origin field when returning Home.


lederer:
Yup, Map View is (deliberately) redundant with the Explore tab. User testing found it helpful.

flibbertigibbet:
Okay, I see their is a back button from the itinerary list. There's an issue where after dragging a waypoint it no longer goes automatically into the first itinerary details, which made it look like the back button went away. Looks like that's also no longer the behavior on production. Will make an issue for that.

PR #710, opened 2016-12-14
Feature/options open from banner

### LINKED ISSUE #708: Open options modal from green banner
As does the blue 'need wheels' banner, green banner with current option selections should also open the trip options modal on click.

flibbertigibbet:
Closes #708.

Open trip options when either green trip info sidebar banner clicked, or blue "need wheels?" banner.
Also use pointer cursor to indicate sidebar banner is clickable.


flibbertigibbet:
trivial; merging

PR #717, opened 2016-12-20
Support browser history navigation

### LINKED ISSUE #707: URL router does not work with browser back button
If after following some links on the site (tab switch, go to map page) that do not cause full page refresh, user clicks to go back, it will leave the site. Expected behavior would be to return to the last view.

Not sure if preserving state is something the current URL router library in use can support easily; if so, would be good to implement state stack.

KlaasH:
The URL routing package we're using supports pushing to browser history,
but it was turned off because it would create a weird loop when we update
the URL from the directions or explore controllers.

This enables it, which required
1. Some checking in the URL router to figure out when we're
   a) updating from code, in which it updates the URL but skips the routing
      callback or
   b) updating from code to the same URL, in which case it does nothing
2. Getting the relevant controllers (tab, directions form, mode options,
   trip options, directions, explore) to update for the new user
   preferences set by the URL router. This is done by throwing an event
   that the Home controller uses to call the relevant methods on all the
   subcontrollers.

**To test**
Click around!  Any change that affects the URL (changing tabs, changing mode, changing settings) should now cause the resulting view to get pushed to browser history as though it were a normal page.  The back button should work and, after going back, forward should work as well.

Resolves #707.

### DIFF
--- a/src/app/scripts/cac/pages/cac-pages-home.js
+++ b/src/app/scripts/cac/pages/cac-pages-home.js
@@ -125,10 +125,14 @@ CAC.Pages.Home = (function ($, ModeOptions,  MapControl, TripOptions, SearchPara
 
         mapControl.events.on(mapControl.eventNames.mapMoved, SearchParams.updateMapCenter);
 
+        tabControl.events.on(tabControl.eventNames.tabShown, onTabShown);
+
         modeOptionsControl.events.on(modeOptionsControl.eventNames.toggle, toggledMode);
 
         directionsFormControl.events.on(directionsFormControl.eventNames.selected,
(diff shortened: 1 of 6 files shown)


src/app/scripts/cac/pages/cac-pages-home.js
```diff
@@ -289,6 +297,20 @@ CAC.Pages.Home = (function ($, ModeOptions,  MapControl, TripOptions, SearchPara
         showHideNeedWheelsBanner();
     }
 
+    // Handler for prefernces changes due to URL change, i.e. browser back/forward
```
flibbertigibbet: typo ('preferences')

Issue #656, opened 2016-11-29
Incorporate design changes for explore


flibbertigibbet:
Implement the changes made to the prototype in #641.

#593 and #594 depend on this being done first.

PR #718, opened 2016-12-20
Feature/isochrone scrubber

### LINKED ISSUE #594: Add time slider for isochrone
To explore map, debounced.

Control goes where destination field is on directions form.

Depends on #593.

flibbertigibbet:
Closes #594.

image

Adds a slider to the explore view in the place where the `to` field appears in directions view.

Does not attempt to cache response client side (would be too much to store on mobile). Limits options to 15, 30, 45, and 60 minutes, to minimize server load.

@lederer this could probably use some styling attention; wasn't sure where to label selected value. Also, should the selection be also reflected in the green bar? If so, will likely overflow to a second line.

### DIFF
--- a/python/cac_tripplanner/templates/home.html
+++ b/python/cac_tripplanner/templates/home.html
@@ -41,6 +41,11 @@
                 <button class="btn-reverse" title="Swap starting point and destination"
                     tabindex="4" type="button" name="reverse"><i class="icon-reverse"></i></button>
             </div>
+            <div class="isochrone-control directions-text-input hidden">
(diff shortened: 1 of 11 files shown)


flibbertigibbet:
Just realized #641 added a slider to the prototype; will revise for styling.

flibbertigibbet:
Updated for slider styling from #641. Which partially completes #656, but does not rename form controls (maybe we can take a pass on that?) or add styling to hide explore on mobile (maybe we can try it out first and see how well it works?)

Issue #721, opened 2016-12-22
Geocode error handler only triggers for destination


flibbertigibbet:
When user drags origin or destination marker to an invalid location (including anywhere outside the DVRPC area), an error message should display. However, if the origin is bad, the marker disappears and the itineraries clear, and the itinerary error message display triggers, but does not show.

Bad destination (correct behavior):
![image](https://cloud.githubusercontent.com/assets/960264/21439981/c5293360-c85e-11e6-86a0-bed30fbbdde6.png)

Bad origin (should have error message):
![image](https://cloud.githubusercontent.com/assets/960264/21440026/eef339ac-c85e-11e6-8ff4-0e17048d875c.png)

(Note: the error message styling there is currently WIP. Right now it's just bolded text in the sidebar.)

flibbertigibbet:
Found it: both directions and isochrone tabs have `onGeocodeError` handlers only partially guarded by tab check. The one for the isochrone tab would run on any origin change, tromping the sidebar DOM the directions tab is also attempting to modify.

PR #648, opened 2016-11-21
Feature/itinerary alert

CloudNiner:
## Overview

Re-adds the alert displayed for trips involving public transit + bike. 

## Demo

cac-itinerary-bike-alert

## Testing

Plan a trip using public transit + bike, the alert will appear below the directions header. If removed, it will reappear each time the directions list is re-rendered.

A bit of styling cleanup potentially necessary before #614 can be closed.

PR #723, opened 2016-12-22
Feature/update alerts

### LINKED ISSUE #614: Implement new itinerary messages
There will be a new banner style for the messages that may appear at the top of the itinerary details, such as the message containing transit agency bike policy information link.

Replace existing banner with new design.

flibbertigibbet:
Updates sidebar alerts for new styling. Closes #614 and #721.

#648 only updated the bicycle message alert, which is handled separately to allow for the embedded links. This updates the sidebar alerts for directions and isochrone errors. Easy way to check: try dragging a marker into the Atlantic. Alerts should be close with a click on the 'x' button and disappear on requery or page/tab change.

@lederer: I added some alert classes in keeping with the Bootstrap alert types we'd been using before; currently we're only using the `danger` type.

Also fixed an issue where navigating back to the home tab from the map page would leave behind a little white box between the form controls and the destinations list.

### DIFF
--- a/src/app/styles/pages/_home.scss
+++ b/src/app/styles/pages/_home.scss
@@ -12,8 +12,4 @@
     height: 100%;
     overflow: auto;
     -webkit-overflow-scrolling: touch;
-
-    .places {
-        display: none;
-    }
 }
--- a/src/app/styles/pages/_home.scss
+++ b/src/app/styles/pages/_home.scss
@@ -12,4 +12,8 @@
     height: 100%;
     overflow: auto;
     -webkit-overflow-scrolling: touch;
+
+    .places {
+        display: none;
+    }
 }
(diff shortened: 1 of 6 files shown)


src/app/styles/pages/_home.scss
```diff
@@ -12,4 +12,8 @@
     height: 100%;
     overflow: auto;
     -webkit-overflow-scrolling: touch;
+
+    .places {
+        display: none;
+    }
```
KlaasH: No places on home view?

flibbertigibbet:
White box on home view appears if there are no places, as happens if last origin geocode failed.

image

flibbertigibbet:
Dropped the `places` hide; will look at addressing the white box on homepage with #719.

Issue #715, opened 2016-12-16
Explore tab link behavior


KlaasH:
Part 1, a question:
Should the Explore tab link do something different from what the "Map View" button on the places list does, i.e. open the Explore tab, whether or not there's an origin?  My assumption has been no, based on the fact that clicking Explore from https://marvelapp.com/2he4i01/screen/16206347 goes to https://marvelapp.com/2he4i01/screen/16206348.

Part 2, a thing to fix:
On the Learn tab and the place detail view (which is labeled as part of the Explore tab but which is a separate Django view), the tab links are actual links.  The behavior of the Explore tab on the Home view is a javascript action, telling tabControl to activate the tab, but that only works when the Home view is already loaded.  We don't currently have a way to go straight to the Explore tab by link when no origin is set.

One option would be to use the URL router to detect a path or parameter and set the UserPreferences method to 'explore', then clear the URL or parameter.  The Home view checks UserPreferences method on load and goes to the corresponding view if `isDefault` is false.

Doing only that might cause nasty flashing, though, because the body class won't be set in the Django template.  So if there's a way to make a direct link work in Django that would probably be preferable.

lederer:
@KlaasH Re part 1: Correct! The Explore tab link and the Map View button should have precisely the same behaviors, both for when an origin is set and when an origin is not set.

flibbertigibbet:
This is the current (same) behavior.

PR #725, opened 2017-01-03
Feature/klaas updates time to places

### LINKED ISSUE #557: Implement places minutes from origin

flibbertigibbet:
Cherry-pick of flibbertigibbet/cac-tripplanner#1 with merge conflicts fixed into #715. Supersedes both.

Closes #557.


flibbertigibbet:
056faf0 sorts out white box on homepage/places list reset.

src/app/scripts/cac/pages/cac-pages-home.js
```diff
@@ -305,12 +305,12 @@ CAC.Pages.Home = (function ($, ModeOptions,  MapControl, TripOptions, SearchPara
...
     function onUrlChanged() {
-        setActiveTab();
         modeOptionsControl.setMode(UserPreferences.getPreference('mode'));
-        directionsControl.setFromUserPreferences();
-        exploreControl.setFromUserPreferences();
         directionsFormControl.setFromUserPreferences();
         showHideNeedWheelsBanner();
+        directionsControl.setFromUserPreferences();
+        exploreControl.setFromUserPreferences();
+        setActiveTab();
```
KlaasH: I think there's a catch-22 here, or at least a bit of a tangled web.  I may have been wrong in moving toward pushing "do the right thing" logic out of Home and into the controllers.  With `setActiveTab()` last I'm seeing a bug whereby using browser navigation to go from Explore to Home results in the URL getting rewritten to the Explore URL because `exploreControl.setFromUserPreferences();` gets triggered while Explore is still active, so it calls `clickedExplore`, which runs (debounced) and updates the URL to match the parameters it's running with.

It's the same series of events I was talking about [here](https://github.com/azavea/cac-tripplanner/pull/719#discussion_r93785240), but I didn't realize that superfluous calls on exiting tabs would have visible negative effects.  Here's one possibility, which seems inelegant but is easier than rethinking the whole setup:

```
diff --git a/src/app/scripts/cac/control/cac-control-tab.js b/src/app/scripts/cac/control/cac-control-tab.js
index a1db9c8..77c2834 100644
--- a/src/app/scripts/cac/control/cac-control-tab.js
+++ b/src/app/scripts/cac/control/cac-control-tab.js
@@ -79,6 +79,9 @@ CAC.Control.Tab = (function ($) {
             }
 
             this.events.trigger(eventNames.tabShown, tabId);
+            return true;
+        } else {
+            return false;
         }
     }
 
diff --git a/src/app/scripts/cac/pages/cac-pages-home.js b/src/app/scripts/cac/pages/cac-pages-home.js
index d6ee99b..dba61e0 100644
--- a/src/app/scripts/cac/pages/cac-pages-home.js
+++ b/src/app/scripts/cac/pages/cac-pages-home.js
@@ -224,12 +224,12 @@ CAC.Pages.Home = (function ($, ModeOptions,  MapControl, TripOptions, SearchPara
         if (!UserPreferences.isDefault('method')) {
             var method = UserPreferences.getPreference('method');
             if (method === 'directions') {
-                tabControl.setTab(tabControl.TABS.DIRECTIONS);
+                return tabControl.setTab(tabControl.TABS.DIRECTIONS);
             } else if (method === 'explore') {
-                tabControl.setTab(tabControl.TABS.EXPLORE);
+                return tabControl.setTab(tabControl.TABS.EXPLORE);
             }
         } else {
-            tabControl.setTab(tabControl.TABS.HOME);
+            return tabControl.setTab(tabControl.TABS.HOME);
         }
     }
 
@@ -308,9 +308,11 @@ CAC.Pages.Home = (function ($, ModeOptions,  MapControl, TripOptions, SearchPara
         modeOptionsControl.setMode(UserPreferences.getPreference('mode'));
         directionsFormControl.setFromUserPreferences();
         showHideNeedWheelsBanner();
-        directionsControl.setFromUserPreferences();
-        exploreControl.setFromUserPreferences();
-        setActiveTab();
+        var tabChanged = setActiveTab();
+        if (!tabChanged) {
+            directionsControl.setFromUserPreferences();
+            exploreControl.setFromUserPreferences();
+        }
     }
 
     function closedTripModal(event) {
```

I.e. since `onTabSelected` methods run `setFromUserPreferences` when the tab is entering, only manually run it when the URL change doesn't result in a tab change.  Or it might make sense to only run the `setFromUserPreferences` for the active tab.

flibbertigibbet:
Just pushed a fix for a typeahead control issue where it would default/reset to `null`, but when the field has focus, it's an empty string. It was causing superfluous field change events to fire when an empty field had focus, preventing click events elsewhere from firing.

Also shuffled around `setFromUserPreferences` calls to only happen for the activating tab; not sure if we still need a separate issue to resolve that completely.

src/app/scripts/cac/control/cac-control-explore.js
```diff
@@ -296,8 +299,6 @@ CAC.Control.Explore = (function (_, $, Geocoder, MapTemplates, HomeTemplates, Ro
         if (exploreLatLng && tabControl.isTabShowing(tabControl.TABS.EXPLORE)) {
             clickedExplore();
         }
-
-        getNearbyPlaces();
```
KlaasH: Without this call, times don't load when the Explore view is loaded from URL.  With this call, they get loaded twice.  :frowning: 

Passing `{leading: false}` to the `throttle` above call solves that and seems reasonable to me.

flibbertigibbet: Calling it as part of `setActiveTab` in the home controller instead seems sufficient and results in no extra calls. Modifying the throttle behavior would just mask an extraneous call being made and introduce some extra delay.

src/app/scripts/cac/pages/cac-pages-home.js
```diff
@@ -230,6 +230,7 @@ CAC.Pages.Home = (function ($, ModeOptions,  MapControl, TripOptions, SearchPara
             } else if (method === 'explore') {
                 exploreControl.setFromUserPreferences();
                 tabControl.setTab(tabControl.TABS.EXPLORE);
+                exploreControl.getNearbyPlaces();
```
KlaasH: :+1: Yup, this is better than what I was suggesting.  I don't see any remaining problems when moving between views.

PR #731, opened 2017-01-05
Color highlighted route by mode

### LINKED ISSUE #440: Coloring different modes (enhancement)
A user suggested that the line sections for bike/walk/transit sections of trips be colored slightly differently.

flibbertigibbet:
Color each section of highlighted route by mode.

Closes #440.

image

### DIFF
--- a/src/app/scripts/cac/routing/cac-routing-itinerary.js
+++ b/src/app/scripts/cac/routing/cac-routing-itinerary.js
@@ -32,10 +32,12 @@ CAC.Routing.Itinerary = (function ($, cartodb, L, _, moment, Geocoder, Utils) {
         // not actually GeoJSON, but a Leaflet layer made from GeoJSON
         this.geojson = cartodb.L.geoJson({type: 'FeatureCollection',
                                           features: this.getFeatures(otpItinerary.legs)});
-        this.geojson.setStyle(getStyle(true, false));
 
-        // expose functions
-        this.getStyle = getStyle;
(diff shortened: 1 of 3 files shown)


flibbertigibbet:
@lederer I went with some random colors and they look pretty hideous. They're defined [here](https://github.com/azavea/cac-tripplanner/pull/731/files#diff-dfce195bf102934a6762081645bacfbaR91).

src/app/scripts/cac/routing/cac-routing-itinerary.js
```diff
@@ -290,30 +292,41 @@ CAC.Routing.Itinerary = (function ($, cartodb, L, _, moment, Geocoder, Utils) {
...
+            return;
         }
+
         var defaultStyle = {clickable: true, // to get mouse events (listen to hover)
-                            color: '#d02d2d',
-                            dashArray: null,
-                            lineCap: 'round',
-                            lineJoin: 'round',
-                            opacity: 0.75};
+                        color: '#d02d2d',
```
KlaasH: It would be nice to have the colors defined all in one place and not duplicated.  What about adding a `function getDefaultModeColor(highlighted) { ... }` to Utils?  Plus then you could just call that to set the color here and wouldn't have to replace it in the background case below.

flibbertigibbet: `Utils` module export is frozen, so can safely refer directly without a getter.

PR #732, opened 2017-01-05
Activate directions button on place cards

### LINKED ISSUE #545: Implement 'Go Here' button from places
When user clicks 'Go Here' button on a place card on the home page, open directions to the place.

KlaasH:
Makes the "Directions" button on the "Places we love" cards take you to the directions view with the selected place as the destination.  If there's an origin set it will generate directions.  If not you'll get the map but no directions, just as you would see if you followed a one-sided directions link.

Resolves #545.

### DIFF
--- a/src/app/scripts/cac/control/cac-control-directions-form.js
+++ b/src/app/scripts/cac/control/cac-control-directions-form.js
@@ -32,8 +32,6 @@ CAC.Control.DirectionsFormControl = (function ($, Typeahead, Geocoder, UserPrefe
     };
     var options = {};
 
-    var typeaheadTo = null;
-    var typeaheadFrom = null;
     var typeaheads = {};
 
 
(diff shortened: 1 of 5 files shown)


src/app/scripts/cac/control/cac-control-directions-form.js
```diff
@@ -95,6 +95,12 @@ CAC.Control.DirectionsFormControl = (function ($, Typeahead, Geocoder, UserPrefe
         events.trigger(eventNames.selected, [key, location]);
     }
 
+    // For setting origin or destination from code, e.g. directions links
+    function setLocation(key, location, text) {
```
flibbertigibbet: This does not set the text for the typeahead `to` box.

KlaasH: Whoops.  I changed things around a little and must have checked it with a stale tab.  Filling in the typeahead was definitely the intention.  Stand by...

KlaasH:
Did those two things (destination text gets set, origin field goes red and gets selected when they click a directions link without having set an origin).

Also got annoyed with the `typeaheadFrom` and `typeaheadTo` variables in the directions from controller and refactored them into a single object with `origin` and `destination` keys (f08a735 and 65768db).  Does that seem good to you, or do you prefer the old way?  I felt strongly enough to do the find-and-replace but no stronger than that, and it'd be easy to throw those commits away.

flibbertigibbet:
I'm fine with the `typeaheads` object; looks cleaner.

PR #735, opened 2017-01-06
Highlight marker when mouse is over place card

### LINKED ISSUE #728: Hover interactivity for places in map view
When hovering over a place in the sidebar list, highlight place map marker and change card background. @lederer may need to style.

See #713.

KlaasH:
Restores pre-redesign functionality of highlighting the marker on the map corresponding to a featured location when you hover over the card for that location in the sidebar, and highlighting plus centering the map on a location if you click on its card.

Styling needed for what color the markers should change to and for what the cards themselves should do when hovered. Pre-redesign, they get ever-so-slightly lighter, due to [this style](https://github.com/azavea/cac-tripplanner/blob/develop/src/app/styles/components/_blocks.scss#L24).

@lederer: how would you like to do the styling?  You could add a commit to this branch, or describe what to do, or we could merge this as-is and you could get it in a separate styling PR, ...

marker hover

Current production:
hover marker old

With styling, resolves #728.

### DIFF
--- a/src/app/scripts/cac/map/cac-map-isochrone.js
+++ b/src/app/scripts/cac/map/cac-map-isochrone.js
@@ -27,7 +27,7 @@ CAC.Map.IsochroneControl = (function ($, Handlebars, cartodb, L, turf, _) {
     var highlightIcon = L.AwesomeMarkers.icon({
         icon: 'default',
         prefix: 'icon',
-        markerColor: 'orange'
+        markerColor: 'blue',
     });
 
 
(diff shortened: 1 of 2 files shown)


flibbertigibbet:
I think for now at least it would be good to go with the same light-blue highlight color as used for itinerary card hover interactivity.

lederer:
@KlaasH I say merge when ready and I'll follow up with a styling PR.

lederer:
Turns out no styling PR needed. LGTM.

PR #738, opened 2017-01-06
Mode colors in map and sidebar.

lederer:
Applied brand colors to mode segments:

- walk, ferry: blue
- bike: purple
- bus: yellow
- all rails: orange
- car: black

Also color coded the mode icons in the sidebar to serve as a key. Only in the "list of routes" mode, not in the step-by-step route detail mode (yet?). 

Worried this may be too colorful. We can discuss.

Also removed the red/blue path in the sidebar route summaries, since the colors won't match. Adjusted the selected/unselected visual treatment of summaries to compensate (though this probably needs a bit more work).

screen shot 2017-01-06 at 4 26 09 pm


lederer:
If we stick with these colors, we'll probably need to come up with new (off-brand) colors for the sidebar banners (options and alerts), to avoid semantic confusion. They currently use on-brand green and yellow.

PR #744, opened 2017-01-13
Implement directions button in marker popup

### LINKED ISSUE #743: Get directions from marker popup
'Get Directions' link in destination popup should go to directions with destination set, as does button on place card.

flibbertigibbet:
For when user clicks to get directions to a place from the place marker popup window.

Closes #743.

### DIFF
--- a/src/app/scripts/cac/pages/cac-pages-home.js
+++ b/src/app/scripts/cac/pages/cac-pages-home.js
@@ -328,15 +328,14 @@ CAC.Pages.Home = (function ($, ModeOptions,  MapControl, TripOptions, SearchPara
         event.preventDefault();
 
         var placeCard = $(event.target).closest(options.selectors.placeCard);
-        var placeId = placeCard.data('destination-id');
 
         var destination = {
-            id: placeId,
+            id: placeCard.data('destination-id'),
             address: placeCard.find(options.selectors.placeCardName).text(),
             location: { x: placeCard.data('destination-x'), y: placeCard.data('destination-y') }
         };
 
-        goToDestinationDirections(placeId, destination);
+        goToDestinationDirections(destination);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


flibbertigibbet:
The event gets passed through the map control because the isochrone control that it owns does not get instantiated until the map has loaded, and the home controller may not yet have a loaded map when it's setting up the listener for the event.

src/app/scripts/cac/pages/cac-pages-home.js
```diff
@@ -343,19 +321,41 @@ CAC.Pages.Home = (function ($, ModeOptions,  MapControl, TripOptions, SearchPara
...
+    /**
+     * When user clicks 'Get Directions' in a destination marker popup
+     */
+    function clickedDestinationDirectionsPopup(event, place) {
+        // match the format expected
+        place.location = {
+            x: place.point.coordinates[0],
+            y: place.point.coordinates[1]
+        };
+        goToDestinationDirections(place);
```
flibbertigibbet: Not sure if I'm missing something, but that's not the behavior I'm seeing.

Clicking 'Get Directions' in the popup (as for this handler), clicking on the place card, and autocomplete all fill in the place name and not the address in the typeahead form box.

src/app/scripts/cac/pages/cac-pages-home.js
```diff
@@ -343,19 +321,41 @@ CAC.Pages.Home = (function ($, ModeOptions,  MapControl, TripOptions, SearchPara
...
         var placeId = placeCard.data('destination-id');
-        UserPreferences.setPreference('placeId', placeId);
+
         var destination = {
+            id: placeId,
             address: placeCard.find(options.selectors.placeCardName).text(),
             location: { x: placeCard.data('destination-x'), y: placeCard.data('destination-y') }
         };
+
+        goToDestinationDirections(placeId, destination);
```
flibbertigibbet: Right, just pushed a fix for that.

flibbertigibbet:
Okay, glad it's fine. Maybe that was an artifact of the function call with the extra parameter I'd fixed.

Issue #713, opened 2016-12-16
Restore/implement higlighting and limiting destinations in Explore sidebar


KlaasH:
In the olden days, mousing over a destination in the sidebar would highlight the marker on the map.  I assume that behavior should be restored.

It also appears from the code, but doesn't happen in production, that there was an idea of limiting the list of destinations to ones that are reachable within the given time limit.  This also seems to be the behavior shown in the wireframe: https://marvelapp.com/2he4i01/screen/16206349.  So that should be implemented if it's desired.  If not, it might make sense to add some separation between the destination logic and the isochrone logic.

flibbertigibbet:
Original implementation was to filter to limit to destinations within the isochrone. It got modified to show all instead and ignore the isochrone bounds due to (non-developer) feedback.

PR #745, opened 2017-01-14
Feature/load destinations always

### LINKED ISSUE #714: Show all destinations on Explore map when no origin set
Related to #713.  If we don't want to limit destinations by what's reachable, then we should just be showing all destinations whenever the Explore tab is active.  If we do want to limit to reachable destinations, then we need to have a codepath that shows all destinations when no origin is set.

flibbertigibbet:
Closes #714 and #713.

Always load destination markers on explore tab, even when no origin set (and no isochrone shown).

Also modifies isochrone call to fetch it directly from OTP, instead of the Django endpoint that filters for destinations within isochrone, as those destinations aren't used. (Django endpoint left in, should we have use for it later.)

@lederer: Destination map marker popups still need styling attention, at least to get the `Get Directions` link to pull right, as the bootstrap class for that is no longer going to work.

### DIFF
--- a/python/cac_tripplanner/destinations/views.py
+++ b/python/cac_tripplanner/destinations/views.py
@@ -17,13 +17,14 @@ from cms.models import Article
 DEFAULT_CONTEXT = {
     'debug': settings.DEBUG,
     'fb_app_id': settings.FB_APP_ID,
+    'isochrone_url': settings.ISOCHRONE_URL,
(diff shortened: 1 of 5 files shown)


flibbertigibbet:
Good to know. I hadn't seen the markers hanging around, but will keep an eye out for that.

PR #747, opened 2017-01-16
Feature/disable mobile explore

### LINKED ISSUE #742: Disable explore mode on mobile
As in prototype

flibbertigibbet:
Hide explore tab header button on mobile, effectively disabling explore mode for mobile devices.

Implement toggling modes with label for `to` field.

Closes #742. Also adds text label with the number of minutes for the explore mode slider, as was in #641.

See #641 for source for most of the styling. Also got some input from @lederer on behavior (decided to hide explore tab link on mobile).

### DIFF
--- a/src/app/styles/components/_directions-form.scss
+++ b/src/app/styles/components/_directions-form.scss
@@ -419,17 +419,13 @@
      #isochrone-slider {
         @include inputrange();
         margin-right: 1rem;
-        width: 70%;
+        width: 80%;
     }
 
-    &.directions-destination-mode-within {
-
-        .directions-destination.isochrone-control {
-            display: flex;
-        }
-
-        .directions-destination.directions-text-input {
-            display: none;
+    .directions-tab-button.isochrone-control {
+            &:after {
+                background: $gophillygo-blue;
+            }
         }
 
         .directions-to label {
@@ -451,9 +447,8 @@
                 content: ' min';
             }
         }
-    }
 
-    @include respond-to('sm-up') {
+    @include respond-to('md-up') {
         .directions-tab-button label {
             position: relative;
             background-color: $gophillygo-blue-dark;
--- a/src/app/styles/components/_directions-form.scss
+++ b/src/app/styles/components/_directions-form.scss
@@ -53,12 +53,6 @@
         }
     }
 
-    #isochrone-slider {
-        @include inputrange();
-        margin-right: 1rem;
-        width: 75%;
-    }
-
     .isochrone-control {
         display: flex;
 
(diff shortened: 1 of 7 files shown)


flibbertigibbet:
Smallest breakpoint:
image

Larger, but still mobile device breakpoint:
image

All else (desktop/tablet):
image

src/app/styles/components/_directions-form.scss
```diff
@@ -411,4 +405,60 @@
...
+         }
+     }
+
+     #isochrone-slider {
+        @include inputrange();
+        margin-right: 1rem;
+        width: 70%;
+    }
+
+    &.directions-destination-mode-within {
```
KlaasH: I don't see this class anywhere, which seems related to the slider text looking rather un-styled
image

flibbertigibbet: Yeah, some classes were renamed in the prototype merge, for which I didn't refactor our base.

flibbertigibbet:
Regarding the to/within toggle, I agree it is very redundant and suggested yesterday we simply not implement it, but Scott wants it in, so here it is. It came up because it was tied up in the styling around the slider bar, and added a breakpoint that at first seemed intended to hide explore mode on mobile, but really only served the purpose of showing/hiding the toggle button background change.

Cleaned up the styling some more (putting it in was complicated by a bunch of confusing class renaming in the prototype) and disabled the toggle label functionality on mobile.

KlaasH:
- What about only showing the toggle in desktop sidebar mode, i.e. also disabling it (and returning the styling to white background) on the Home view?  I still feel that having a click on "to" from a blank Home view take you to the Explore tab is weird and not useful.

- The time for the slider isn't getting set when you load from URL: 
image
(e.g. [this](http://localhost:8024/?origin=39.9616133%2C-75.1541914&originText=990%20Spring%20Garden%20St%2C%20Philadelphia%2C%20Pennsylvania%2C%2019123&mode=WALK%2CTRANSIT&maxWalk=482802&wheelchair=false&bikeTriangle=any&arriveBy=false&dateTime=&placeId=&exploreMinutes=30) shows a 30-minute isochrone and sets the slider in the right spot but says '15 min')

flibbertigibbet:
That's the current behavior: the to/within toggle only displays when the map is loaded, which is also the breakpoint for "sidebar mode." The change from white to blue background is unrelated and happens on a smaller, mobile breakpoint. I still agree it's not useful.

Also: the slider updates work for me. Any hints on how to reproduce?

flibbertigibbet:
Oh, I see. You've directly (re-)loaded a page with an explore time set on the URL, but the slider doesn't reflect that.

flibbertigibbet:
Okay, slider text initialization tidied.

KlaasH:
"sidebar mode" wasn't the clearest description.  I meant it to refer only to the desktop versions of the Home and Explore views, i.e. when there's a map with a sidebar actually showing.  The thing I was suggesting nixing was the toggle on the desktop Home view, i.e.:
image

Because if I click on that I'll get the same thing I would get by clicking the Map View button:
image

which seems strange and not useful.  So in this scenario the to/within toggle is active *only* when the form is in the sidebar next to the map (i.e. the desktop Home and Explore views).

flibbertigibbet:
Okay, so you mean in the map view and not on the home view. Yes, that is a redundant, third way to get to the explore view, and an intentional part of the design as I understand it.

Maybe create a separate issue and assign to @lederer?

lederer:
Sorry, just noticing this thread now. FWIW the to/within toggle will only be active in desktop map view. Not in desktop home view or in any mobile views. 

It is intended as a contextually meaningful way to move between map view modes (in contrast to the Explore button in the sidebar footer, which has always been a strange and distant position for what amounts to a mode toggle; the client was keen on replicating the main Explore/Learn nav somehow in map view, however, and this will do for now).

The dark blue bg of the mode toggle is taken from the dark blue bg of the options button in the top right. Basically that color is used to indicate a sense of secondary interactivity. Ie, the options button is of a different sort than the controls next to it, and the to/within button is of a different sort than the 'from' label. Agree it's not the most elegant solution, and I'm fiddling with an idea for affording the toggle iconographically as well (or instead). See below. But again it'll do for now.

image

flibbertigibbet:
Should be current behavior now #759 is in.

PR #759, opened 2017-01-24
Disable explore toggle in home view

### LINKED ISSUE #753: Unbutton "to" label in home view
In Home view, clicking the destination "to" label should not switch to Map view, and its styling should be the same as the origin "from" label.

That label should only serve as a (to/within) toggle while in Map view.

After the behavior fix, feel free to reassign back to me for the styling fix.

flibbertigibbet:
Closes #753.


flibbertigibbet:
Good point. Pushed a fix to check breakpoint width directly, which is unfortunately brittle, but threw in some comments.

PR #762, opened 2017-01-24
Hide options bar when no options are overridden

### LINKED ISSUE #748: Hide options bar when no options set
In map view, the green options bar should only appear while one or more options is not in its default state.

Currently it is shown even when no options are set:
image

KlaasH:
Makes the options bar stop showing up when none of the options have been changed.

Since the definition of "changed" includes "set by the user to a default value", any page loaded from URL was showing up as non-default, since all the options get written into the URL then saved to UserPreferences on load.  I dealt with that by adding an option to `UserPreferences.setPreference` to ignore default values (i.e. only save a preference if it differs from the default value) and made the URL routing controller use that option in its calls.

There was also an issue whereby the directions list was always setting the `body-has-sidebar-banner` on close, which messed up the sidebar height if there was not in fact a sidebar banner active.  I dealt with that by having it check the DOM for non-hidden sidebar banners.  Connecting it directly to the code that controls the banners, which is in CAC.Pages.Home, would in principle be better but is tricky since they're a few layers apart in the controller hierarchy.

Defaults:
image 

Not default:
image

Resolves #748.

### DIFF
--- a/src/app/scripts/cac/control/cac-control-directions-list.js
+++ b/src/app/scripts/cac/control/cac-control-directions-list.js
@@ -28,7 +28,8 @@ CAC.Control.DirectionsList = (function (_, $, ShareModal, MapTemplates) {
             googlePlusShareButton: '.modal-list-google',
             shareModalButton: '.share-directions',
             stepByStepClass: 'body-step-by-step',
-            sidebarBannerClass: 'body-has-sidebar-banner'
+            sidebarBannerClass: 'body-has-sidebar-banner',
+            activeSidebarBanner: '.sidebar-banner:not(.hidden)'
         }
     };
     var options = {};
@@ -55,7 +56,6 @@ CAC.Control.DirectionsList = (function (_, $, ShareModal, MapTemplates) {
         setItinerary: setItinerary,
         show: show,
         hide: hide,
-        toggle: toggle
     };
 
     return DirectionsListControl;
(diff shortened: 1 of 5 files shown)


src/app/scripts/cac/pages/cac-pages-home.js
```diff
@@ -429,27 +445,32 @@ CAC.Pages.Home = (function ($, ModeOptions,  MapControl, TripOptions, SearchPara
...
-        var $banner = $(options.selectors.sidebarTripOptionsBanner);
         $banner.html(html);
+        $(options.selectors.mapContainer).addClass(options.selectors.sidebarBannerClass);
         $banner.removeClass(options.selectors.hiddenClass);
     }
 
     /**
      * The 'need wheels?' sidebar banner should only display when trip options have
-     * never been seen and currently in bicycle mode. Check on initial load and mdoe toggle.
+     * never been seen and currently in bicycle mode. Check on initial load and mode toggle.
```
flibbertigibbet: :space_invader:

src/app/scripts/cac/control/cac-control-directions-list.js
```diff
@@ -28,7 +28,8 @@ CAC.Control.DirectionsList = (function (_, $, ShareModal, MapTemplates) {
             googlePlusShareButton: '.modal-list-google',
             shareModalButton: '.share-directions',
             stepByStepClass: 'body-step-by-step',
-            sidebarBannerClass: 'body-has-sidebar-banner'
+            hasSidebarBannerClass: 'body-has-sidebar-banner',
```
flibbertigibbet: A nit to pick: I would prefer this keep its name as `sidebarBannerClass`, in keeping with the naming of other selectors. Not clear what the intent of the rename is here.

KlaasH: It's kind of a weird one in that it's a class that gets added to the body div to control some height styles on the sidebar.  I think having `sidebarBannerClass` right next to `activeSidebarBanner` made it strike me that `sidebarBannerClass` sounds like a class that identifies a sidebar banner when in fact it's something completely different.  But I don't feel strongly about it.  I'll change it back.

Issue #768, opened 2017-01-25
Get Directions button on the place detail page not implemented


flibbertigibbet:
Added in #760

flibbertigibbet:
Problematic as the place detail page is currently static.

flibbertigibbet:
As there is no JS loaded on the places detail page, there is neither an origin to reference nor a reference to the JS directions controller we use on the home page to load directions, so this will be non-trivial to implement.

KlaasH:
My comment on PR #770:
Since URL links to directions works, we could provide a link to the directions view with only the destination filled in, like we do if someone clicks a directions button under "Places we love" without having filled in an origin.  The destination is already in the template, so I think `href="/?destination={{ destination.point.y }}%2C{{ destination.point.x }}&destinationText={{ destination.name|urlencode }}"` would do it.

PR #770, opened 2017-01-25
Remove dynamic link from static details page

flibbertigibbet:
See #768.

Remove dynamic "get directions" link added in #760 from static place detail page, as the necessary JS is not present to either reference an origin (any origin set on the home page is lost on navigation to place details) or to reference the directions controller.

Postpone #768 to another phase.

### DIFF
--- a/python/cac_tripplanner/templates/partials/destination-detail.html
+++ b/python/cac_tripplanner/templates/partials/destination-detail.html
@@ -9,10 +9,5 @@
         <p class="info-article-p-link-inline">
             <a href="{{ destination.website_url }}" target="_blank">Visit website</a>
         </p>
-        <div class="info-article-p-actions">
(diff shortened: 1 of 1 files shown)


KlaasH:
URL links to directions work, though, so we could provide a link to the directions view with only the destination filled in, like we do if someone clicks a directions button under "Places we love" without having filled in an origin.  The destination is already in the template, so I think `href="/?destination={{ destination.point.y }}%2C{{ destination.point.x }}&destinationText={{ destination.name|urlencode }}"` would do it.  Or are there nuances I'm overlooking?

flibbertigibbet:
I did not attempt that as I expect building out the URL would introduce complications (we're not doing that anywhere else), and we need to clean up to deploy at this point.

PR #771, opened 2017-01-25
Feature/sml/swipe-itineraries

lederer:
Use the [slick carousel library](http://kenwheeler.github.io/slick/) to make itineraries swipeable on mobile.

Things to note:
- It's still not obvious you need to tap the itinerary to open step-by-step. I'll address this in a follow-up PR. And it should be less of a concern once we do [on-map route cards](https://marvelapp.com/30a0g54/screen/16895433).
- The progress indicator dots overlay the content while swiping. Minor and cosmetic, but I'll plan to revisit the layout in a follow up sweep to fix this.
- Swiping on desktop tends to trigger the click event. Annoying, but will only happen when the viewport is < 480px wide (hopefully never).

### DIFF
--- a/src/app/scripts/cac/control/cac-control-itinerary-list.js
+++ b/src/app/scripts/cac/control/cac-control-itinerary-list.js
@@ -17,7 +17,23 @@ CAC.Control.ItineraryList = (function (_, $, MapTemplates) {
             alert: '.alert',
             container: '.directions-list',
             hiddenClass: 'hidden',
+            itineraryList: '.routes-list',
             itineraryItem: '.route-summary'
+        },
+        // Settings for 'slick' carousel for swiping itineraries on mobile
+        slick: {
+            arrows: false,
+            dots: true,
+            infinite: false,
+            mobileFirst: true,
+            variableWidth: true,
+            responsive : [
+                {
+                    // Breakpoint must match 'xxs' in _breakpoints.scss
+                    breakpoint: 480,
+                    settings: 'unslick'
+                }
+            ]
         }
     };
     var options = {};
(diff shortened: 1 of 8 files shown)


src/app/scripts/cac/control/cac-control-itinerary-list.js
```diff
@@ -17,7 +17,22 @@ CAC.Control.ItineraryList = (function (_, $, MapTemplates) {
...
+        // Settings for 'slick' carousel for swiping itineraries on mobile
+        slick: {
+            arrows: false,
+            dots: true,
+            infinite: false,
+            mobileFirst: true,
+            variableWidth: true,
+            responsive : [
+                {
+                    breakpoint: 480,
```
flibbertigibbet: Would be good to note here and in `_breakpoints.scss` that this magic number needs to match, so it doesn't get changed in one place but not the other in future.

src/app/styles/vendors/_slick-theme.scss
```diff
@@ -0,0 +1,194 @@
+@charset "UTF-8";
```
flibbertigibbet: Is this the same as the slick theme included in the bower package? If so, it should be picked up automatically by the gulp build process.

lederer: It's an optional add-on that cleans up the default slick styles. Not the same as `slick.{css, less, scss}` that's included in the [bower package](https://github.com/kenwheeler/slick/blob/master/bower.json).

src/app/scripts/cac/control/cac-control-itinerary-list.js
lederer: Donezo.

PR #774, opened 2017-01-26
Clean up directions form, including:

lederer:
- Clean up layout and styles of labels, including to/within toggle.
- Make isochrone slider narrower, so 15min steps feel better.
- Use extra space from above to make "15 min" -> "15 minutes".
- Stop red error background from occluding white borders in map view.
- Tighten up typeahead hint.
- Reposition and resize typeahead dropdown.

### DIFF
--- a/src/app/styles/components/_directions-form.scss
+++ b/src/app/styles/components/_directions-form.scss
@@ -8,7 +8,7 @@
     }
 
     @include respond-to('xxs') {
-        margin: ($home-section-margin - 10px) auto;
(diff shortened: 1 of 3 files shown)


src/app/styles/components/_directions-form.scss
```diff
@@ -150,7 +142,7 @@
         }
 
         @include respond-to('xxs') {
-            margin-right: 30px;
+            // margin-right: 30px;
```
flibbertigibbet: delete this

PR #782, opened 2017-01-27
Feature/bump package versions

flibbertigibbet:
Mostly minor version bumps, which fixed a provisioning issue with a package version no longer available (forget which one, now). Also drops `ipython` from the list of packages pip-installed on the app server, due to a current issue with `setuptools`, which is an indirect dependency of `ipython`, and resulted in a broken `pip` installation.

Used to provision current production.

### DIFF
--- a/deployment/ansible/group_vars/all
+++ b/deployment/ansible/group_vars/all
@@ -3,7 +3,9 @@ app_username: "vagrant"
 
 packer_version: "0.7.5"
 
-nodejs_version: 4.6.0
+nodejs_version: 4.6.2
+
+virtualenv_version: 15.1.0
 
 otp_router: "default"
 
--- a/deployment/ansible/roles.yml
+++ b/deployment/ansible/roles.yml
@@ -5,7 +5,7 @@
   version: 0.2.6
 
 - src: azavea.opentripplanner
-  version: 1.0.3
+  version: 1.0.4
 
 - src: azavea.nginx
   version: 0.2.2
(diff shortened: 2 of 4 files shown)


flibbertigibbet:
build again

flibbertigibbet:
don't think Jenkins doesn't like it when switching between `develop` and `redesign` for builds; just destroyed the VMs so it can do a fresh build

flibbertigibbet:
Okay, test failure was due to a new linter warning for some old code.

flibbertigibbet:
build again

deployment/ansible/group_vars/all
```diff
@@ -3,7 +3,9 @@ app_username: "vagrant"
 
 packer_version: "0.7.5"
 
-nodejs_version: 4.6.0
+nodejs_version: 4.6.2
+
+virtualenv_version: 15.1.0
```
KlaasH: Is this useful, or was it an attempt to fix the bug that was ultimately fixed by removing ipython?  Not that it hurts anything, but might as well not add overrides if they're not accomplishing anything.

flibbertigibbet: I'm not sure now if that one was necessary. In any case, the default version in upstream is in need of an update. How about I create an issue there, and an issue here to remove this with role update?

flibbertigibbet: Oh, now I remember. Ran into: azavea/ansible-virtualenv#1

So not totally certain if this version bump was necessary, but I think it did move me past that to a different error.

PR #789, opened 2017-01-30
Feature/sml/ui tweaks

lederer:
Couple of QA fixes…
- Fixed and normalized map marker popup styling, which had regressed
- Hide itineraries list on mobile while loading. This gets rid of an ugly Chrome Android state that would be pretty common. Map view with keyboard open shows a big blank area.

bs_realdroid_mobile_samsung galaxy s7-6 0-1440x2560


lederer:
Doh. Will do.

flibbertigibbet:
With new image sizes:

image

PR #793, opened 2017-01-30
Guard against map control not loaded yet

### LINKED ISSUE #792: Hover interactivity console error
When first loading directions, get console error: `Cannot read property 'highlightDestination' of undefined`.

image

Appears to be referencing `mapControl.isochroneControl` before it is defined.

flibbertigibbet:
Fixes #792.

Guard against map control still loading when first navigating to directions view on mobile (where map may not have loaded yet).

In hover activity handlers, so fine to just do nothing.


flibbertigibbet:
Going to go ahead and merge this little bugfix for staging deploy

PR #803, opened 2017-01-31
Use templated link for directions from destination details

### LINKED ISSUE #801: Directions link on learn details page
Revert #770 and implement suggested hard link with template variables.

flibbertigibbet:
Closes #801.


flibbertigibbet:
Feature wanted, and suggestion on #770 works just fine.

PR #807, opened 2017-01-31
Modify image sizes for redesign

### LINKED ISSUE #795: Put new image dimensions in admin
- [x] Article narrow: 310x218
- [ ] Article wide: 680x200
- [ ] Destination narrow: 310x155 
- [ ] Destination wide: 680x400

Once those are up, we should ask client to generate new imagery.

flibbertigibbet:
Closes #795.

### DIFF
--- a/python/cac_tripplanner/destinations/models.py
+++ b/python/cac_tripplanner/destinations/models.py
@@ -37,9 +37,9 @@ class Destination(models.Model):
                                           'but may be overridden manually if incorrect.'))
     point = models.PointField()
     image = models.ImageField(upload_to=generate_filename, null=True,
-                              help_text='The full-size image. Will be displayed at 400x400.')
+                              help_text='The full-size image. Will be displayed at 310x155.')
     wide_image = models.ImageField(upload_to=generate_filename, null=True,
-                                   help_text='The half-height image. Will be displayed at 400x200.')
+                                   help_text='The half-height image. Will be displayed at 680x400.')
     published = models.BooleanField(default=False)
     priority = models.IntegerField(default=9999, null=False)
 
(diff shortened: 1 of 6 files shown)


flibbertigibbet:
@lederer if you could please share updated default destination images for `default_media`, I'll add those in.

python/cac_tripplanner/destinations/models.py
```diff
@@ -37,9 +37,9 @@ class Meta:
                                           'but may be overridden manually if incorrect.'))
     point = models.PointField()
     image = models.ImageField(upload_to=generate_filename, null=True,
-                              help_text='The full-size image. Will be displayed at 400x400.')
+                              help_text='The full-size image. Will be displayed at 310x155.')
     wide_image = models.ImageField(upload_to=generate_filename, null=True,
-                                   help_text='The half-height image. Will be displayed at 400x200.')
+                                   help_text='The half-height image. Will be displayed at 680x400.')
```
lederer: The "Full-size" / "Half-height" and "Narrow" / "Wide" terms no longer really apply to the new sizes.

Would it be possible to change the display names of the image fields in the admin UI on both Articles and Destinations to simply: "Large image" and "Small image"?

Regardless of whether we can do that, let's change the help text, for both articles and destinations, to:

"The large image.  Will be displayed at {w}x{h}."
"The small image. Will be displayed at {w}x{h}."

lederer:
*Assuming the validator only checks the size when the image is uploaded in the admin UI, and not when the image is served in the product*, then all existing `default_media` images should work fine for now, even though they are the old sizes.

The existing small ones are the same aspect ratio required by the new design, and are larger, so should automatically downsize nicely. 

The existing large ones have a different aspect ratio from that required by the new design, but CSS's `background-size: cover` algorithm makes that a non-issue. They are smaller than required but look good enough for now when upsized by the browser. This will become irrelevant once the client uploads all new images based on their full-size source images.

flibbertigibbet:
Although the existing media may still be added successfully during provision, they could not be edited later in the admin interface, because of the changed aspect ratio for the large images. Not requesting for client use (production database does not use the default media), but to preserve development environment.

flibbertigibbet:
merging for staging deploy

PR #808, opened 2017-02-01
Use setValue in geolocate

KlaasH:
The typeaheads have a function that gets called on `blur` and resets the
value if there's a mismatch between what the field says and the last
selected value. The geolocation callback was setting the field's value
but not using the special `setValue` method that keeps the internal
lastSelectedValue in sync.

This produced no symptoms until PR #787, which caused a blur event to
happen when it wasn't before, which resulted in the value set by the
geolocation being compared to the blank original value, triggering the
reset.

This makes the geolocation callback use `setValue`.

Note: I don't know if there might be a more proper way to get at the typahead object inside the geolocation callback.  It's so many levels deep that assigning it to a variable in the `createTypeahead`scope seemed simpler than whatever would be required to make it available as `this` in the scope where we need it.


flibbertigibbet:
merging for staging deploy

PR #809, opened 2017-02-01
Feature/sml/ui tweaks

lederer:
This PR updates the Indego font icon with a version that renders across browsers. And increases the `min-height` of the Places We Love section of the home page so the article card below it doesn't jump up too high when the places are refreshed (eg, when the origin changes).


flibbertigibbet:
merging for staging deploy

Issue #811, opened 2017-02-01
Override default message for out-of-bounds error


flibbertigibbet:
If non-navigable origin or destination set, look for default message returned by OTP:
```
"Trip is not possible. your start or end point might not be safely accessible (for instance, you might be starting on a residential street connected only to a highway)."
```
and override to replace with "Make sure the origin and destination are accessible addresses within the Greater Philadelphia area."

PR #812, opened 2017-02-01
Override default out-of-bounds message

flibbertigibbet:
Connects #811.

image


flibbertigibbet:
Trivial; merging for staging deploy

PR #813, opened 2017-02-02
Resized default_media imagery

lederer:
See #807


lederer:
Ah, I see what happened. I swapped their directories. I put the small ones in `half-square` and the larges ones in `square`, which makes more sense at first blush based on the directory names and images sizes. But in reality they need to be swapped. Swap forthcoming…

lederer:
Relatedly the help texts (in that screengrab above) are swapped. Eg, "The **large image**. Will be displayed at 310x155" should be "The small image…"

flibbertigibbet:
Could you update help text on this PR?

PR #818, opened 2017-02-03
Feature/sml/route colors and misc

lederer:
Remove close button from share modal. Was getting in the way of all the share options, and not necessary, with click-outside working.

Fix some padding and size issues for the directions "back" and "share" buttons.

Changed walk color to GoPhlGo blue and made all transit modes GoPhlGo yellow. Rationale:
- On-brand colors aren’t necessary, but good to use if we can.
- SEPTA's primary lines don’t use yellow AFAICT, so in theory our yellow transit paths won’t imply a certain line.
- Purple (bike) is a SEPTA color for Norristown HSL, but that’s not in the downtown area and has a very specific footprint, so hopefully won’t be confusing.
- Blue (walk) is the MFL, but that has a very specific route, while walking can happen *anywhere*. Further, blue is our main UI color, so hopefully the semantic gravity [TM] of matching the UI will outweigh that of the MFL line. They’re also quite different blues.

### DIFF
--- a/python/cac_tripplanner/templates/partials/modals.html
+++ b/python/cac_tripplanner/templates/partials/modals.html
@@ -106,9 +106,6 @@
             <p>Here's a link to these directions:</p>
             <p><a id="modalDirectionsLink" href=""></a></p>
         </div>
-        <footer class="modal-footer">
-            <button name="close-modal" class="btn-close-modal">Close</button>
-        </footer>
     </div>
 
 </div>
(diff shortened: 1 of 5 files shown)


flibbertigibbet:
Close button is still on trip options modal.

lederer:
The inconsistency is acceptable in this case. 

Options is a complex component, with consequences for the state of the app. Having an explicit close button can help the user feel like they're actively completing an action of consequence.

A share sheet OTOH is a fleeting experience with no impact on the app. Having a close button on it would be preferable, but not necessary.

Issue #824, opened 2017-02-10
Admin geocoder not setting location


flibbertigibbet:
The admin interface should geocode location and update point on map when an address is entered.

Currently it throws a JavaScript error and does not update the map marker; likely due to geocoder updates.

PR #825, opened 2017-02-10
Fix admin geocoder

flibbertigibbet:
Update for changed result object from geocoder.
Fixes console error/not updating map on address change in admin editor.
Connects #824.


flibbertigibbet:
A small fix; will go ahead and merge for staging update.

PR #826, opened 2017-02-13
Shorter hero image on destination detail page

lederer:
Tweak to make place detail image and title fit mobile screens better.


lederer:
Fixed that. Destination hero image height now resizes to honor aspect ratio in mobile. 

This only applies to destinations, not articles. Article images are short enough that it would be better to maintain the current cropping on mobile.

flibbertigibbet:
Looks nice! :+1:

PR #865, opened 2017-07-28
Add app manifest and service worker for a PWA

### LINKED ISSUE #845: Define web app manifest
To ease users adding web site to their phone, define web app manifest. Process [described here](https://developers.google.com/web/fundamentals/engage-and-retain/app-install-banners/).

flibbertigibbet:
## Overview

Adds an app manifest and service worker to support 'add to home screen', making the site behave similar to a native app. The service worker caches some assets offline. Also added a no-cache header to nginx. (It's a misnomer. It allows caching, but says to always check for a new file version before using the cached one.)

#### Testing on BrowserStack with a Nexus 6 and Chrome:

On first page load in browser, an 'add to homescreen' modal pops up at the bottom of the screen.
cac_nexus_browserstack

Home screen after adding:
cac_nexus_homescreen_browserstack

On first launch, it take a second to load, so Chrome shows a splash screen using the background color:
cac_nexus_loading_screen

After it's loaded, there's no URL bar. The app is fullscreen.
cac_nexus6_loaded_fullscreen

The 'add to desktop' prompt on Chrome/Windows (can override a couple of options).
cac_add_to_desktop_prompt_chrome_win

When in non-debug mode, the service worker will explicitly cache JS files as well as the home page and CSS. (Not caching JS in dev mode as there are a lot of files, and we don't need to.)
cac_production_build_service_worker_cache

What happens when loaded without a network connection, in dev mode, as a demo of service worker caching actually functioning. The service worker's explicitly cached files are still available, and something loads. As this app won't work offline anyways, probably okay that this looks broken.
cac_cached_pwa_no_network

### Notes

Couldn't test on Mac or iOS, as BrowserStack doesn't support local domains those platforms.

Information on the manifest and service workers can be seen under the Application tab in Chrome dev tools.

Added all of the square logo assets we have to the manifest, but there are better sizes to use. There's some documentation on those [here](https://developers.google.com/web/fundamentals/engage-and-retain/web-app-manifest/?utm_source=devtools). The lighthouse PWA audit tool (under the Chrome dev tools 'Audits' tab) complains that the app needs some icons at least 512px. Over [here](https://developers.google.com/web/updates/2014/11/Support-for-theme-color-in-Chrome-39-for-Android) they recommend 192px for Chrome/Android.

The colors in the manifest and theme meta tags are the blue and dark blue colors in the app.

A couple quirks in the Chrome dev tools: the 'Add to homescreen' link by the manifest info doesn't do anything, at least on my platform. The non-devtools methods do work, though (described below). Another is that it seemed often necessary to close and reopen dev tools for the service worker file list to update, under Application -> Cache Storage.

## Testing Instructions

 * Picking up the nginx configuration change will require reprovisioning the app server:
   `vagrant provision app`
 * Rebuild JS assets in the app vagrant VM: `cd /opt/app/src && npm run gulp-development`
 * Visit the home page at http://localhost:8024 (be sure site cache is clear)
 * There should be no console errors
 * Refresh page to trigger the service worker to fetch; should still be no console errors
 * 'Size' under Chrome dev tools network tab should show '(from ServiceWorker)' some assets
 * Add to desktop on Chrome under 'more tools' in the upper right three grey dots menu
 * Should be a GoPhillyGo icon added to desktop that's capable of opening site in new window

Closes #865

### DIFF
--- a/python/cac_tripplanner/templates/manifest.json
+++ b/python/cac_tripplanner/templates/manifest.json
@@ -1,0 +1,79 @@
+{
+  "short_name": "GoPhillyGo",
+  "name": "GoPhillyGo Trip Planner",
+  "icons": [
+    {
+      "src": "/static/images/apple-touch-icon-57x57.png",
+      "type": "image/png",
+      "sizes": "57x57"
+    },
+    {
+      "src": "/static/images/apple-touch-icon-60x60.png",
+      "type": "image/png",
+      "sizes": "60x60"
+    },
+    {
+      "src": "/static/images/apple-touch-icon-57x57.png",
+      "type": "image/png",
+      "sizes": "57x57"
+    },
+    {
+      "src": "/static/images/apple-touch-icon-72x72.png",
+      "type": "image/png",
+      "sizes": "72x72"
+    },
+    {
+      "src": "/static/images/apple-touch-icon-76x76.png",
+      "type": "image/png",
+      "sizes": "76x76"
+    },
+    {
+      "src": "/static/images/apple-touch-icon-114x114.png",
+      "type": "image/png",
+      "sizes": "114x114"
+    },
+    {
+      "src": "/static/images/apple-touch-icon-120x120.png",
+      "type": "image/png",
+      "sizes": "120x120"
+    },
+    {
+      "src": "/static/images/apple-touch-icon-144x144.png",
+      "type": "image/png",
+      "sizes": "144x144"
+    },
+    {
+      "src": "/static/images/apple-touch-icon-152x152.png",
+      "type": "image/png",
+      "sizes": "152x152"
+    },
+    {
+      "src": "/static/images/logo-blue.png",
+      "type": "image/png",
+      "sizes": "50x50"
+    },
+    {
+      "src": "/static/images/logo-blue@2x.png",
+      "type": "image/png",
+      "sizes": "100x100"
+    },
+    {
+      "src": "/static/images/logo.png",
+      "type": "image/png",
+      "sizes": "180x180"
+    },
+    {
+      "src": "/static/images/logo@2x.png",
+      "type": "image/png",
+      "sizes": "360x360"
+    },
+    {
+      "src": "/static/images/favicon.png",
+      "type": "image/png",
+      "sizes": "32x32"
+    }
+
+  ],
+  "start_url": "?launcher=true",
+  "display": "standalone"
+}
--- a/python/cac_tripplanner/templates/manifest.json
+++ b/python/cac_tripplanner/templates/manifest.json
@@ -1,6 +1,6 @@
 {
   "short_name": "GoPhillyGo",
-  "name": "GoPhillyGo Trip Planner",
+  "name": "GoPhillyGo",
   "icons": [
     {
       "src": "/static/images/apple-touch-icon-57x57.png",
--- a/python/cac_tripplanner/templates/manifest.json
+++ b/python/cac_tripplanner/templates/manifest.json
@@ -74,6 +74,8 @@
     }
 
   ],
-  "start_url": "?launcher=true",
-  "display": "standalone"
+  "start_url": "/?launcher=true",
+  "display": "standalone",
+  "background_color": "#245280",
+  "theme_color": "#2e68a3"
 }
--- a/python/cac_tripplanner/templates/partials/ios-weblinks.html
+++ b/python/cac_tripplanner/templates/partials/ios-weblinks.html
@@ -1,0 +1,5 @@
+<script type="text/javascript" charset="utf-8">
+// Keep iOS in standalone mode when following links
+// https://gist.github.com/irae/1042167
+(function(a,b,c){if(c in b&&b[c]){var d,e=a.location,f=/^(a|html)$/i;a.addEventListener("click",function(a){d=a.target;while(!f.test(d.nodeName))d=d.parentNode;"href"in d&&(chref=d.href).replace(e.href,"").indexOf("#")&&(!/^[a-z\+\.\-]+:/i.test(chref)||chref.indexOf(e.protocol+"//"+e.host)===0)&&(a.preventDefault(),e.href=d.href)},!1)}})(document,window.navigator,"standalone");
+</script>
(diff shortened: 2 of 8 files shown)


lederer:
### iOS
iOS doesn't support PWAs. I tried it on the iOS simulator just in case but none of the work in this PR had any impact. 

I was able to manually use Mobile Safari's "Add to Home Screen" action, but that works with any website anyway. It adds the icon to the home screen, but it merely acts as a bookmark. The site opens as a tab in Safari, with the usual browser chrome, rather than as a fullscreen web app.

[Here’s Apple's howto](https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html) for doing this the iOS way. IIUC, we can implement it thusly *and* as an android-friendly PWA at the same time and they won't conflict.

### Assets
I can create assets at all the appropriate sizes. @flibbertigibbet Should I get those ready and commit them to this PR? Or do it as a follow-up PR? I suppose it doesn't matter, but figured I'd ask.

flibbertigibbet:
Okay. Knew Safari wouldn't support PWAs, but wasn't aware they had their own add-to-homescreen meta tags. Will add those. Just wanted to be sure this didn't break anything on Safari/iOS.

A follow-up PR for additional assets might be easier, but a PR to this PR would also be welcome.

python/cac_tripplanner/templates/manifest.json
```diff
@@ -0,0 +1,81 @@
...
+      "type": "image/png",
+      "sizes": "114x114"
+    },
+    {
+      "src": "/static/images/apple-touch-icon-120x120.png",
+      "type": "image/png",
+      "sizes": "120x120"
+    },
+    {
+      "src": "/static/images/apple-touch-icon-144x144.png",
```
lederer: I think you can remove all the icon entries from the manifest except this one (144x144). I'll add one or two more (192, maybe 96) in a follow-up PR.

flibbertigibbet: It does look like the Nexus (screenshot above) picked one of the white background logos. I'll let you tinker with this list in your PR.

lederer: SGTM

python/cac_tripplanner/templates/manifest.json
```diff
@@ -0,0 +1,79 @@
+{
+  "short_name": "GoPhillyGo",
+  "name": "GoPhillyGo Trip Planner",
```
lederer: Here's another instance of "Trip Planner" to be removed. This is what ends up in the PWA splash.

lederer:
Copy edits look good. Thanks.

---

The iOS "Add to Home Screen" functionality is now operational on the simulator. But with a wrinkle:

Navigating from the map page to a different page in the app (eg, clicking "Learn" in the top nav, or "More info" on any place card) causes that page to open in Mobile Safari instead of within the standalone "app". The experience is pretty jarring. Makes the web app not feel like a web app.

[The solution](https://www.google.com/search?q=ios+web+applications+link+opens+in+safari) amounts to hijacking clicks on links to in-app pages. Roughly:
```
event.preventDefault();
location.href = $( event.target ).attr( "href" );
```

@flibbertigibbet Would fixing this have other consequences for the app?

flibbertigibbet:
That solution does seem pretty hacky, but I think it would be okay as long as it only needs to be run on the static Django-served pages. Otherwise it might tangle with our JavaScript router. @KlaasH have any opinion on this?

flibbertigibbet:
Link above is for a web search. Discussion on [this gist](https://gist.github.com/kylebarrow/1042026) seems particularly relevant.

flibbertigibbet:
Pushed an update that loads a JS snippet on linked pages to force iOS standalone mode.

Could you please test on emulator or device @lederer?

And @KlaasH please code review.

python/cac_tripplanner/templates/partials/ios-weblinks.html
```diff
@@ -0,0 +1,5 @@
+// Keep iOS in standalone mode when following links
```
lederer: These comments should be inside the `<script>` tags.

KlaasH: Yeah, I was also wondering what the risk or cost of putting `ios-weblinks.html` in `base.html` would be.  Having it in 5 or 6 places is kind of a drag, and seems to add risk of a future situation in which it becomes necessary on a new template but gets forgotten (and it wouldn't be that easy/likely for us to notice...).  Is the concern that it could have unforeseen negative effects on some links so we don't want to include it anywhere we don't definitely need it?

KlaasH: Oops, that's a reply to the line comment.  I wish it would let you reply to the overall review comment...

flibbertigibbet:
The script is in a partial template and not in the base template in order to intentionally exclude it from the main page (the one-page app), where it shouldn't be needed as it is the home page first loaded, and where the script might interfere with the JS-based router.

Are you requesting this on the home page to support links back from the other templated pages? Maybe those also break out of standalone mode, but I don't have an environment to test.

flibbertigibbet:
Okay, I'll try just putting it in the base template and we can see if it causes any weird behavior.

lederer:
It doesn't work on the home page (unless I include the partial in `home.html`).

flibbertigibbet:
I take it you mean links *out* from the homepage wouldn't work without it in the base template?

Anyways, pushed change to put it in home template.

lederer:
>I take it you mean links out from the homepage wouldn't work without it in the base template?

That's right. While on the home page, going to a different page within the site — eg clicking "Learn" in the top nav or clicking "More info" on a place card — would open that page in Mobile Safari rather than within the standalone "app".

Taking a look at the changes now…

lederer:
Seems to be working fine now. I'm able to go to all the different pages in the site, while remaining in the standalone "app". And getting directions works just like always.

@flibbertigibbet Any other specific use cases I should try out to test the routing?

flibbertigibbet:
No, I think that should do it. Thanks for testing it out @lederer!

PR #867, opened 2017-07-31
Add current location marker on mobile

### LINKED ISSUE #848: Show current location on mobile
Detect mobile users (as users with fine (not coarse) geolocation results?) and display current location icon for those users.

flibbertigibbet:
Use breakpoint for mobile detection, as HTML5 does not support detecting if GPS geolocation is supported.
Use Leaflet `locate` to watch user location.

Closes #848.

### Demo

image

### Notes

Since HTML5 feature support detection does not allow for finding out if GPS is present or not, instead use screen width for mobile device detection.

## Testing Instructions

 * Rebuild JS in app VM with `cd /opt/app/src && npm run gulp-development`
 * Resize browser to width < 992 px
 * Visit home page at http://localhost:8024/, location marker should *not* show
 * Go to map page for directions or explore page; location marker should now show
 * Should be able to navigate back and forth between home and map pages with expected behavior

### DIFF
--- a/src/app/scripts/cac/map/cac-map-control.js
+++ b/src/app/scripts/cac/map/cac-map-control.js
@@ -5,6 +5,7 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
         id: 'map',
         selectors: {
             id: '#map',
+            isMobile: false,
             leafletMinimizer: '.leaflet-minimize',
             leafletLayerList: '.leaflet-control-layers-list',
             leafletLayerControl: '.leaflet-control-layers',
@@ -15,6 +16,7 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
     };
 
     var map = null;
+    var currentLocationMarker = null;
     var geocodeMarker = null;
     var directionsMarkers = {
         origin: null,
@@ -123,7 +125,11 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
         tabControl = this.options.tabControl;
 
         // put zoom control on top right
-        zoomControl = new cartodb.L.Control.Zoom({ position: 'topright', zoomInText: '<i class="icon-plus"></i>', zoomOutText: '<i class="icon-minus"></i>' });
+        zoomControl = new cartodb.L.Control.Zoom({
+            position: 'topright',
+            zoomInText: '<i class="icon-plus"></i>',
+            zoomOutText: '<i class="icon-minus"></i>'
+        });
 
         initializeBasemaps();
 
@@ -135,6 +141,11 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
                                         function(event, place) {
                                             events.trigger(eventNames.destinationPopupClick, place);
                                         });
+
+        if (this.options.isMobile) {
+            showCurrentLocation();
+        }
+
         mapLoaded = true;
     }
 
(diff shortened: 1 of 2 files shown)


flibbertigibbet:
@lederer might want to update the icon to something other than the default blue circle.

src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -350,6 +361,9 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
             zoomControl.addTo(map);
             initializeOverlays();
             initializeLayerControl.apply(this, null);
+            if (currentLocationMarker) {
+                currentLocationMarker.addTo(map);
+            }
```
KlaasH: This doesn't work here, because `currentLocationMarker` gets initialized in the `on('locationfound', ...`, which might (in my experience so far, usually does) fire after this runs, so the marker is still null when this checks it and the layer never shows up.  I think it would make sense to remove this from here and put the `currentLocationMarker.addTo(map);` in the `if (!currentLocationMarker)` block of `showCurrentLocation`.

Actually, it looks like my proposed solution causes the marker to show behind the form on the home screen.  
image

Or that could be an independent problem that was masked by the failure of the marker to show up at all.

src/app/scripts/cac/pages/cac-pages-home.js
```diff
@@ -59,7 +59,8 @@ CAC.Pages.Home = (function ($, ModeOptions,  MapControl, TripOptions, SearchPara
         });
 
         mapControl = new MapControl({
-            tabControl: tabControl
+            tabControl: tabControl,
+            isMobile: $(window).width() < MD_UP_BREAKPOINT
```
KlaasH: Since this comparison is being done twice now, it could be moved to a component-level variable set during initialization.  Then again, it's only twice.

src/app/scripts/cac/map/cac-map-control.js
```diff
@@ -123,7 +125,11 @@ CAC.Map.Control = (function ($, Handlebars, cartodb, L, turf, _) {
         tabControl = this.options.tabControl;
 
         // put zoom control on top right
-        zoomControl = new cartodb.L.Control.Zoom({ position: 'topright', zoomInText: '<i class="icon-plus"></i>', zoomOutText: '<i class="icon-minus"></i>' });
+        zoomControl = new cartodb.L.Control.Zoom({
+            position: 'topright',
+            zoomInText: '<i class="icon-plus"></i>',
+            zoomOutText: '<i class="icon-minus"></i>'
+        });
```
KlaasH: :+1:

src/app/scripts/cac/pages/cac-pages-home.js
flibbertigibbet: The comparison needs to be done repeatedly in case the view resizes. For the other usage of this breakpoint at least, it's important to check for resize.

src/app/scripts/cac/map/cac-map-control.js
flibbertigibbet: In narrow screen widths, the map doesn't initialize on the home page, but only on the map view. In that case the location marker may be initialized before the map, so the check here is needed. I'll see if I can reproduce the marker not appearing.

flibbertigibbet: Do you get geolocation errors in the console when using the home page `from` geolocator button? It may be the location simply wasn't found.

KlaasH: Nothing in the console.  I'm forcing the location using Chrome developer tools, and my debugger inside the `map.on('locationfound', function(event) {` function does trip, but no dot.

flibbertigibbet:
Pushed a change to add the marker when location is found. The test for `componentsLoaded` will ensure it doesn't show on the home page view. Adding a layer that's already on the map doesn't seem to have any negative effects.

flibbertigibbet:
build again

PR #868, opened 2017-07-31
Feature/kak/display travel times with minutes#836

### LINKED ISSUE #836: Always display travel times with minutes
When humanizing summary information, always display minutes in addition to hours, particularly for differentiation of trip options with the same rounded travel hours.

flibbertigibbet:
## Overview

Display travel durations greater than an hour with minutes. Default moment `humanize` behavior is to round to the largest unit.

### Demo

image

### Notes

Related feature request is at moment/moment#348.

Can cause text wrap in the trip summary:
image

## Testing Instructions

 * `cd /opt/app/src && npm run gulp-development`
 * Set origin on home page at http://localhost:8024
 * Wait for the place cards to populate the travel times
 * Travel times over an hour should also display minutes
 * Minutes should no longer round up to hours over 45 minutes
 * Check times on map page itinerary list and detail as well

Closes #836

### DIFF
--- a/src/app/scripts/cac/pages/cac-pages-directions.js
+++ b/src/app/scripts/cac/pages/cac-pages-directions.js
@@ -1,4 +1,4 @@
-CAC.Pages.Directions = (function ($, _, DirectionsList, Itinerary, Settings) {
+CAC.Pages.Directions = (function ($, _, DirectionsList, Itinerary, Settings, Utils) {
     'use strict';
 
     var center = [39.95, -75.1667];
@@ -24,11 +24,7 @@ CAC.Pages.Directions = (function ($, _, DirectionsList, Itinerary, Settings) {
 
     Directions.prototype.initialize = function () {
 
-        moment.updateLocale('en', {
-            relativeTime : {
-                mm: '%d min',
-            }
-        });
+        Utils.initializeMoment();
 
         // Note: date/time does not get passed and so always defaults to departing now
 
@@ -162,4 +158,4 @@ CAC.Pages.Directions = (function ($, _, DirectionsList, Itinerary, Settings) {
 
     return Directions;
 
-})(jQuery, _, CAC.Control.DirectionsList, CAC.Routing.Itinerary, CAC.Settings);
+})(jQuery, _, CAC.Control.DirectionsList, CAC.Routing.Itinerary, CAC.Settings, CAC.Utils);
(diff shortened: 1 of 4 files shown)


flibbertigibbet:
That sorts out the text wrapping:
image

lederer:
This is great. 

Couple tweaks needed for the hours + minutes case:
1. Remove the space between the number and the unit.
2. Use "m" for the minutes unit.

So, this: "2h 47m"
Instead of: "2 h 47 min"

Then I think we're golden.

flibbertigibbet:
image

lederer:
Awesome. Thanks for working through the details with me. Go team.

lederer:
>The comment says it should be "1 hour".

Nice catch @KlaasH. 🔍 

I do think we should switch to "1 hour", since we're using actual numerals for all other cases (at least within 24h, which is all that matters).

flibbertigibbet:
Humanize will still return some other non-numeric responses, like "a few"
image

flibbertigibbet:
If there are both hours and minutes, this will not use humanize and instead format directly. For exact hours without a minute component or durations less than an hour or longer than a day, humanize formatting is applied. Currently the only override to its formatting via `moment.updateLocale` is @lederer's modification to plural minutes (one minute is still "a minute"). See [docs](https://momentjs.com/docs/#/customization/relative-time/).

lederer:
Humanize is fine for > 1 day since it should be extremely rare / not worth the hassle of perfecting the formatting.

For ≤ 1 hour, let's skip humanize and just do "N units", where "units" is {hour, min, sec}.

eg:
- 1 hour
- 59 min
- 1 min
- 59 sec

All the humanize alternatives ("a few seconds", "a few minutes", "an hour") take up more precious space, and look out of place when rendered near non-humanized formats. Eg, on the home page with an origin set and all the place cards show their travel times.

flibbertigibbet:
Okay, locale formatting overrides should work for the rest. So you want 'seconds' to be changed to 'sec' as well then @lederer?

lederer:
>  So you want 'seconds' to be changed to 'sec' as well then @lederer?

Yup. Thanks!

flibbertigibbet:
One hour:
image

Times in seconds generally won't display, due to the changes with per-mode formatting not showing short mode travel, and OpenTripPlanner refusing to route too-short trips, which seems to cover anything under two minutes' walk. But, there's formatting for those, too.

PR #870, opened 2017-08-03
Fix issue when all legs shorter than minimum

### LINKED ISSUE #869: Cannot read property 'toLowerCase' of undefined
For some very short trips, the itinerary list fails to populate with a console error. Issue appears to be that there is no `modeText`.

image

flibbertigibbet:
## Overview

When building per-mode summaries, if all modes have distances shorter than the minimum,
reuturn a summary for the longest mode used.

Error was result of trying to display a single-mode summary, but having no modes to show.

### Demo

Example trip that would result in a blank itinerary list and a console error before:
image

## Testing Instructions

 * Try planning very short trips
 * Should always get itinerary list (or handled error alert that origin/destination are too close)
 * Should get no console errors

Fixes #869.


flibbertigibbet:
build again

flibbertigibbet:
Right, when you ask for a super-short trip when transit is enabled, it will give the obvious just-walk-there option, then also return seemingly whatever the least ridiculous option is that involves transit.
:woman_shrugging:

PR #875, opened 2017-08-07
Use just street address on place cards

### LINKED ISSUE #872: Remove city etc from origin label on place cards
When the user has an origin set, the place cards show the travel duration, eg:
image

We should use just the street address as the origin text there. They know the city (assuming they chose the correct address in the geocoder), and so it's just noise in this context.

flibbertigibbet:
## Overview

When adding travel times from origin to place cards on home page, show just the street address.
Closes #872.

### Demo

image

## Testing Instructions

 * Set an origin address on the home page form and wait for the place card travel times to populate
 * The "x min from y" text on the place cards should only show the origin street address


flibbertigibbet:
Good find. Just pushed a change to show featured destination name, if present.

PR #888, opened 2017-09-06
Do not use service worker to cache admin pages

### LINKED ISSUE #887: Error saving new destinations in admin panel
Nick and Erika are getting the error below when creating new destinations. Both are using Chrome and Windows. 

Nick saw the issue last week when creating Parks on Tap 2. The issue seemed to resolve itself when he cleared his cache. Erika is now seeing the same error when attempting to create a Parking Day destination (on 9/5); she presses save, she gets to an error screen. When she presses back in the web browser, nothing is saved and she has to start over, with nothing ever getting saved. She has cleared her cache and logged out/logged in.

gophillygo_bug

flibbertigibbet:
## Overview

Service worker requests will not include credentials unless specifically requested in the fetch.
See `credentials` parameter discussion [here](https://developers.google.com/web/fundamentals/getting-started/primers/service-workers).

However, including credentials for all fetch requests will cause some requests to fail, such as for third-party fonts. Instead exclude `admin` page requests (to the Django CMS app) from the cache, which require the CSRF token for form posts.

## Notes

Posting the failed request as a curl command copied from Chrome results in a 403 error regarding the CSRF token missing, although the posted form contains the expected CSRF hidden field with its value set. This led to finding that service worker `fetch` requests do not include cookies by default.

## Testing
 * can first reproduce issue on current `develop` branch by attempting to save anything in admin panel
 * in Chrome network tab, the admin form page loaded before failed save should show `(from ServiceWorker)` under `size` column
 * testing fix: now pull this branch
 * restart app: `vagrant ssh app -c "sudo service restart cac-tripplanner-app"`
 * clear browser cache under Application tab (remove service worker)
 * visit app home page to load new service worker and let it cache things
 * saving changes in admin panel should work
 * page requests to admin panel should no longer cache the page HTML via service worker

Fixes #887.

### DIFF
--- a/python/cac_tripplanner/templates/service-worker.js
+++ b/python/cac_tripplanner/templates/service-worker.js
@@ -27,14 +27,10 @@ self.addEventListener('fetch', function(event) {
             return response;
         } else {
             return fetch(event.request).then(function (response) {
-                // Cache fetched request if it is on this domain,
-                // does not involve a routing request (has an origin),
-                // and does not require authentication credentials (cookies are not passed).
+                // Only cache static and media assets on this domain
                 var url = event.request.url;
                 if (url.startsWith(location.origin) &&
-                    !url.includes('origin=') &&
-                    !url.includes('/admin') &&
-                    event.request.method === 'GET') {
+                    (url.includes('/static') || url.includes('/media'))) {
 
                     var responseClone = response.clone();
                     caches.open(CACHE_NAME).then(function (cache) {
--- a/python/cac_tripplanner/templates/service-worker.js
+++ b/python/cac_tripplanner/templates/service-worker.js
@@ -1,7 +1,7 @@
 // Service Worker to support functioning as a PWA
 // https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers
 
-var CACHE_NAME = 'cac_tripplanner_v1';
+var CACHE_NAME = 'cac_tripplanner_v2';
 
 var cacheFiles = {{ cache_files | safe }};
 
@@ -28,7 +28,7 @@ self.addEventListener('fetch', function(event) {
         } else {
             return fetch(event.request).then(function (response) {
                 // cache fetched request if it is on this domain
-                if (request.url.startsWith(location.href)) {
+                if (!location.href.includes('/admin') && request.url.startsWith(location.href)) {
                     var responseClone = response.clone();
                     caches.open(CACHE_NAME).then(function (cache) {
                         cache.put(event.request, responseClone);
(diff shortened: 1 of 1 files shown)


python/cac_tripplanner/templates/service-worker.js
```diff
@@ -28,7 +28,7 @@ self.addEventListener('fetch', function(event) {
         } else {
             return fetch(event.request).then(function (response) {
                 // cache fetched request if it is on this domain
-                if (request.url.startsWith(location.href)) {
+                if (!location.href.includes('/admin') && request.url.startsWith(location.href)) {
```
KlaasH: This is a troublesome one.

I've finally gotten it to stop doing the crash.  I think the fact that I had earlier installed the site as an app meant that no matter how much I tried to get rid of the service worker (by either "Clear site data" under Application or in chrome://serviceworker-internals/), it didn't delete the one saved with the app, and just kept using it.  To make matters more confusing, when I clicked on `service-worker.js` in the Application tab
image
it didn't open in Sources, it opened a new tab with http://localhost:8024/service-worker.js, which of course showed the version being served, which was not the version actually being used by my browser.

Uninstalling GoPhillyGo from my Apps and restarting my browser finally cleared that stale service worker, which means both that I can actually see changes and that clicking it in the Application tab opens it in Sources where I can drop a debugger.

I also found the actual storage location of the `cac_tripplanner_v2` cache (under `~/.config/google-chrome/Default/Service Worker/CacheStorage`) and deleted it, but that might have been a red herring.

Anyway, now that I can get the new service worker and don't have stale stuff in my cache, I can confirm that the crashes go away.

However, I think there are some problems with this check in both its current and former forms.  

When I drop a debugger at this line and run the test in the console, I get "Uncaught ReferenceError: request is not defined".  I think that's getting caught by the `catch`, but it means we're not actually putting anything in the cache.  If I change it to `event.request.url`, it still always comes up false because it appears that `location.href`, in the context of the service worker, is always http://localhost:8024/service-worker.js.

So I think the test should be something like `!event.request.url.includes('/admin') && event.request.url.startsWith(location.origin)` to exclude 3rd party and admin stuff but cache everything else on the domain.

Then again, if the existing condition was no good, there shouldn't have been anything in the service worker's cache at all, so how could we have seen an error caused by over-caching?  I may still be missing something.

flibbertigibbet: I suspect the debugger might be untrustworthy in this context, as I've also noticed console logs to not print as expected from the service worker. The URL checks had come from some documentation, though I'll have to dig some to find the reference. Given that the caching had been occurring and now no longer cache admin files, I think that should be enough to go on along with behavior to review.

flibbertigibbet:
Updated to only cache local static media and assets (also fixes `request` reference to that of the event.)

Also fixes service-worker related console errors on sharing to social media, also due to POST requests requiring cookies, as well as admin pages. An intermediate commit made the same fixes in a more complicated way, by excluding non-GET requests. Went with this to simplify things and avoid potential issues with unexpected caching.

flibbertigibbet:
Bump. (Ready for review.)

PR #889, opened 2017-09-06
Feature/kak/update python dependencies

flibbertigibbet:
## Overview

Updates python app dependencies and adds Django logging to file.

Not necessary for anything at the moment. Done as part of debugging for #888.

## Testing Instructions

 * `vagrant provision app`
 * Django app should still function as expected

### DIFF
--- a/python/cac_tripplanner/cac_tripplanner/settings.py
+++ b/python/cac_tripplanner/cac_tripplanner/settings.py
@@ -166,8 +166,7 @@ MEDIA_ROOT = '/media/cac/'
 MEDIA_URL = '/media/'
 
 # LOGGING CONFIGURATION
-LOGGING_CONFIG = None
-logging.config.dictConfig({
+LOGGING = {
     'version': 1,
     'disable_existing_loggers': False,
     'formatters': {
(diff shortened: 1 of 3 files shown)


python/cac_tripplanner/cac_tripplanner/settings.py
```diff
@@ -180,16 +179,49 @@
...
+            'filename': os.path.join(BASE_DIR, 'cac-tripplanner-app.log'),
+        },
     },
     'loggers': {
         'django': {
-            'handlers': ['console'],
-            'level': 'INFO',
-        }
+            'handlers': ['console', 'logfile'],
+            'level': 'DEBUG',
```
KlaasH: Is all this DEBUG-level logging necessary/useful?  Seems like it would be noisier than we'd want.

flibbertigibbet: Probably, yeah. I was trying to see if I could get any useful info for debugging an issue, but it is pretty noisy (and didn't help me anyway). I'll back it off to `INFO`.

flibbertigibbet:
Log level reduced for core Django. Ready for another look?

PR #900, opened 2017-10-15
Pluralize mode miles

### LINKED ISSUE #885: Travel time by mode- 1 mile or less should be singular
Currently displays "1 miles"

image

flibbertigibbet:
## Overview

Check for distances less than one mile or foot in per-mode summary text.
Use singular for those distances.

### Demo

image

### Notes

Also made modification for feet vs foot, although we shouldn't be displaying any distances that short.

## Testing Instructions

 * Plan a trip with a single mode distance of one mile or less
 * Trip detail text should display `mile` instead of `miles`
 * Screenshot above is of [this local trip plan url](http://localhost:8024/?origin=39.9615655%2C-75.154192&originText=990%20Spring%20Garden%20St%2C%20Philadelphia%2C%20Pennsylvania%2C%2019123&mode=WALK&maxWalk=482802&wheelchair=false&bikeTriangle=any&arriveBy=false&dateTime=&destination=39.9505412%2C-75.1489535&destinationText=N%205th%20St%20%26%20Market%20St%2C%20Philadelphia%2C%20Pennsylvania%2C%2019106)

## Checklist
- [x] No gulp lint warnings

Fixes #885.

### DIFF
--- a/src/app/scripts/cac/routing/cac-routing-itinerary.js
+++ b/src/app/scripts/cac/routing/cac-routing-itinerary.js
@@ -171,11 +171,13 @@ CAC.Routing.Itinerary = (function ($, cartodb, L, _, moment, Geocoder, Utils) {
     function getFormattedDistance(distanceMeters) {
         // less than ~0.2 miles
         if (distanceMeters < 322) {
-            return Math.round(distanceMeters * 3.28084).toString() + ' feet';
+            var feet = Math.round(distanceMeters * 3.28084);
+            return feet > 1 ? feet.toString() + ' feet' : feet.toString() + ' foot';
         }
 
         // return miles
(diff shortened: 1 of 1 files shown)


flibbertigibbet:
Personally I'd use plural for amounts less than one, but this is the behavior @lederer specifically requested on the issue.

flibbertigibbet:
Going to go ahead and merge this; if we want to change it to just 1 being singular later, it will be trivial.

lederer:
Sorry for lagging on this. Been consumed by an urgent project. Will take a look later today or tomorrow.

lederer:
I just tested and LGTM functionally. 

Agree we should only use the singular only when the value is integer 1. If I said otherwise somewhere, I don't know what I was smoking. I'll open a new issue for that.

PR #903, opened 2017-10-19
Only show singular for 1

### LINKED ISSUE #901: Use singular {mile, foot} only when value is integer 1
Connects #900 

We should display the singular unit ("mile" or "foot") only when the **displayed** value is the integer 1, not when it's less than 1.

flibbertigibbet:
When pluralizing miles/feet, only show singular for integer 1.
Closes #901.

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -8,10 +8,10 @@ set -x
 trap 'mark_unstable' ERR
 
 # Python linting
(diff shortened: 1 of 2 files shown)


flibbertigibbet:
less_than_one_mile

more_than_one_mile

one_mile

flibbertigibbet:
Trivial, merging

Issue #917, opened 2017-11-29
Create activities


flibbertigibbet:
Create a new model in the `destinations` app for `Activity`. Add related `activites` to `Destination` and `Event` models; allow multiple (or no) activities be assigned to a destination or event in the admin panel.

Create migration to add activities. Activities cannot themselves be modified via the admin interface.

PR #925, opened 2017-12-01
Add accessibility flag

### LINKED ISSUE #918: Add accessibility flag
Add a field to destinations and events for whether it is ADA accessible or not. Return field with API response.

flibbertigibbet:
## Overview

Add `accessible` flag to destinations and events, to note if they are ADA accessible or not.

### Demo

Editing an event in the admin panel:
accessible

## Testing Instructions

 * Run migrations
 * Should have new `accessible` field as a checkbox in the admin site for destinations and events
 * Flag should be returned in serialized destinations [here](http://localhost:8024/api/destinations/search?text=)

Closes #918.


flibbertigibbet:
build again

python/cac_tripplanner/destinations/models.py
```diff
@@ -90,6 +91,7 @@ class Meta:
                                    help_text='The large image. Will be displayed at 680x400.')
     published = models.BooleanField(default=False)
     priority = models.IntegerField(default=9999, null=False)
+    accessible = models.BooleanField(default=False, help_text='Is it ADA accessible?')
```
KlaasH: This duplication makes me wonder if these two should be children of a base class (`Attraction`??).  They share 8 properties ([`name`, `website_url`, `description`, `published`, `priority`, `image`, `wide_image`, `accessible`]) and have 6 and 3 of their own, respectively ([`city`, `state`, `zip`, `address`, `point`, `categories`] for `Destination`, [`start_date`, `end_date`, `destination`] for `Event`).

The activities created for #917 will also be going on both.

KlaasH: Awesome, looks great :circus_tent:

Issue #933, opened 2017-12-06
Make map destinations filter a drop-down


flibbertigibbet:
Instead of a row of toggle buttons.

See: https://marvelapp.com/59eh08g/screen/35591677

Issue #911, opened 2017-11-16
Display events in UI


flibbertigibbet:
Query for and present events in app alongside destinations. Show extra events fields (start and end date/times, and possibly an associated destination).

flibbertigibbet:
Blocked by #910.

flibbertigibbet:
See https://marvelapp.com/ae7jfde

flibbertigibbet:
Blocked by #934

lederer:
We need to determine the correct behavior for when there are no upcoming events to display.

Options:
1. Hide the events tab/filter completely.
1. Show a message.

I recommend showing a message, since (1) conditionally hiding the feature might confuse people who expect it, and (2) even when there's no event content, the presence of the feature naturally promotes itself to newcomers.

We'll need a better message than the current holdover from the pre-events era:

![image](https://user-images.githubusercontent.com/128699/34180080-5d1a2b4a-e4db-11e7-9d01-705cfe36ec2f.png)

flibbertigibbet:
The existing, complicated message display rules [from here](https://github.com/azavea/cac-tripplanner/issues/873#issuecomment-320709646) will need to be updated generally for the new filtering.

PR #936, opened 2017-12-06
Feature/kak/filter destinations ui#847

### LINKED ISSUE #847: UI for displaying and filtering destinations by category
Possibly on new page like the articles page ('learn' tab), and/or by adding a form above front page (home/map) list.

flibbertigibbet:
## Overview

Adds filter bar to top of destinations list. Filter the destinations list on filter bar item selection.

### Demo

Optional. Screenshots, `curl` examples, etc.
image

### Notes

Created #933 for the map view filter, which is supposed to be a drop-down instead of a button bar.

The "events" category is not actually a destination category, but will return the new `Event` type instead of destinations. #911 will add those. For now, it returns nothing.

## Testing Instructions

 * `vagrant ssh app`
 * `cd /opt/app/src'
 * `npm run gulp-development`
 * Be sure to clear the service worker cache
 * Try filtering destinations from the homepage.

## Checklist
- [x] No gulp lint warnings
- [x] Gulp tests pass
- [x] No python lint warnings
- [x] Python tests pass

Closes #847


src/app/scripts/cac/control/cac-control-explore.js
```diff
@@ -405,6 +414,21 @@ CAC.Control.Explore = (function (_, $, Geocoder, MapTemplates, HomeTemplates, Ro
         displayPlaces(destinations, $(options.selectors.isochroneSlider).val());
     }
 
+    /**
+     * Filter destinations by category client-side.
+     */
+    function filterPlaces(allPlaces, filter) {
+        if (filter === 'All') {
```
KlaasH: When I load the home page and click "Map View", I get this:
image

Because `filter` is undefined when it gets here. (If I click "All" it then loads the destinations.)

python/cac_tripplanner/templates/home.html
```diff
@@ -62,8 +62,35 @@
...
+                                data-filter="Exercise">
+                                <span class="filter-label">Exercise</span>
+                            </div>
+                            <div class="relax filter-option" title="Relax" data-filter="Relax">
+                                <span class="filter-label">Relax</span>
+                            </div>
+                            <div class="educational filter-option" title ="Educational"
+                                data-filter="Educational">
+                                <span class="filter-label">Educational</span>
+                            </div>
```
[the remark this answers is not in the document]
flibbertigibbet: I'd really like to not have the header section templated. There was a reason for that before, but maybe it can be ripped out.

There's going to be styling dependent on the category list, which is supposed to be preset for that reason. It might just be icons for each though, but that would be information the front-end should have and not just the python template.

src/app/scripts/cac/control/cac-control-explore.js
flibbertigibbet: This is just on the map view, correct? #933 exists for the map view filter.

python/cac_tripplanner/templates/home.html
```diff
@@ -62,8 +62,35 @@
     <div class="places">
         <div class="places-content">
             <header class="places-header">
-                <h1>Places we love</h1>
-                <a href="#" class="map-view-btn">Map View</a>
+                <div class="places-header-content">
```
lederer: Is `div.places-header-content` needed? Looks like it's the only child of `header.places-header`.

python/cac_tripplanner/templates/home.html
```diff
@@ -62,8 +62,35 @@
...
+                    <h1>Places we love</h1>
+                    <a href="#" class="map-view-btn">Map View</a>
+                    <div class="filter-picker">
+                        <div class="filter-toggle">
+                            <div class="all filter-option on" title="All" data-filter="All">
+                                <span class="filter-label">All</span>
```
lederer: Is there a structural need for `span.filter-label`?

src/app/scripts/cac/control/cac-control-explore.js
KlaasH: Yeah, just on the map view.  I was interpreting #933 as just about the interface, but yeah, this can be handled there.

python/cac_tripplanner/templates/home.html
flibbertigibbet: Added for styling reasons. The outer `div` forces its margins out to fill the background color to light grey, as in the Marvel prototype. The inner `div` has margins to match the content sizing.

This was the best I could come up with, as Marvel is a black box that cannot be inspected to see elements or styling in use, so there's no styling or functional prototype to go off of. A cleanup PR for styling would be welcome.

python/cac_tripplanner/templates/home.html
flibbertigibbet: The class is there for styling, to match the usage from the mode selectors in the directions form above it. The class is not presently referenced in the JavaScript code, if that's what's meant by "structural."

flibbertigibbet:
@lederer unclear what changes are being requested. Is there something specific you'd like modified on this PR?

python/cac_tripplanner/templates/home.html
lederer: Got it. No problem.

python/cac_tripplanner/templates/home.html
lederer: Great. That answers my question.

Issue #940, opened 2017-12-12
Reload place list header and content separately


flibbertigibbet:
Currently, whenever the place list or its header change, both are reloaded.

This used to be the case because the header would go away entirely to display error or no-results messages, and to display the loader icon. Now that there is a filter bar present that can return no results for a filter, messages need to be displayed below the header instead of removing it, so the filter can be changed again. Although the message content has been moved, the header still gets reloaded with the list.

When navigating between the home and map views, swap just the header content. There are already functions present to get the compiled HTML for both headers.

When reloading the destinations list, do not also redraw the header unless needed. Ensure the loader animation continues to display appropriately.

PR #941, opened 2017-12-12
Filter destinations in the map sidebar

### LINKED ISSUE #933: Make map destinations filter a drop-down

### LINKED ISSUE #937: Map destinations filter should filter displayed map pins
Category filter. Related to #933

flibbertigibbet:
## Overview

Implement a dropdown destinations filter for the map sidebar. Continue to use a button bar for filtering on the homepage, and switch between the two as appropriate.

### Demo

no_events_isochrone

### Notes

The [prototype](https://marvelapp.com/59eh08g/screen/35591677) shows the word 'Filter' at the top of the filter bar, instead of the current selection. That is not typical drop-down behavior, and gives no clear benefit; there will always be a selection, as 'All' is the default. A standard `select` element is in use here.

As visible above, destinations that do not match the filter currently have a lower opacity, as do those outside the isochrone bounds. #937 is to not display locations that do not match the filter at all.

Created #940 to separate place list header and content reloading. Adding the filter bar has changed interaction such that the header shouldn't disappear on list reload; for now, it reloads whenever the list does.

## Testing Instructions

 * `vagrant ssh app`
 * `cd /opt/app/src && npm run gulp-development`
 * Ensure service worker cache is cleared
 * Filtering on home and map pages should behave as expected
 * Navigating between home and map pages should switch between the two filter types

## Checklist
- [x] No gulp lint warnings
- [x] Gulp tests pass

Closes #933
Closes #937

### DIFF
--- a/src/app/scripts/cac/control/cac-control-explore.js
+++ b/src/app/scripts/cac/control/cac-control-explore.js
@@ -473,7 +473,8 @@ CAC.Control.Explore = (function (_, $, MapTemplates, HomeTemplates, Routing, Use
         return dfd.promise();
     }
 
-    function _getNearbyPlaces(filter) {
+    function _getNearbyPlaces() {
+        var filter = UserPreferences.getPreference('destinationFilter');
         showSpinner();
         var $placeCards = $(options.selectors.placeCard);
         // hide existing times to places now showing (if any)
--- a/src/app/scripts/cac/control/cac-control-explore.js
+++ b/src/app/scripts/cac/control/cac-control-explore.js
@@ -38,6 +38,7 @@ CAC.Control.Explore = (function (_, $, MapTemplates, HomeTemplates, Routing, Use
     var fetchingIsochrone = false;
 
     var allDestinations = []; // cache full list of destinations
+    var isochroneDestinationIds = null; // cache IDs of destinations within isochrone
 
     var events = $({});
     var eventNames = {
@@ -109,8 +110,7 @@ CAC.Control.Explore = (function (_, $, MapTemplates, HomeTemplates, Routing, Use
             $(options.selectors.isochroneSliderContainer).removeClass(options.selectors.hiddenClass);
             clickedExplore();
         } else {
-            $(options.selectors.alert).remove();
-            mapControl.isochroneControl.clearIsochrone();
+            clearIsochrone();
             mapControl.isochroneControl.clearDestinations();
             $(options.selectors.isochroneSliderContainer).addClass(options.selectors.hiddenClass);
         }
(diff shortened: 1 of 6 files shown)


KlaasH:
The interaction between this category filtering and the existing distance filtering for destinations seems to be causing some trouble.  If I load [destinations within a 15-minute bike ride of the office](http://localhost:8024/?origin=39.9615655%2C-75.154192&originText=990%20Spring%20Garden%20St%2C%20Philadelphia%2C%20Pennsylvania%2C%2019123&mode=BICYCLE&maxWalk=482802&wheelchair=false&bikeTriangle=any&arriveBy=false&dateTime=&placeId=&exploreMinutes=15), it initially works fine (shows Fairmount Waterworks and Independence Seaport Museum), but if I select a category from the filter droplist, it loads all the desinations matching that filter, even if they're outside the given range.  Then when I change it back to "All" it shows all destinations.

flibbertigibbet:
Hmm, good find @KlaasH. @lederer should the explore view list be filtered for both?

lederer:
Yup. The list should be filtered by both the travelshed and the category. 

Nice catch @KlaasH. Glad this PR had two reviewers. 😳

flibbertigibbet:
Updated filtering; ready for another look. Also added map marker category filtering to also close #937.

src/app/scripts/cac/control/cac-control-explore.js
```diff
@@ -401,6 +415,8 @@ CAC.Control.Explore = (function (_, $, Geocoder, MapTemplates, HomeTemplates, Ro
 
         // now places list has been updated, go fetch the travel time
         // from the new origin to each place
+
+        // TODO: cache travel times for last origin/routing params set?
```
KlaasH: Is there an issue about this?

src/app/scripts/cac/control/cac-control-filter-options.js
```diff
@@ -30,26 +30,46 @@ CAC.Control.FilterOptions = (function ($) {
...
+        $(options.selectors.filterToggle).on('click', options.selectors.filterOption, function(e) {
+            e.preventDefault();
+            toggleOn(this);
+            events.trigger(eventNames.toggle, getFilter());
+        });
+
+        // filter drop-down button event handler
+        $(options.selectors.filterToggle).on('change', function(e) {
+            e.preventDefault();
+            if (!e.target.selectedOptions || ! e.target.selectedOptions.length) {
```
KlaasH: Space after `!` looks odd, especially right after one with no space.

src/app/scripts/cac/control/cac-control-explore.js
flibbertigibbet: Not yet, but it occurred to me that with the new filters, requeries will be fired off more frequently and often times will be unnecessary. I'll go make an issue and note it here.

src/app/scripts/cac/control/cac-control-filter-options.js
flibbertigibbet: Oops, that was accidental. Good catch.

src/app/scripts/cac/control/cac-control-explore.js
flibbertigibbet: #945

Issue #934, opened 2017-12-06
Styling for 2017 enchancements


flibbertigibbet:
Add styling needed to implement changes in https://marvelapp.com/59eh08g

Includes adding new icons for the watershed alliance, accessibility, activities, and the destination categories, adding a layout for events, and modifying the destination layout to accommodate the new attributes.

lederer:
See #936, #941

lederer:
This is done

PR #951, opened 2017-12-19
Style categories UI

lederer:
## Overview

This PR styles the category tabs on the home page and the category dropdown menu on the map page.

### Demo

image

image

### Notes

A couple of things to beware of. I'll open issues for these.

1. Category _tabs_ are shown on the home page. Category _dropdown_ is shown on the map page. However, we'll need to change this so the _dropdown_ replaces the _tabs_ on the home page in viewports < 768px wide. Otherwise we get:
image

1. Map view loads _tabs_ first, before switching to _dropdown_, in this case: 
    - Go to Map view (Explore)
    - Set an origin
    - Expand travelshed if necessary
    - Click "More info" on a place card
    - Click the back button

Connects #934, #936, #941


lederer:
Added a commit to make the home page category tabs fit on narrow screens. Once we remove the "Relax" category, they should fit fine:

image

lederer:
Thanks for reviewing @flibbertigibbet 

> Note: this doesn't address the new things for #934

Yup, I'm working on the updated card format now. And will follow up with place/event detail page.

PR #955, opened 2017-12-20
Allow activities and categories to be empty

### LINKED ISSUE #954: Remove requirement to select activity on destinations/events
Right now, the admin user is required to select both an activity and a category for any newly created destinations or events before saving. While category should be required, activity should not.

gpg screenshot 12 7 17

flibbertigibbet:
Fixes #954.


flibbertigibbet:
build again

Issue #935, opened 2017-12-06
Add attributes to destinations


flibbertigibbet:
Add watershed alliance icon, accessibility icon, available activities, and categories (possibly with icons) to the destinations layout.

Blocked by #934

flibbertigibbet:
See https://marvelapp.com/59eh08g/screen/35591679

PR #960, opened 2017-12-22
Add categories, activities, etc to place cards, pages. Add events UI.

lederer:
## Overview

This PR integrates categories, activities, AWE membership, and wheelchair accessibility into place cards and place detail pages.

It also integrates events into place detail cards and introduces a new partial for event detail pages.

This PR is HTML and SASS only, and will require application-level changes to wire up the new features (see Notes below).

##### Place, without origin

image

##### Place, with origin

image

##### Event, without origin

image

##### Event, with origin

image

##### Desktop

image

##### Mobile

image

#### Event cards
I considered generalizing the `.place-card` class to something like `.preview-card`, since they'll be used for both places and events. But`.preview-card` is already in use elsewhere in the app and I ultimately decided to leave it as is. 

When a `.place-card` is used for an event, the `.event-card` class must be added to the `.place-card` element, and the date and time elements should be populated.

#### New images
`awe-logo.png` and `awe-logo@2x.png` need to be wired up to the `src` and `srcset` attributes of `.place-awe-link .image` in `destination-detail.html`.

`awe-icon.png` and `awe-icon@2x.png` need to be wired up to the `src` and `srcset` attributes of `.place-card-badges .image` in `home.html` and `cac-home-templates.js`.

#### Activities
I've added icons for bike trails (aka cycling) and wheelchair accessibility (not an activity per se, but we're including it with the activity icons in the UI) to the fontello bundle.

I can add additional icons as additional activities are determined.

#### Event detail page
I created `partials/event-detail.html`, based on `partials/destination-detail.html`. I reckon we can use the `place-detail.html` template for both place and event detail pages, and just include the appropriate partial.

I changed all the `destination.*` variables to `event.*` and added some new ones (eg, `event.location`, `event.facebook_url`). Consider these pseudocode to get the point across. 

Note the event date and time are broken down into components, for styling.

Once this new partial is wired up to render events, I can do a follow-up PR to style the page.

## Testing Instructions

 * Clear your cache.
 * Ensure cards on the home page look like the images above.
   * Add the `.event-card` class to a `.place-card` to force it into event mode.
 * Go to a place's detail page. Ensure it looks like the images above.
 * Note the event partial can't be tested in this PR.

Connects #934, #935


lederer:
> On clicking a home page filter option, stray anchor tags show

Doh! Fixed and force pushed. 

Thanks for reviewing. Will wait for your formal approval after #911 is done.

flibbertigibbet:
Merging to do #911

Issue #963, opened 2018-01-05
Revisit cycling icon tooltip


lederer:
@KlaasH writes in #962:

> This is minor, but the icon that comes up for an event with the "cycling" activity attached has the title text "Biking trails". This is probably reasonable for destinations (though I wonder if there are cases where a destination would be marked "cycling" but not because of trails, e.g. somewhere that's good for on-road cycling), but for an event it seems potentially incongruous.

Options:
- "Bike paths" / "Biking paths" / "Cycling paths" (more general than "trails"?)
- "Cycling" (super general)

PR #971, opened 2018-01-15
Events UI polish

lederer:
## Overview

This PR polishes up the format, layout, and visual treatment of event dates and times, as a follow-up to the new events system in #962 

I also slipped a couple of unrelated tweaks in:
- Cycling activity tooltip: "Biking trails" -> "Cycling" (#963)
- Changed the spinner color to fit overall site visd

#### Multi-day and one-day event preview cards
image

#### One-day event detail, wide
image

#### One-day event detail, narrow
image

#### Multi-day event detail, wide
image

#### Multi-day event detail, narrow
image

### Notes

As a consequence of getting the narrow event detail page to look reasonable, the narrow *place* detail page has also changed slightly. The activities are now rendered above the place name instead of below the place actions. I think this is actually an improvement.

image

## Testing Instructions

 * Create a one-day and a multi-day event
 * Confirm their previews cards look reasonable
 * Confirm their detail pages look reasonable, across screen widths

Connects #962, #963

### DIFF
--- a/python/cac_tripplanner/templates/home.html
+++ b/python/cac_tripplanner/templates/home.html
@@ -158,7 +158,7 @@
                         <div class="place-card-badges">
                             {% has_activity destination 'cycling' as has_cycling %}
                             {% if has_cycling %}
-                            <span class="badge activity" title="Biking trails">
+                            <span class="badge activity" title="Cycling">
                                 <i class="icon-cycling"></i>
                             </span>
                             {% endif %}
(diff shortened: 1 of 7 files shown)


flibbertigibbet:
I'm seeing a blue line to the right of the places list, over the home page map:

image

lederer:
Thanks for reviewing!

Blue line is weird. Can't repro on major Windows and Mac browsers. Are you seeing on Linux Chrome?

flibbertigibbet:
Yes, only seeing on initial page load for home page. Does not require cache clear.

Not seeing on Firefox. Further oddity: only seeing when browser window is on external monitor. Probably safe to ignore that issue for now. In any case, it's not related to this PR. I'll open an issue if I can reproduce it under some other conditions.

lederer:
Ok sounds good. So weird.

PR #993, opened 2018-02-05
Feature/kak/optimizer flags#986

### LINKED ISSUE #986: Replace bike triangle weighting with optimize flags
For bicycle trips, the UI offers options to specify 'any', 'flat', 'fast', or 'safe' as a ride preference, which [set weights](https://github.com/azavea/cac-tripplanner/blob/123ece8fa5a47c4f560ed39d83d975124918db08/src/app/scripts/cac/utils.js#L57) for the [OTP optimizer "triangle" parameter](http://dev.opentripplanner.org/apidoc/0.20.0/resource_PlannerResource.html) in .

For these options, switch from sending "triangle" optimizer values to using the optimizer flags:
 - 'any': no flag
 - 'flat': `FLAT`
 - 'fast': `QUICK`
 - 'safe': `GREENWAYS`

Note that `GREENWAYS` does not differentiate path type in another way that `SAFE`, but rather simply uses a [heavier weighting](https://github.com/opentripplanner/OpenTripPlanner/blob/34a24d183324e01f7e428d53924226bc56b3988b/src/main/java/org/opentripplanner/routing/edgetype/StreetEdge.java#L338), so it is something of a misnomer. A "Greenway" is [assumed to be any bike lane 10x as safe as an ordinary street](https://github.com/opentripplanner/OpenTripPlanner/commit/ec9db2e36e5f099af80ecfb0c378865c20f80cb8).

flibbertigibbet:
## Overview

Switch away from using the "bike triangle" optimizer parameters, which set values summing to one for how much to weight for safety, speed, and elevation change when planning a bicycle trip, to using optimizer mode flags for each three of those considerations. The "safe" flag replacement, `GREENWAYS`, also works for walking trips and is sent for all non-bike trips. The default for bike trips has been changed from 'any' (no optimizer flag sent) to 'safe' (the new `GREENWAYS` flag).

Also, this PR fixes the spinner not showing on the homepage on place list reload, and modifies the way origin/destination marker drags are handled. The marker locations when getting directions are updated to the start/end points pulled from the encoded polyline for the first itinerary returned, in order to pin the marker to the street network (in case it was dropped somewhere unroutable), instead of moving the marker to its reverse geocoded location. This fixes issues around attempting to route to/from places that are routable, but not by a street address known to the geocoder, like on paths.

### Demo

Default bike route, now going along the Camden waterfront (with 'safe' mode):
image

On changing to 'any' mode (no optimization flag set; OTP defaults to maximize for speed):
image

### Notes

[Here](https://github.com/opentripplanner/OpenTripPlanner/commit/ec9db2e36e5f099af80ecfb0c378865c20f80cb8) is what the `GREENWAYS` flag actually does.

The inbuilt OTP trip planner app runs locally [here](http://localhost:9090/). It can be helpful to see behavior without origin/destination markers getting snapped to the itinerary/street network.

## Testing Instructions

Note: going to pull updated OSM tomorrow (Tuesday) and provide a rebuilt graph for testing that should reflect behavior that would be in the next deployment.

 * Toggling routing options in modal should function as expected
 * Try planning trips to check for any strange behavior, especially around paths
 * Try moving markers around; if dropped somewhere unroutable, should move to start/end of path
 * Spinner on home page should function as expected on origin change or filter apply

Closes #986

### DIFF
--- a/src/app/scripts/cac/control/cac-control-directions.js
+++ b/src/app/scripts/cac/control/cac-control-directions.js
@@ -152,6 +152,12 @@ CAC.Control.Directions = (function (_, $, moment, Control, Routing, UserPreferen
                 itineraryControl.draggableItinerary(currentItinerary);
             }
 
+            // snap start and end points to where first itinerary starts and ends
+            // (in case one or both markers is someplace unroutable, like in a river)
+            directions.origin = [currentItinerary.from.lat, currentItinerary.from.lon];
+            directions.destination = [currentItinerary.to.lat, currentItinerary.to.lon];
+            updateLocationPreference('origin');
+            updateLocationPreference('destination');
             // put markers at start and end
             mapControl.setDirectionsMarkers(directions.origin, directions.destination);
             itineraryListControl.setItineraries(itineraries);
@@ -419,6 +425,31 @@ CAC.Control.Directions = (function (_, $, moment, Control, Routing, UserPreferen
     // Updates the URL to match the currently-selected options
     function updateUrl() {
         urlRouter.updateUrl(urlRouter.buildDirectionsUrlFromPrefs());
+    }
+
+    /** Helper to save current directions origin or destination to user preferences.
+     *
+     * @param location {String} Either 'origin' or 'destination'
+     */
+    function updateLocationPreference(location) {
+        var preferenceLocation = UserPreferences.getPreference(location);
+        var newLocationY = directions[location][0];
+        var newLocationX = directions[location][1];
+
+        preferenceLocation.location = {
+            x: newLocationX,
+            y: newLocationY
+        };
+
+        preferenceLocation.extent = {
+            xmax: newLocationX,
+            xmin: newLocationX,
+            ymax: newLocationY,
+            ymin: newLocationY
+        };
+
+        // update location in user preferences, but not the string for it
+        UserPreferences.setPreference(location, preferenceLocation);
     }
 
     /**
--- a/src/app/scripts/cac/control/cac-control-directions.js
+++ b/src/app/scripts/cac/control/cac-control-directions.js
@@ -2,7 +2,7 @@
  *  View control for the directions form
  *
  */
-CAC.Control.Directions = (function (_, $, moment, Control, Routing, UserPreferences, Utils) {
+CAC.Control.Directions = (function (_, $, moment, Control, Routing, UserPreferences) {
 
     'use strict';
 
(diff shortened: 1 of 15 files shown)


src/app/scripts/cac/control/cac-control-directions.js
```diff
@@ -421,6 +427,31 @@ CAC.Control.Directions = (function (_, $, moment, Control, Routing, UserPreferen
         urlRouter.updateUrl(urlRouter.buildDirectionsUrlFromPrefs());
     }
 
+    /** Helper to save current directions origin or destination to user preferences.
+     *
+     * @param location {String} Either 'origin' or 'destination'
+     */
+    function updateLocationPreference(location) {
```
KlaasH: I feel like the behavior of this is a bit not initially obvious, and maybe some naming changes would help.  Seeing `updateLocationPreference(location)`, my first thought would be that this is going to save a location stored in `location` to user preferences.  I think renaming the argument to something that makes it more obvious that it's the key rather than the actual location would help (I can't think of anything better than `key` off hand, though that's not ideal).  And though it would make the function name a little ridiculous, calling it `updateLocationPreferenceFromDirections` would make it more obvious that the source of the location is the `directions` variable on the parent scope.

flibbertigibbet: Since user preference has a helper method called `setLocation` which is how we generally set `origin` and `destination`, this seemed like a reasonable name. Added the docstring to clarify the param usage, but sure, can rename the param too.

lederer:
Works as described!

Defaulting "ride quality" to Safe makes sense, but also breaks the current UI. When there is a choice between a set of specific values and "Any", it feels odd for "Any" to not be the default. Relatedly, it feels odd for the UI to highlight when "Any" is turned on, and to have to choose the non-obvious "Safe" to turn then off the highlighting / reset to default. 

An easy solution might be to eliminate "Any" altogether. The ride item in the main options menu could be "Ride quality: Safe" by default (and "Ride quality: Fast", etc, otherwise). In the ride quality submenu we could indicate that Safe is the default with a "Default" label.

(It would also be nice to redo the options menu entirely at some point. Definitely room for improvement there.)

flibbertigibbet:
Agreed, could leave out 'any' at this point, since it will generally function the same as 'fast'. I'll take it out, and change the 'safe' label to say it's the default.

lederer:
Great. And I'll style the default label.

PR #1000, opened 2018-02-22
Cancel outstanding trip plan request

### LINKED ISSUE #998: Changing directions inputs while request in progress
If directions inputs are modified while a directions request is already in progress, usually the first request cancels and the request is re-submitted with the latest parameters. However, on rare occasion, if the user toggles an input with a request in progress, sometimes stale results get displayed (for input state that is not the latest).

This seems to occur if the input was toggled in the brief window after the results of the first request have returned, but before they have been displayed.

Transit results returned when transit button toggled off after request sent with transit on:
image

flibbertigibbet:
## Overview

When finding directions in response to directions form input, cancel previous routing request
before issuing a new one.

Fixes outdated results being displayed when the routing requests resolve out of order.

### Notes

Changed `throttle` to `debounce` in case [this lodash issue](https://github.com/lodash/lodash/issues/3051) still exists (issue was closed in favor of a related PR, but the PR was never merged).

Changed trip plan routing function to optionally return `Deferred` instead of an immutable `Promise` so it can be cancelled in the requesting control. Did that instead of modifying the trip plan routing function to cancel requests itself, as in other cases we do want to allow multiple routing requests to fire simultaneously and return as soon as they resolve, particularly when finding travel times from the origin to all destinations.

## Testing Instructions

(To reproduce the issue, try the steps below on the production site.)

 * Plan a bicycle trip; the longer the better, to increase query time
 * After loading animation displays, toggle the transit button repeatedly while animation is still showing
 * Loading animation should continue to show until results display
 * Itineraries displayed should match the last directions form input options set (with transit, or not)
 * Finding travel times to full destinations list should continue to work as expected

Fixes #998.

### DIFF
--- a/src/app/scripts/cac/control/cac-control-directions.js
+++ b/src/app/scripts/cac/control/cac-control-directions.js
@@ -22,7 +22,6 @@ CAC.Control.Directions = (function (_, $, moment, Control, Routing, UserPreferen
     };
     var options = {};
 
-    // cancel previous trip request when issuing a new one
     var planTripRequest = null;
     var OUTDATED_REQUEST_ERROR = 'outdated request';
 
@@ -180,9 +179,7 @@ CAC.Control.Directions = (function (_, $, moment, Control, Routing, UserPreferen
             itineraryListControl.show();
             // highlight first itinerary in sidebar as well as on map
             findItineraryBlock(currentItinerary.id).addClass(options.selectors.selectedItineraryClass);
-            planTripRequest = null;
         }, function (error) {
-            planTripRequest = null;
             // Cancelled requests are expected; do not display an error message to user.
             // Just keep showing the loading animation until a subsequent request completes.
             if (error && error.message === OUTDATED_REQUEST_ERROR) {
@@ -196,7 +193,7 @@ CAC.Control.Directions = (function (_, $, moment, Control, Routing, UserPreferen
             itineraryListControl.setItinerariesError(error);
             itineraryListControl.show();
         });
-    }, DIRECTION_THROTTLE_MILLIS);
+    }, DIRECTION_THROTTLE_MILLIS, {leading: true, trailing: true});
 
     return DirectionsControl;
 
--- a/src/app/scripts/cac/control/cac-control-directions.js
+++ b/src/app/scripts/cac/control/cac-control-directions.js
@@ -22,6 +22,10 @@ CAC.Control.Directions = (function (_, $, moment, Control, Routing, UserPreferen
     };
     var options = {};
 
+    // cancel previous trip request when issuing a new one
+    var planTripRequest = null;
+    var OUTDATED_REQUEST_ERROR = 'outdated request';
+
     var currentItinerary = null;
 
     var directions = {
@@ -98,7 +102,7 @@ CAC.Control.Directions = (function (_, $, moment, Control, Routing, UserPreferen
      * Set user preferences before planning trip.
      * Throttled to cut down on requests.
      */
-    var planTrip = _.throttle(function() {  // jshint ignore:line
+    var planTrip = _.debounce(function() {  // jshint ignore:line
         showPlaces(false);
         if (!(directions.origin && directions.destination)) {
             directionsFormControl.setError('origin');
@@ -128,8 +132,20 @@ CAC.Control.Directions = (function (_, $, moment, Control, Routing, UserPreferen
 
         tabControl.setTab(tabControl.TABS.DIRECTIONS);
 
-        Routing.planTrip(directions.origin, directions.destination, date, otpOptions)
-        .then(function (itineraries) {
+        // If a previous request is in progress, cancel it before issuing the new one.
+        // Prevents callbacks being executed out of order.
+        if (planTripRequest) {
+            planTripRequest.reject(Error(OUTDATED_REQUEST_ERROR));
+        }
+
+        planTripRequest = Routing.planTrip(directions.origin,
+                                           directions.destination,
+                                           date,
+                                           otpOptions,
+                                           true);
+
+
+        planTripRequest.then(function (itineraries) {
             $(options.selectors.spinner).addClass(options.selectors.hiddenClass);
             // Add the itineraries to the map, highlighting the first one
             var isFirst = true;
@@ -164,7 +180,15 @@ CAC.Control.Directions = (function (_, $, moment, Control, Routing, UserPreferen
             itineraryListControl.show();
             // highlight first itinerary in sidebar as well as on map
             findItineraryBlock(currentItinerary.id).addClass(options.selectors.selectedItineraryClass);
+            planTripRequest = null;
         }, function (error) {
+            planTripRequest = null;
+            // Cancelled requests are expected; do not display an error message to user.
+            // Just keep showing the loading animation until a subsequent request completes.
+            if (error && error.message === OUTDATED_REQUEST_ERROR) {
+                return;
+            }
+
             console.error('failed to plan trip');
             console.error(error);
             $(options.selectors.spinner).addClass(options.selectors.hiddenClass);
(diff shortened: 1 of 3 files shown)


tnation14:
build again

src/app/scripts/cac/control/cac-control-directions.js
```diff
@@ -98,7 +102,7 @@ CAC.Control.Directions = (function (_, $, moment, Control, Routing, UserPreferen
      * Set user preferences before planning trip.
      * Throttled to cut down on requests.
      */
-    var planTrip = _.throttle(function() {  // jshint ignore:line
+    var planTrip = _.debounce(function() {  // jshint ignore:line
```
KlaasH: What are the pros and cons of this?  On the pro side, it will drop more useless requests in the case of fast switching back and forth (though that doesn't seem like a huge real-world case) or someone clicking into directions then immediately changing mode options.  But the downside is that for simple interactions where the user clicks into directions with the desired mode options already selected--i.e. there's only one request total--this will make it .75 seconds slower.

We could set the option `leading=true` to avoid that, though that would mean the "get directions, switch mode" cases would always produce 2 requests rather than sometimes dropping the first.  It would still save requests vs. the current throttle in a rapid-toggling situation.

src/app/scripts/cac/places/cac-places.js
```diff
@@ -127,7 +127,7 @@ CAC.Places.Places = (function(_, $, moment, Routing, UserPreferences) {
             if (xCoord && yCoord) {
                 var placeCoords = [yCoord, xCoord];
                 // get travel time to destination and update place card
-                Routing.planTrip(exploreLatLng, placeCoords, date, otpOptions)
+                Routing.planTrip(exploreLatLng, placeCoords, date, otpOptions, false)
```
KlaasH: Having `Routing.planTrip` switch its return type based on an argument seems not ideal to me.  This is only invoked here and in `cac-control-directions.js`, right?  Couldn't we just tack `.promise()` onto the end of this line and have the function always return a Deferred?

src/app/scripts/cac/control/cac-control-directions.js
```diff
@@ -164,7 +180,15 @@ CAC.Control.Directions = (function (_, $, moment, Control, Routing, UserPreferen
             itineraryListControl.show();
             // highlight first itinerary in sidebar as well as on map
             findItineraryBlock(currentItinerary.id).addClass(options.selectors.selectedItineraryClass);
+            planTripRequest = null;
```
KlaasH: If `planTrip` got invoked again while the previous one was in this block, it would call `planTripRequest.reject()` after it had already resolved.  Maybe that's not possible in practice, and maybe it wouldn't be a big deal if it happened, but I'd be inclined to put this at the top of the callback just in case, since I don't see any disadvantage to doing so.

src/app/scripts/cac/control/cac-control-directions.js
flibbertigibbet: See notes section above: there is an issue with the lodash throttle implementation.

In testing, I haven't found any noticeable difference in responsiveness; the actual query time is much longer than the debounce window.

src/app/scripts/cac/places/cac-places.js
flibbertigibbet: Given that it's convention everywhere else in this app and recommended by jQuery to always pass around non-mutable promises, I thought keeping that as the default behavior would be the least confusing thing to do. Could change that if you feel strongly otherwise.

src/app/scripts/cac/control/cac-control-directions.js
flibbertigibbet: Reasoning in nullifying the request as late as possible is so that display of known outdated results can be aborted, rather than doing unnecessary work and potentially delaying display of the newer response. According to this [docs example](https://api.jquery.com/deferred.promise/), though, it looks like it would have no effect on a completed promise.

KlaasH: Ok, yeah, no difference then, and calling `reject` after it's `resolved` is ignored (I was thinking it might throw an error).  So this is fine as-is.

src/app/scripts/cac/places/cac-places.js
KlaasH: I see what you mean.  Though this still looks weird to me, at least it's explicit.  I don't feel strongly either way.

src/app/scripts/cac/control/cac-control-directions.js
KlaasH: I had to go back and read [the debounce vs. throttle explainer](https://css-tricks.com/debouncing-throttling-explained-examples/) again and by the time I had finished I had somehow dropped the bit about the throttle bug from my brain.  And yeah, I think debounce makes more sense here, since even without the bug in throttle, sending requests in the middle of a stream of rapid invocations wouldn't be useful.

But I still think debounce-with-leading-edge seems preferable.  It still wouldn't be the same as throttle (which is, for Lodash, debounce-with-leading-edge-and-max-wait), though it would be susceptible to a similar effect, where if the user waits 751ms between requests the debounced grouping would close and send the trailing request and 1ms later the leading request of the next batch would go.

Running a few simple directions lookups on the live site, I'm getting responses in the 400-500ms range.  For longer trips, I'm seeing 1-1.25 seconds.  Adding 750ms to all of those seems to me like a potentially significant/noticeable difference, and I don't think it's worth it to add that lag to what seems to me like a core/common interaction to save load in what seem like less common scenarios.

flibbertigibbet: Added leading edge.

flibbertigibbet: On adding leading edge, the issue is reintroduced:

image

flibbertigibbet: Since there's logic now to throw away outdated, pending requests, maybe simply reducing the debounce timeout would be fine.

KlaasH: Even with calls made in quick succession, as can happen with `leading` turned on, it shouldn't be possible for stale results to get displayed because the calls will always be in order and superseded ones should be getting canceled.  I think I see how it's happening.  I set `leading` to true and added some logging, then tested with [this trip](http://localhost:8024/?origin=40.338648%2C-75.11031&originText=Boxwood%20Cir%20%26%20Cold%20Spring%20Creamery%20Rd%2C%20Doylestown%2C%20Pennsylvania%2C%2018902&mode=WALK&maxWalk=482802&wheelchair=false&bikeOptimize=GREENWAYS&arriveBy=false&dateTime=&destination=39.878307%2C-75.268882&destinationText=John%20Heinz%20National%20Wildlife%20Refuge&waypoints=40.2171598%2C-75.0112152%3B40.0486427%2C-75.0359344), which is significantly faster without transit than with (maybe that's true in most cases, but just to have a working example written down...).  Start from no-transit, I click three times quickly (on-off-on), wait the moment it takes for the debounce timeout to pass and the spinner to appear, then click once more to turn transit back off.  So it sends the first click (with transit) as the first leading request, ignores the second click, sends the third (no transit) as the trailing request, then sends the fourth click (with transit) as the leading request of the next debounce grouping.  That leaves me in a state where I should be getting walking directions but the transit directions are reliably coming back last and showing up.

The console looks like this:
image

I've annotated it with my interpretation of which requests the lines are talking about.  The interesting part is that when the trailing request comes back, it should have been canceled by the 2nd leading request, and in fact `planTripRequest` is null, but the `.then` callback is still firing.  My interpretation is that when `planTripRequest` gets set to `null`, that doesn't destroy the object, and the callback is still attached to its success handler.  On this theory, the fact that the right itinerary is always winning currently has mainly to do with the 750ms minimum gap between requests being enough to keep things in order.

Adding a check and bailing out if `planTripRequest` is falsy at the top of the `then` function appears to cause the final selection to always end up being the one displayed.

flibbertigibbet: Modified to never nullify the trip request, since rejecting an already completed request does nothing and does not error, and it seems to resolve the leading edge issue.

KlaasH: Yay!  Repeating the above with the new code, I now see two rejections--i.e. Leading 1 gets rejected by Trailing, which in turn gets rejected by Leading 2--which is as it should be and results in it showing the right thing.

Issue #1005, opened 2018-03-07
Service worker errors preventing trip display


flibbertigibbet:
Planned trips fail to render in some browsers with a cached version of the site.

Seems to be dependent on having an older version of the service worker previously installed. On Chrome, the planned trips fail to render; on Firefox, they show, but there are console errors.

See [this similar issue with Chrome 64](https://stackoverflow.com/questions/48463483/what-causes-a-failed-to-execute-fetch-on-serviceworkerglobalscope-only-if).

Chrome errors:
`Failed to execute 'fetch' on 'ServiceWorkerGlobalScope': 'only-if-cached' can be set only with 'same-origin' mode`

<Details>
Failed to load resource: net::ERR_FAILED
main.js:1 failed to plan trip
(anonymous) @ main.js:1
main.js:1 Object
(anonymous) @ main.js:1
jquery-3.2.1.min.js:2 jQuery.Deferred exception: Cannot read property 'indexOf' of undefined TypeError: Cannot read property 'indexOf' of undefined
    at c.setItinerariesError (https://gophillygo.org/static/scripts/main.js:1:60186)
    at https://gophillygo.org/static/scripts/main.js:1:49295
    at j (https://code.jquery.com/jquery-3.2.1.min.js:2:29999)
    at k (https://code.jquery.com/jquery-3.2.1.min.js:2:30313) undefined
r.Deferred.exceptionHook @ jquery-3.2.1.min.js:2
service-worker.js:42 TypeError: Failed to execute 'fetch' on 'ServiceWorkerGlobalScope': 'only-if-cached' can be set only with 'same-origin' mode
    at service-worker.js:29
    at <anonymous>
(anonymous) @ service-worker.js:42
Promise.catch (async)
(anonymous) @ service-worker.js:41
Promise.then (async)
(anonymous) @ service-worker.js:23
service-worker.js:43 FetchEvent {isTrusted: true, request: Request, clientId: "88f03957-8695-49a7-9b76-3a6097187387", isReload: false, preloadResponse: Promise, …}
(anonymous) @ service-worker.js:43
Promise.catch (async)
(anonymous) @ service-worker.js:41
Promise.then (async)
(anonymous) @ service-worker.js:23
[...]

PR #1007, opened 2018-03-07
Disable service worker cache

flibbertigibbet:
As a workaround for #1005, do not cache files with the service worker.
Service worker still installs, but does not attempt to use files from cache.

Test by disabling network and reloading page: nothing should load.


flibbertigibbet:
Merging for redeploy.

PR #1012, opened 2018-03-09
Feature/kak/update packages#994

### LINKED ISSUE #994: Update packages
Check for new dependencies and update. Includes JS, python, and ansible role dependencies.

flibbertigibbet:
## Overview

Update dependencies.

## Testing Instructions

 * `vagrant destroy app`
 * Delete contents of `src/node_modules` and `src/bower_components`
 * `vagrant up`
 * User-facing and admin sites should continue to work as expected

Closes #994


deployment/ansible/roles.yml
```diff
@@ -2,13 +2,13 @@
   version: 0.1.0
 
 - src: azavea.java
-  version: 0.2.6
+  version: 0.6.1
```
KlaasH: I'm getting warnings like this:
```
- azavea.opentripplanner (1.0.8) was installed successfully
 [WARNING]: - dependency azavea.git from role azavea.opentripplanner differs from already installed version
(0.1.0), skipping

 [WARNING]: - dependency azavea.java from role azavea.opentripplanner differs from already installed version
(0.6.1), skipping
```

Maybe that's not new--I didn't check what it was doing before.  It seems to be a case of the `opentripplanner` role having a unrestricted dependencies (https://github.com/azavea/ansible-opentripplanner/blob/develop/meta/main.yml#L17-L18) that presumably resolve to newer versions.

flibbertigibbet: If these also come from another role, maybe they can be left out of this file altogether. I'll see what that does.

flibbertigibbet: Can simply omit them here, so will do that. Looks like ansible role versioning support is limited: https://github.com/ansible/proposals/issues/23

flibbertigibbet:
The `gulp-test-dev` task has not been in use for some time; `gulp-test` is the meta-task this project uses for tests. It runs tests for production and builds both production and development, but does not run tests also on development; that was due to timing conflicts with async tasks, as I recall.

In any case, tests in general (`gulp-test`) were broken by this, so reverted the npm package changes.

flibbertigibbet:
build again

flibbertigibbet:
build again

PR #1013, opened 2018-03-13
Disable slider when no origin

### LINKED ISSUE #905: Disable travelshed slider when no origin is set
In Explore mode, when no origin is set, it looks odd to have the travelshed slider saying "within N min", since the value has no bearing on anything and sliding the slider does nothing. Need to figure out a different treatment. 

Rough ideas:
- Disable and fade it. 
- Replace it with a hint like "Choose an origin first".

image

flibbertigibbet:
Set a `disabled` class and the `disabled` property on the explore mode range slider and parent control
whenever the origin field is empty or not set to a location yet.

(Not styled, so it looks the same whether enabled or disabled.)

Closes #905.


tnation14:
build again

Issue #1021, opened 2018-03-29
Investigate posting images with social media share


flibbertigibbet:
Share relevant static image (destination image or blog post image) when sharing to social media.

flibbertigibbet:
Blocked by #1023

lederer:
I interpreted the issue here as unrelated to sharing buttons per se, in which case #1023 need not be a blocker. As long as the page has a legit `<meta property="og:image" … >` tag, the preview image should show up in FB regardless of the sharing mechanism (eg, share button vs. manually pasting the link into FB). I think that was the only real concern.

@flibbertigibbet @jcahail Were your takeaways different?

flibbertigibbet:
We have `og:image` tags set already for the relevant pages, so not sure what else we should be doing.

Was thinking of explicitly setting the image via a share dialog, to hopefully have more control over what the resulting post looks like.

lederer:
Interesting.  [Facebook's Sharing Debugger](https://developers.facebook.com/tools/debug/sharing/) doesn't seem to be finding the `og:image`. Seems to often think it's the logo, even when it's set in the html to something else.

Seems to be using the homepage as the scraped page, instead of the actual detail page itself. I think it may be due to `og:url` being set to the homepage, even for detail pages.

![image](https://user-images.githubusercontent.com/128699/38505908-5a5ace02-3be6-11e8-9000-e50ca67c7572.png)

flibbertigibbet:
I wonder if the incorrect image reference found by the Facebook sharing debugger is possibly is related to the redirects on the canonical URL:

![image](https://user-images.githubusercontent.com/960264/38583005-9d69f098-3cdf-11e8-985a-9acf3ead781c.png)

flibbertigibbet:
Looks like the image tags have a relative path. They should be absolute to resolve.

On current production, they look like:
`<meta property="og:image" content="/media/destinations/57a2720ffafd41649676b1646ed5e90c.jpg.680x400_q85_box-destinations/57a2720ffafd41649676b1646ed5e90c.jpg_crop_detail.jpg">`

flibbertigibbet:
Images seem to be getting parsed correctly now, according to the Facebook graph debugger:

### A place
![facebook_graph_debugger_place_detail](https://user-images.githubusercontent.com/960264/43741387-29b85e04-999c-11e8-94de-0468d5d7c722.png)

### The home page (now using the new image)

[...]

PR #1026, opened 2018-04-13
Feature/kak/multiple destination images#1014

### LINKED ISSUE #959: Ease image sizing upload requirements
Retain a fixed required image ratio for destination and article image uploads in the admin UI, but make the image upload process less onerous.

One option would be to give the user the ability to crop/resize the image to match the required ratio within the admin site. Potential libraries to help with this:
https://pypi.python.org/pypi/django-image-cropping
https://github.com/matthewwithanm/django-imagekit

### LINKED ISSUE #1014: Allow adding multiple images for a destination
Currently destinations and events have exactly two images associated with them.

Allow administrators to upload additional `wide_image` sized images for a single destination or event.

Supports displaying multiple place images, as implied by the carousel for the places detail wireframe for the mobile app.
image

flibbertigibbet:
## Overview

Add extra images fields to destinations and events. Adds cropping widget to admin interface and removes aspect ratio restriction on image file uploads, which is no longer needed, as the images are cropped instead. The extra image files are cropped to both the wide and narrow image sizes.

Displays extra images, if any, in a carousel with the primary wide image for a destination or event, if present. If destination or event has no extra images, continues to display single wide image still.

### Demo

Events carousel:
output

Destinations required images with cropping tool:
image

Inline admin extra images:
image

### Notes

It's necessary to save the admin form after uploading an image file and go back into the form to see the cropping widget for the new image.

## Testing Instructions

 * Reprovision app server: `vagrant provision app`
 * Cropping images should work as expected for articles, destinations, and events
 * Uploading extra images should work as expected
 * Detail views for events and destinations should display extra images in a resizable carousel
 * Check `extra_images` and `extra_wide_images` [JSON](http://localhost:8024/api/destinations/search?text=)

Closes #1014
Closes #959 
Connects #1021


KlaasH:
I read through the code and tried it out a bit.  Looks good.

I was initially getting two copies when adding an image, but it seems like running collectstatic and clearing cache made that go away.  Which I wasn't expecting, but that'll work.

A couple small things:
- On Destinations, the images are named by their path (`/media/destinations/UUID.png`), but for Events they get the name "ExtraEventPicture object"

- A usability thing rather than a bug: since you can't crop on upload but have to add the pictures, save the Event or Destination, then go back and edit again to do the cropping, it might be helpful to put some text in the "EXTRA EVENT PICTURES" section to note that users should use "Save and continue editing" to see the cropping options for new pictures.  Not sure how hard it is to add helper text into an admin section, but if people are in the habit of just clicking "Save" they could end up back and the list and not realize until they see it on the site that their image was cropped automatically.

flibbertigibbet:
Thanks for the notes! Should be ready for another look.

Was able to put in a fix for #1029 here that also fixes the duplicate images odd behavior.

Here's what the added help text look like:
image

flibbertigibbet:
Created #1034.

PR #1035, opened 2018-04-20
Update README

maurizi:
There were seemingly no variables that needed to be updated for a development environment.
I checked with @KlaasH and he didn't have to modify any variables in `development_template` either.


flibbertigibbet:
`otp_process_mem` in particular would be good for users to check before firing up the VMs, and sometimes `otp_host` might be useful, for testing against a different OTP instance. (The `otp` VM separately has memory set as well as the OTP process.)

Noting these would probably be good, or at least the default memory consumption of the app in development, since it's not small.

PR #1040, opened 2018-04-26
Replace routing library and fix browser history infinite loop

### LINKED ISSUE #1015: Unable to go back from map page after using 'destination=...' parameter
Steps to reproduce:

 - Got to `https://gophillygo.org/?destination=39.952691343216124%2C-75.14891230525465&destinationText=National%20Constitution%20Center`
 - The URL will be changed to `https://gophillygo.org/?origin=&originText=&mode=TRANSIT%2CWALK&maxWalk=482802&wheelchair=false&bikeOptimize=GREENWAYS&arriveBy=false&dateTime=&destination=39.9526913%2C-75.1489123&destinationText=National%20Constitution%20Center`
 - Attempt to use the browser Back button to leave the page - you will not go to your previous page but instead be forced to remain on the current page, with the map location re-centered.

This is likely due to using `Location.assign()` when we should be using `Location.replace()`

### LINKED ISSUE #1024: Replace routing library
The JS routing library in use for this project [is going to be rewritten by the author](https://github.com/krasimir/navigo/issues/172#issuecomment-377742161).

 - [ ] Replace `navigo`, perhaps with the [riot router](http://riotjs.com/api/route/) or dependency-free [director](https://github.com/flatiron/director).
 - [ ] Push browser history explicitly on user route changes

Currently the app does not explicitly track whether route changes were user initiated state changes that should be pushed to browser "back" history, or made to keep the URL in sync with the form controls automatically, making it difficult to control the browser history as needed for #1015.

maurizi:
## Overview

Replaces the routing library `navigo` with `riot-route`, enabling us to fix an issue with URL routing that resulted in an infinite redirect in the browser history for URLs missing some parameters.

### Notes

The Riot routing library hijacks events for all links that match the base URL (in our case, `/`).
In order to work-around this, I added `target="_top"` to any inter-application links on the home/explore pages, which was suggested as a work-around in the issue: https://github.com/riot/route/issues/25.

## Testing Instructions

 * `vagrant provision app`
 * Issue #1015 should be fixed
 * Other than the above fix, routing should function as before

## Checklist
- [x] No gulp lint warnings
- [ ] ~No python lint warnings~
- [ ] ~Python tests pass~

Fixes #1015 
Fixes #1024

### DIFF
--- a/python/cac_tripplanner/templates/base.html
+++ b/python/cac_tripplanner/templates/base.html
@@ -117,7 +117,7 @@
     <script src="{% static 'scripts/vendor/turf-point-on-line.js' %}"></script>
     <script src="{% static 'scripts/vendor/moment.js' %}"></script>
     <script src="{% static 'scripts/vendor/moment-duration-format.js' %}"></script>
-    <script src="{% static 'scripts/vendor/navigo.js' %}"></script>
+    <script src="{% static 'scripts/vendor/route.js' %}"></script>
     <script src="{% static 'scripts/vendor/slick.js' %}"></script>
 
     <script src="{% static 'scripts/main/cac/cac.js' %}"></script>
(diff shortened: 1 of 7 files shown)


maurizi:
@KlaasH This is ready for another look. I was able to remove the `target="_top"` work-around in https://github.com/azavea/cac-tripplanner/pull/1040/commits/b0a2aac88c8ddb7bfbc7b60099492cf2f5fae9fc

PR #1047, opened 2018-05-29
Expand nginx configuration for admin editing

flibbertigibbet:
Fix `413: Request entity too large` nginx errors when uploading images in admin interface.
Expand rule set allowing large uploads on destination add to also include destination edit
and event add or edit.


maurizi:
LGTM

PR #1048, opened 2018-05-30
Feature/kak/record user flags#1002

flibbertigibbet:
## Overview

Opening in favor of #1041.


python/cac_tripplanner/destinations/migrations/0043_auto_20180506_2103.py
```diff
@@ -0,0 +1,78 @@
...
+            name='EventUserFlags',
+            fields=[
+            ],
+            options={
+                'verbose_name': 'Event User Flag Summary',
+                'proxy': True,
+                'verbose_name_plural': 'Event User Flags Summary',
+                'indexes': [],
+            },
+            bases=('destinations.destination',),
```
flibbertigibbet: Deleting and recreating this migration results in this same `destination` base, so it's not stale.

flibbertigibbet:
build again

python/cac_tripplanner/destinations/admin.py
```diff
@@ -92,5 +98,22 @@ def make_unpublished(self, request, queryset):
     make_unpublished.short_description = 'Unpublish selected events'
 
 
+class AttractionUserFlagsAdmin(admin.ModelAdmin):
+
+    list_display = ('name', 'been', 'want_to_go', 'liked', 'not_interested',)
+    readonly_fields = ('name', 'been', 'want_to_go', 'liked', 'not_interested',)
+    ordering = ('name',)
+    list_display_links = None
```
maurizi: Not strictly necessary, as `has_delete_permission` prevents it from working, but adding `actions = None` will remove the actions dropdown with the "Delete selected ..." action from the page entirely.

python/cac_tripplanner/destinations/models.py
```diff
@@ -58,6 +62,47 @@ def __unicode__(self):
...
+
+    # generic foreign key to abstract Attraction model
+    # see: https://docs.djangoproject.com/en/1.11/ref/contrib/contenttypes/#generic-relations
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.PositiveIntegerField(null=False, db_index=True)
+    attraction = GenericForeignKey('content_type', 'object_id')
+    is_event = models.BooleanField(default=False, db_index=True)
+    timestamp = models.DateTimeField(default=now, editable=False, db_index=True)
+    user_uuid = models.UUIDField(editable=False, db_index=True)
+    flag = models.CharField(choices=UserFlags.CHOICES, max_length=32, db_index=True)
```
maurizi: The `flag` field needs to have `blank=True` set to allow for unsetting the flag. Without it, unsetting the flag causes an uncaught `ValidationError`.

python/cac_tripplanner/destinations/views.py
```diff
@@ -378,16 +379,87 @@ def get(self, request, *args, **kwargs):
...
+            if fld in json_data:
+                user_flag_data[fld] = json_data.get(fld, None)
+            else:
+                missing = True
+                missing_fields_msg += fld + ', '
+        if missing:
+            return return_400(missing_fields_msg, 'Missing required value')
+
+        # parse string JSON value to boolean
+        user_flag_data['is_event'] = user_flag_data['is_event'] == 'true'
```
maurizi: The `is_event` property is already coming through as a boolean value, not a string, so this line needs to be removed or all flags are interpreted as Destinations.

python/cac_tripplanner/destinations/models.py
flibbertigibbet: There is an explicit option for `none` in the `CHOICES set, which is an empty string, so still needing `blank` set here surprises me somewhat.

flibbertigibbet:
Updated.

PR #1051, opened 2018-06-01
Remove fetch handler from service worker

maurizi:
## Overview

This handler was causing all network requests to be done twice, the primary
symptom of which was that new items added on the Admin site would be
duplicated.

## Testing Instructions

 * Ensure that when creating a new Event, Destination or Article from the Admin site, only one item is created
 * Use `ngrok.io` to proxy your local instance over `https://`, and verify from a phone that the application can continue to be installed as a progressive web app to your home screen

## Checklist
- [x] No gulp lint warnings
- [x] No python lint warnings
- [x] Python tests pass


flibbertigibbet:
Merging for redeploy.

PR #1062, opened 2018-08-03
Feature/kak/social sharing improvements#1021

flibbertigibbet:
## Overview

Improve social sharing meta tags, so that images for places and events will be used in the thumbnail cards generated by the sites.

### Demo

A destination:
cac_sharing_full_path_after

Home page:
cac_social_sharing_default_image

Footer links:
image

## Testing Instructions

 * Check the meta tags on the top level page; image URL should be a working link to new logo
 * Image URLs on the details pages (articles, destinations, and events) should be full, working URL
 * Footer links to privacy policy and TOS should work

Closes #1021 
Closes #1028 
Closes #1060

PR #1064, opened 2018-08-06
Fix image sharing URLs

flibbertigibbet:
Build absolute path using production domain.
(Sharing will not work in development anyways.)

Fixes an issue with #1062 where the sub-path to the detail page was also included in the image URL.


flibbertigibbet:
Merging for redeploy

PR #1067, opened 2018-08-08
Bugfix/kak/return event times from search#1066

### LINKED ISSUE #1066: Event dates from search endpoint return wrong day for some times
`end_date` on events returned from https://gophillygo.org/api/destinations/search?text= returns the next day when the time portion of the end date is less than five hours from midnight (likely a UTC offset issue).

flibbertigibbet:
## Overview

Fix JSON-serialized dates by applying server timezone. Also includes python minor package updates.

## Testing Instructions

 * Create an event at http://localhost:8024/admin
 * Check event dates at http://localhost:8024/api/destinations/search?text=
 * Times should be included and be correct EST times (same as set in the admin interface)
 * Check event on homepage and details page; dates should still be correct there too

Fixes #1066


flibbertigibbet:
Small fix; will merge to redeploy

Issue #1080, opened 2018-08-30
Upgrade staging environment to PostgreSQL 9.4


hectcastro:
- Execute an in-place upgrade via the console
- Update the PostgreSQL version strings in CloudFormation (just to maintain consistency)
- Upgrade PostGIS using

```sql
ALTER EXTENSION POSTGIS UPDATE TO '...'
```

For all of the specific steps, see: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.html

rbreslow:
@hectcastro Do you want to schedule time to pair on this next week?

hectcastro:
Might be good to defer until after Kathryn gets a chance to run through the app in development to ensure that there aren't any regressions. I added her as a reviewer to the development env PR so that she can start that process once it is in a good state.

PR #1081, opened 2018-09-04
Upgrade development environment to PostgreSQL 9.4

### LINKED ISSUE #1079: Upgrade development environment to PostgreSQL 9.4
The production database is currently running PostgreSQL 9.3, which is losing support in AWS RDS on Sept 5, 2018. On or nearly after Nov 1, 2018, AWS will auto-upgrade any remaining RDS instances still on 9.3 to the latest 9.4 bugfix release. This task is to upgrade the DB to at least the latest 9.4 bugfix release before the Nov 1 deadline. If time/budget allows, we could explore bumping to 9.6 or higher.

More complete details from AWS:

> As part of the retirement plan for Amazon RDS PostgreSQL 9.3, we have disabled creation of any new PostgreSQL 9.3 instances using the AWS console. Starting September 5th, 2018, we will start to disable new PostgreSQL 9.3 instance creation using the CLI/API. We anticipate the process to complete by September 30th, 2018. You can continue to do PostgreSQL 9.3 restores and read replica creations until October 31st, 2018.  In November 2018, your remaining 9.3 instances will be force upgraded to the most recent minor version of PostgreSQL 9.4. Once this occurs, any instance restored from a snapshot of a deprecated version will automatically and immediately be upgraded to a currently supported version. 

- https://www.postgresql.org/support/versioning/
- https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.PostgreSQL.html
- https://aws.amazon.com/support
- https://aws.amazon.com/rds/faqs/ [search for “guidelines for deprecating database engine versions”]

rbreslow:
## Overview

* Refactor cac-tripplanner.database role
* Utilize azavea.postgresql and azavea.postgis ansible modules
* Bump references for PostgreSQL 9.3 and PostGIS 2.1.5

Fixes #1079 

## Demo

image

## Testing Instructions

* Clone the repository. 
* Download the `.osm` and GTFS data from the `cleanair-otp-data` S3 bucket, to your  local `otp_data` folder. 
* Run `vagrant-up.sh`.
* Visit http://localhost:8024/ and http://localhost:9090/


deployment/ansible/group_vars/development_template
```diff
@@ -9,7 +9,7 @@ otp_process_mem: 5G
 # for cac_secrets
 allowed_hosts: ['127.0.0.1', 'localhost']
 internal_ips: ['0.0.0.0', '127.0.0.1']
-postgis_version: [2, 1, 5]
+postgis_version: [2, 4, 4]
```
hectcastro: The version of PostGIS on Amazon RDS for PostgreSQL is 2.1.8. We should try to match that in development as well (at least 2.1.x).

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#PostgreSQL.Concepts.General.FeatureSupport.Extensions.94x

It is possible that there isn't a specific package for 2.1.8—if that's the case, we've worked around it in the past by building the appropriate version of PostGIS from source:

https://github.com/azavea/cicero/blob/develop/deployment/ansible/roles/cicero.postgresql/tasks/install-postgis-from-source.yml

deployment/ansible/roles/cac-tripplanner.database/meta/main.yml
```diff
@@ -1,3 +1,5 @@
 ---
   dependencies:
+    - { role: "azavea.postgresql" }
```
hectcastro: Good job porting this over to the common Ansible modules.

deployment/ansible/roles/cac-tripplanner.database/vars/main.yml
```diff
@@ -0,0 +1,15 @@
+---
```
hectcastro: I think that it would be better to surface this in `group_vars/development_template` vs. `vars` so that they are more visible. The `postgres_*` ones in `all` possibly, since they are needed to provision the `app` virtual machine.

rbreslow: @hectcastro Does ansible interact with any of the production resources?

hectcastro: Yeah, Ansible applies `app.yml` and `otp.yml` to EC2 instances via Packer as part of the AMI creation process: https://github.com/azavea/cac-tripplanner/blob/develop/deployment/packer/cac.json#L104-L125

But the database setup is all RDS, so `cac-tripplanner.database` won't get used in that context.

rbreslow: Since before the upgrade, the `postgresql_*` variables were only scoped to `cac-tripplanner.database`, I don't have a strong desire to expand the scope. 

Originally, the `postgresql_*` variables were in `cac-tripplanner.database/defaults/main.yml`.

After including the `azavea.postgresql` and `azavea.postgis` modules, I moved them to `cac-tripplanner.database/vars/main.yml`.

This is what we do in HunchLab:

https://github.com/hunchlab/hunchlab/blob/develop/deployment/ansible/roles/hunchlab.database/vars/main.yml.

As an alternative solution, maybe we could define the variables in the `database.yml` playbook?

I experimented with surfacing these variables in `group_vars/development_template` and there was a conflict with an existing `postgis_version` variable.

rbreslow: I missed what you were referring to in the `cac-tripplanner.app/defaults/main.yml` file. 

Maybe as a compromise, we could expand the scope for `postgres_*`, but not `postgresql_*` and `postgis_*`?

rbreslow: After sleeping on it, I think an approach of putting `postgres_*` in `all` and `postgresql_*` in `development` works.

See this example from Cicero: 

https://github.com/azavea/cicero/blob/774967531d832e8665906ba8520a623fb1263c22/deployment/ansible/group_vars/all#L26

There's still the issue of the conflicting `postgis_version` var. It looks like this stuff is supposed to be written to `/etc/cac_secrets`, where it's then parsed into an object in a Python script. But, I can't find any code that actually saves it, unless this is an Ansible built in.

hectcastro: I like the approach from Cicero (and also MMW) because it has historically made it easier for folks who aren't worried about provisioning details day-to-day see the versions of dependencies at a glance.

The conflict with `postgis_version` is likely something that we should try to resolve explicitly. It might require renaming the variable that takes the tuple vs. the one used by the role. I can't take a closer to provide a better response right now, but I can take a look a little later to see if there is a good way to resolve the conflict.

deployment/ansible/group_vars/development_template
```diff
@@ -25,3 +24,15 @@ use_s3_storage: false
 default_admin_username: 'admin'
 default_admin_password: 'admin'
 default_admin_email: 'systems+cac@azavea.com'
+
+# for database
+postgresql_version: 9.4
+postgresql_package_version: "9.4.19-1.pgdg14.04+1"
+
+postgis_version: 2.1
+postgis_package_version: "2.1.8+dfsg-5~97.git43a09cc.pgdg14.04+1"
```
hectcastro: Despite the package name here, after provisioning the `database` virtual machine from scratch, it is installing PostGIS 2.4.4:

```sql
cac_tripplanner=# SELECT PostGIS_full_version();
                                                                              postgis_full_version
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 POSTGIS="2.4.4 r16526" PGSQL="94" GEOS="3.4.2-CAPI-1.8.2 r3921" PROJ="Rel. 4.8.0, 6 March 2012" GDAL="GDAL 1.10.1, released 2013/08/26" LIBXML="2.9.1" LIBJSON="0.11.99" RASTER
(1 row)
```

We may have to build the extension from source if we aren't able to get PostgreSQL to use the correct version.

flibbertigibbet:
Apart from the versioning differences @hectcastro noted above, the full graph build for the OTP VM now hangs for me on the `Copy Built OTP Graph to Host` task, which does not seem to be related to the changes here, but is new.

deployment/ansible/group_vars/development_template
```diff
@@ -25,3 +25,11 @@ use_s3_storage: false
 default_admin_username: 'admin'
 default_admin_password: 'admin'
 default_admin_email: 'systems+cac@azavea.com'
+
+postgresql_version: 9.4
+postgresql_package_version: "9.4.19-1.pgdg14.04+1"
```
hectcastro: I think that we should glob the patch version here: `9.4.*.pgdg14.04+1`

flibbertigibbet:
build again

flibbertigibbet:
I fixed the vagrant hostmanager plugin on Jenkins and destroyed the VMs to let them rebuild, but the database VM build is sill failing on Jenkins. Do the `test` ansible `group_vars` also need to be updated?

JS tests are failing now for me locally; I created #1084 for that. So once the database setup issues are addressed on Jenkins, I expect the build to fail on the JS tests.

rbreslow:
@flibbertigibbet My `group_vars` mirrors everything in the feature branch. The only thing I did was to copy `development_template` to `development`.

I don't believe there was anything that would cause the `app` VM to have the `psql` command before I made this PR.  I am going to test on `develop` to double check.

As far as upgrading staging and production, we have https://github.com/azavea/cac-tripplanner/issues/1080 to upgrade staging to 9.4. We should be in a good place to attempt this, although the steps will be different on RDS. There is a fixed version of the PostGIS extension that comes preinstalled when you use the `9.4` engine on AWS. Also, upgrading an existing database should require a query to alter the PostGIS version.

rbreslow:
Build #601 failed because it was trying to install a version of PostgreSQL 9.6. Most likely, there is an outdated `group_vars` file on Jenkins causing this behavior.

deployment/ansible/group_vars/development_template
```diff
@@ -25,3 +25,11 @@ use_s3_storage: false
 default_admin_username: 'admin'
 default_admin_password: 'admin'
 default_admin_email: 'systems+cac@azavea.com'
+
+postgresql_version: "9.4"
```
hectcastro: These PostgreSQL overrides aren't being seen by CI because `VAGRANT_ENV=TEST` exists in the Jenkins configuration. This makes the Vagrant configuration for Ansible use the `group_vars/test` (in addition to `all`) file for variables override, but not `development`.

https://github.com/azavea/cac-tripplanner/blob/1f9751837917c0bc344307cf74231445fe00e4da/Vagrantfile#L81-L85

rbreslow:
Builds are passing after adding `postgresql_*` and `postgis_*` variables to `group_vars/test`.

PR #1103, opened 2019-05-21
Switch carousel library

### LINKED ISSUE #1102: Image carousels broken
The `slick-carousel` image carousels are no longer rendering properly since #1101 in either place they are used.

Carousels display in two places:
 - on destination and event detail pages, if there are multiple images, they are in a carousel
 - on mobile, if there are multiple itineraries for a trip plan, they are in a carousel below the map

On the place/event detail page, the first image displays as expected, but subsequent images are offset incorrectly, and also the navigation dots. The trip itineraries on mobile render as a list instead of in a carousel, as expected.

The jQuery update seemed a likely cause; however, reverting did not fix the issue.

flibbertigibbet:
## Overview

Replace slick-carousel with tiny-slider.

Updates to the JavaScript build broke use of the carousel library. The exact cause was unclear, and the `slick-carousel` library looks to be effectively unsupported, so here it is replaced with another carousel library, `tiny-slider`, which does not have external dependencies and has recent updates.

The new carousel should look and function just as the old one did.

### Demo

image

Mobile itineraries:
image

Itinerary cards resize height on swipe (using `autoHeight`):
image

### Notes

If the map page gets loaded at a larger width than the mobile breakpoint, then is resized to be narrower than the mobile breakpoint, the dots don't load under the itinerary carousel, although the carousel otherwise renders and functions as expected. This should be fine, as it is unlikely in normal use that a user would resize their desktop browser to be mobile-sized, and the functionality is still present.

## Testing Instructions

 * Add multiple images to at least one destination 
 * Add at least one event with multiple images
 * Carousel should render and autoplay as expected on the place/event detail pages
 * The single image for places/events without multiple images should still display as expected
 * Directions on mobile (<480px width) should display itineraries in carousel as expected

## Checklist
- [x] No gulp lint warnings
- [x] No python lint warnings
- [x] Python tests pass

Fixes #1102


src/package.json
```diff
@@ -18,8 +18,7 @@
     "lodash": "~4.17.11",
     "polyline-encoded": "~0.0.8",
     "riot-route": "~3.1.4",
-    "spinkit": "~1.2.5",
```
lederer: Was spinkit intended to be removed? If so, there are still references to it in the code (eg, search for `sk-spinner`).

flibbertigibbet:
The event images aren't displaying. I'll fix that.

src/package.json
flibbertigibbet: No, it was not. Good catch!

flibbertigibbet:
build again

flibbertigibbet:
build again

flibbertigibbet:
build again

flibbertigibbet:
build again

flibbertigibbet:
build again

flibbertigibbet:
The tests and linting pass locally. This is ready for another look.

lederer:
> flash of all the images as the page loads

The `lazyload` option could potentially help?

flibbertigibbet:
Created #1105.

flibbertigibbet:
Lazy loading images does not fix the initial flash, but does introduce an issue where the opacity for all the images is reduced to 60% (even when not animating).

PR #1117, opened 2019-07-12
Upgrade to Xenial

### LINKED ISSUE #995: Update deprecated ansible usage
Provisioning `otp` with ansible 2.4.3 warns:

```
[DEPRECATION WARNING]: The use of 'include' for tasks has been deprecated. Use 
'import_tasks' for static inclusions or 'include_tasks' for dynamic inclusions.
 This feature will be removed in a future release. Deprecation warnings can be 
disabled by setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: include is kept for backwards compatibility but usage is
 discouraged. The module documentation details page may explain more about this
 rationale.. This feature will be removed in a future release. Deprecation 
warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```

### LINKED ISSUE #1100: Upgrade to Ubuntu Xenial
~~HVM [AMIs for Ubuntu](https://cloud-images.ubuntu.com/locator/ec2/) Trusty are no longer available, blocking redeploy.~~

Attempting to use a paravitualied AMI instead results in the error:
```
Virtualization type 'hvm' is required for instances of type 'm4.large'
```

Upgrade the AMI and vagrant box OS versions to Xenial. Fix ansible roles and service scripts for the upgrade, as done [here](https://github.com/azavea/cicero/commit/ce6d0964ec44736cb4dd77f746744ec140790c92).

flibbertigibbet:
## Overview

Upgrade ansible roles to support using Ubuntu Xenial (16.04) as the OS for all three VMs. Mostly, this meant switching services to use `systemd` instead of `upstart`.

This is to support upgrading to python 3 and then Django 2.x later; here, python 2 is still in use.

Also makes some minor upgrades to ansible syntax, to resolve deprecation warnings.

### Notes

I tried using the gunicorn ansible role instead of installing it within the app role, but ran into issues, so instead upgraded the gunicorn role in place for `systemd`. As I've added support for the `timeout` parameter to the gunicorn ansible role, it's possible we could use it in future instead.

For the OTP VM, it was necessary to explicitly install and use python 2, in order to continue to validate GTFS feeds, as the feed validator library does not yet support python 3: google/transitfeed#412

For the python 3 upgrade, we can handle the OTP machine by one of:
 - removing GTFS validation entirely (if using the [feed fetcher library](https://github.com/azavea/gtfs-feed-fetcher), this validation has already happened on download)
 - leaving the machine on python 2 until the validator project upgrades
 - upgrading the validator to python 3 ourselves, which might necessitate forking the project

I would be inclined towards the first option, as the feeds are currently validated outside the VM as part of the current graph update process, and because validating again slows the build process.

## Testing Instructions

 * destroy and rebuild all VMs
 * provisioning should complete as expected
 * no ansible deprecated usage warnings should remain for roles that are defined in this project (external roles still have some warnings, particularly `postgresql` and `postgresql-support`)
 * site should still function as expected
 * `services should start on reboot (site should still work after `vagrant reload`)

Closes #1100 
Closes #995


flibbertigibbet:
I'll wait for #1119 to go in before destroying the VMs on CI and rebuilding this PR, as it changes the VM base image.

maurizi:
The `database` VM failed to provision with the follow error:
```
TASK [azavea.postgresql : Install PostgreSQL] **********************************
failed: [database] (item=[u'postgresql-9.6=9.6.*-2.pgdg14.04+1', u'postgresql-server-dev-9.6=9.6.*-2.pgdg14.04+1']) => {"cache_update_time": 1563477629, "cache_updated": false, "changed": false, "item": ["postgresql-9.6=9.6.*-2.pgdg14.04+1", "postgresql-server-dev-9.6=9.6.*-2.pgdg14.04+1"], "msg": "'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"     install 'postgresql-9.6=9.6.*-2.pgdg14.04+1' 'postgresql-server-dev-9.6=9.6.*-2.pgdg14.04+1'' failed: E: Version '9.6.*-2.pgdg14.04+1' for 'postgresql-9.6' was not found\nE: Version '9.6.*-2.pgdg14.04+1' for 'postgresql-server-dev-9.6' was not found\n", "rc": 100, "stderr": "E: Version '9.6.*-2.pgdg14.04+1' for 'postgresql-9.6' was not found\nE: Version '9.6.*-2.pgdg14.04+1' for 'postgresql-server-dev-9.6' was not found\n", "stderr_lines": ["E: Version '9.6.*-2.pgdg14.04+1' for 'postgresql-9.6' was not found", "E: Version '9.6.*-2.pgdg14.04+1' for 'postgresql-server-dev-9.6' was not found"], "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\n", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information..."]}
        to retry, use: --limit @/home/local/AZVA-INT/mmaurizi/src/cac-tripplanner/deployment/ansible/database.retry
```

flibbertigibbet:
The `development` `group_vars` I had shared with you separately contained:

```
postgresql_version: "9.5"
postgresql_package_version: "9.5.*"
postgresql_support_libpq_version: "*"
```

Where those set?

These variables go in `development` rather than `all`, as the production database is usually on a slightly different version.

maurizi:
D'oh, I'd updated to the newest Graph but hadn't updated my `group_vars`. Provisioning succeeded after doing so.

I _did_ see one deprecation warning from Ansible for a role defined in this project:
```
TASK [cac-tripplanner.otp-data : Copy Built OTP Graph to Host (develop)] *******
[DEPRECATION WARNING]: Using tests as filters is deprecated. Instead of using 
`result|changed` instead use `result is changed`. This feature will be removed 
in version 2.9. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
skipping: [otp]
```

I'm testing out the site now, but encountered an issue accessing `gunicorn` after runnning `vagrant reload`, and investigating seems to chow that the service was not able to start.

flibbertigibbet:
Thanks for testing, @maurizi. I've removed the file share mounting parameters from the `Vagrantfile`, which has fixed the NFS permissions issues I was seeing using a Linux host, leading to `nginx` and the `gunicorn` server to not start properly. Hopefully the change doesn't cause issues for other hosts. I've also addressed the ansible deprecation warning.

flibbertigibbet:
build again

flibbertigibbet:
build again

flibbertigibbet:
build again

flibbertigibbet:
The Jenkins build is passing now. @maurizi could you please take another look?

python/cac_tripplanner/requirements.txt
```diff
@@ -0,0 +1,16 @@
...
+django-storages==1.7.1
+easy-thumbnails==2.6.0
+gunicorn==19.9.0
+Pillow==6.1.0
+psycopg2-binary==2.8.3
+pytz==2019.1
+PyYAML==5.1.1
+requests==2.22.0
+troposphere==1.8.1
+majorkirby==0.2.1
```
maurizi: :tada: great to get this out of Ansible and into a standard `requirements.txt`!

flibbertigibbet:
build again

PR #1122, opened 2019-07-26
Fix missing Indego stations on map

### LINKED ISSUE #1120: Indego bike layer does not display
Due to the lack of a CORS header on the bike share GeoJSON, the Indego bike station map layer does not display when checked in the layer switcher.

image

This may be addressed by changing the URL for the bike share station layer from https://www.rideindego.com/stations/json/ to another source that supplies CORS headers. If another URL cannot be found, we could proxy the response through the web server, and add the CORS header ourselves.

flibbertigibbet:
## Overview

Note that this branch is based on the Xenial upgrade PR branch.

Fix the Indego stations not showing on the map by changing the endpoint to the one maintained by the city, instead of the Indego-maintained endpoint.

### Demo

image

### Notes

Both endpoints resolve to the same CloudFront endpoint. However, the Indego proxy no longer adds CORS headers, while the city endpoint does.

The city endpoint results in a CORB warning on Chrome, but loads without error:
image

## Testing Instructions

 * Go to map page
 * Open the layer switcher and check 'Bike Share Locations'
 * Blue map markers with the Indego logo should appear where the stations are
 * Map marker popups should contain the expected station address and availability information

Fixes #1120.

### DIFF
--- a/deployment/ansible/group_vars/all
+++ b/deployment/ansible/group_vars/all
@@ -14,7 +14,7 @@ packer_version: "1.0.2"
 python_version: "2.7.*"
 
 nodejs_version: 10.16.0
-nodejs_npm_version: 6.9.0
+nodejs_npm_version: 6.10.2
 
 pip_version: 19.1.*
 virtualenv_version: 16.5.0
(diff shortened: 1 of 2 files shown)


flibbertigibbet:
build again

flibbertigibbet:
build again

flibbertigibbet:
build again

PR #1141, opened 2019-09-04
Update django-wpadmin install location

flibbertigibbet:
Follow up to #1139 to correct the `django-wpadmin` library install version.


flibbertigibbet:
Trivial, merging

PR #1154, opened 2019-10-04
Change directions link to "Map" on tour cards

### LINKED ISSUE #1152: Change directions link to "Map" for tour cards
This was part of issue #1146 but not included in PR #1150. Per my comment there:
> The other task from the issue is 
>> "'Directions' card link should be labelled 'Map' instead for (non-routable) event map overview" 
> 
> This is the same thing as the link on the Tour Detail page, but there's currently only a back-end function for composing it. To get it working in the front-end version of the templates, we need to either write a front-end tag to reproduce the behavior or stuff that link into the API response. I was leaning toward the latter, since it's a fairly involved bit of business, but I'm wondering what you think would be best.

After looking at it a little more, I still think composing the link once, on the back-end, and sending it along with the API response seems like the way to go.

KlaasH:
## Overview

Changes the text of the link on Tour cards from "Directions" to "Map".

Per discussion on issue #1152, I left this out of PR #1150 because I thought the logic to recreate the URL that's happening on the back-end for the Tour detail page needed to be reproduced on the front-end.  Turns out that's already handled by the existing directions controller logic, so only the name change is needed.

### Demo

image

## Testing Instructions

- Go to [the home page](http://localhost:8024/). Tour cards should say "Map", everything else should still say "Directions".
- Click the "Tours" tab to filter to only tours, which also replaces the Django-rendered card with a Javascript-rendered one.  It should look the same.

## Checklist
- [x] No gulp lint warnings
- [x] No python lint warnings
~- [ ] Python tests pass~ (no changes, didn't check)

Resolves #1152

### DIFF
--- a/python/cac_tripplanner/templates/home.html
+++ b/python/cac_tripplanner/templates/home.html
@@ -167,7 +167,9 @@
                             <a class="place-card-action place-action-go"
                                 data-destination-id="{{ directions_id }}"
                                 data-destination-places="{{ place_ids }}"
-                                href="#">Directions</a>
+                                href="#">
+                                {% if destination.is_tour %}Map{% else %}Directions{% endif %}
+                                </a>
                             {% endif %}
                             <a class="place-card-action place-action-details"
                                {% if destination.is_event %}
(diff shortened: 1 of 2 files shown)


flibbertigibbet:
build again

PR #1156, opened 2019-10-07
Feature/kak/events sidebar#1126

### LINKED ISSUE #1126: Show multi-destination events on map page
When the user selects an event from the typeahead autocomplete, present the result as a specialized list in the map sidebar, with a header that includes the event name. Show map markers for the event locations, but do not plan a route to or between them; reordering map sidebar cards, if possible, has no effect.

Map sidebar cards for the event destinations contain directions links that go to directions from the current origin to the selected destination. If the user clicks the back button after going to directions to an event destination, the browser returns to the map page for the event. Permalinking should work for the event map page.

image

flibbertigibbet:
## Overview

Implement map page for events.

Do not route to event destinations, but present them as cards in the sidebar, similar to the 'tour' sidebar, but with a slightly different header. On the map, show markers for the event destinations.

### Demo

image

### Notes

This uses the same sidebar control and elements as the `tour` sidebar, which is why it also has a fixed height as a temporary fix for scrolling.

## Testing Instructions

 * Clicking an event card from the home page should go to event map view
 * Clicking an event card from the map sidebar 'explore' cards should go to event map view
 * Selecting an event from the destination autocomplete options should go to the event map view
 * Event destination map markers should display and map should zoom to fit them all
 * Clicking a destination card from the event sidebar should go to directions to that place
 * Clicking the browser back button from destination directions should return to event map view
 * 'More info' destination card button should link to place detail page
 * On sidebar card hover, 'directions' button should highlight and map marker circle should appear

Closes #1126


flibbertigibbet:
It looks like the sidebar only goes blank if the destination was set, then cleared from the form while viewing the event. Going to the event without a destination set shows the sidebar. I'd like to break that out to a separate issue to clean up; going to a tour without an origin or clearing the origin, I imagine, should also still show the destination list, although the tour is routable.

flibbertigibbet:
After looking at the origin clear issue further, I've pushed a commit to preserve the sidebar contents when in tour mode. The state of tours (not events) without an origin set probably needs further modification in another task.

flibbertigibbet:
build again

KlaasH:
This is working better, but I'm still seeing some issues.

In terms of the items listed above, I'm seeing:
- [x] Clicking an event card from the home page should go to event map view
- [x] Clicking an event card from the map sidebar 'explore' cards should go to event map view
- [x] Selecting an event from the destination autocomplete options should go to the event map view
  (on the Directions view, not from the home page)
- [ ] Event destination map markers should display and map should zoom to fit them all
  (not showing for unpublished destinations)
- [x] Clicking a destination card from the event sidebar should go to directions to that place
  (clicking the card does nothing, but clicking the Directions link works)
- [x] Clicking the browser back button from destination directions should return to event map view
- [x] 'More info' destination card button should link to place detail page
- [x] On sidebar card hover, 'directions' button should highlight and map marker circle should appear

Here's a gif that shows some of the missing pieces above.  I made it after clearing site data and hard-refreshing the home page.  It seems like it dropped some frames, but hopefully it's complete enough to make sense.
Peek 2019-10-10 16-57

The part that's most confusing is where I was trying to show that hovering over the card for an unpublished destination makes a blue circle appear on the map but there's no marker for it.

Also appearing in there are a few other issues:
- Clicking the "Map" link from the Event details page always gets me the view showing the final destination rather than the event map view.
- The carousel for the Tour images isn't always loading. I don't know if that's new or not, but I hadn't seen it before.
- There's no scrollbar for the event side bar, so some of the destinations are off the screen with no way to get to them.

> It looks like the sidebar only goes blank if the destination was set, then cleared from the form while viewing the event. Going to the event without a destination set shows the sidebar. I'd like to break that out to a separate issue to clean up.

Do you mean "if the origin was set"?  In my original screenshot the destination was set.  But as the gif shows, I'm seeing the blank sidebar without ever having set an origin (or any destination besides the event).  So yeah, I'm not clear on what the distinct piece to be spun out into another issue would be.

flibbertigibbet:
Okay, so to get to the event view with a blank sidebar you've been referring to, you clicked the 'map' link from the event details page. That's some helpful context that was missing previously. And yes, I meant setting and then clearing the origin when I was trying to reproduce what you were referring to.

The scrollbar isn't showing due to @lederer removing the fixed height for the sidebar in the design. The last commit here moves elements so he can follow up to fix the styling for scrolling to work. See #1162.

I haven't seen any issue with the carousel, and this doesn't touch that; the design work may have, however. If so, we can make a new issue for that.

So it sounds like what needs addressing here is:

 - map page link on events detail page isn't implemented yet
 - map markers don't show for unpublished destinations

flibbertigibbet:
Regarding the carousel, do you mean tours showing in the sidebar in tall cards with vertically stacked images? That showed up after #1161. I'll make a design issue for that.

flibbertigibbet:
Fix for #1163 (carousel class renamed in design) pushed here.

lederer:
@flibbertigibbet Thanks for fixing up any collateral damage from my merged polish pass, and for the DOM changes to address #1162. I'll follow up with the rest of the scrolling fix in another polish pass after this merges.

flibbertigibbet:
Okay, this should be ready for another look.

PR #1159, opened 2019-10-09
Enable tours mobile carousel

flibbertigibbet:
## Overview

Enable tours mobile carousel. ~~Do not merge: needs styling fixes.~~ Merging for styling afterwards.

### Demo

image

### Notes

Carousel appears and responds to swipe gestures, but destination cards mostly do not appear, likely due to styling issues.

~~This PR is based on #1157. I will rebase it on `tours` after #1157 has been merged.~~

## Testing Instructions

 * Go to an event or tour map view with a mobile-sized screen
 * Destinations should appear in swipe-able carousel at bottom of screen

Closes #1147

PR #1157, opened 2019-10-08
Feature/kak/drag tour destinations#1112

flibbertigibbet:
## Overview

Allow dragging tour destinations in sidebar to reorder them, which triggers refreshing the route to update the waypoint order and destination.

### Notes

I based this PR on #1126. I'll rebase on `tours` once #1126 has merged.

## Testing Instructions

 * Go to the map page for a tour
 * Destinations in sidebar should have expected drag-and-drop interactions
 * On destination reorder directions should update as expected
 * Event destinations should not be draggable (only tours)

Closes #1112

PR #1161, opened 2019-10-09
Initial polish of tours and multilocation events

lederer:
## Overview

- Home page [Desktop && Mobile]
  - Style tabs
  - Style tour card
- Tour detail page [Desktop && Mobile]
  - Style header
  - Style location cards
- Multilocation event detail page [Desktop && Mobile]
  - Style locations
  - Style per-location times
- Style tour map page [Desktop only]
  - Style location cards
  - Refactor DOM to prepare for mobile carousel

### Demo

image

image

image

image

### Notes

The tour map sidebar scrolling issue is still there. The fix requires a DOM change that I wasn't able to implement…

The tour map sidebar currently contains the following DOM:

```
.directions-results > .directions-list > .tours
```

Essentially we need to remove the intermediate `.directions-list` when in tour mode, so the corresponding DOM would look like:

```
.directions-results > .tours
```

Alternatively, tour mode could replace `.directions-results` entirely with a new tour-mode container, for something like this:

```
.tour-results > .tours
```

Either solution should work, and would just require some additional CSS tweaks from me once it's done.

## Checklist
- [ ] No gulp lint warnings
- [ ] No python lint warnings
- [ ] Python tests pass


lederer:
@flibbertigibbet There are some DOM changes in here that might impact some in-progress PRs (probably #1159; maybe #1157 and #1156).

python/cac_tripplanner/templates/tour-detail.html
```diff
@@ -37,6 +37,7 @@
         <h1><a href="{% url 'home' %}">Tours We Love</a></h1>
         {% include "partials/tour-detail.html" %}
 
+        {% if more_tours %}
```
flibbertigibbet: :+1:

src/app/styles/components/_directions-results.scss
```diff
@@ -32,58 +32,68 @@
...
-        font-size: 1.2rem;
-        padding: 4px 12px 0;
-        text-transform: uppercase;
-    }
+    .tours {
 
-    .tour-header {
-        padding: 4px 12px 0;
-    }
+        // FIXME: forcing height for the scrollbar to work
```
flibbertigibbet: The height isn't set any more, so it looks like this comment can go.

PR #1165, opened 2019-10-11
Feature/kak/tour map no origin#1164

### LINKED ISSUE #1164: Modify tour map view when no origin set
Currently, when viewing the map page for a tour without an origin set, the sidebar is empty and a single marker displays for the last destination as the endpoint.

@lederer what should the map page look like in that state?

flibbertigibbet:
## Overview

Changes the initial tour map page to load directions from the first tour destination when there is no origin set in the directions form.

### Demo

image

### Notes

~~This branch is based on #1157. I'll rebase it on `tours` when #1157 has merged.~~

## Testing Instructions

 * Go to the map page for a tour without an origin set
 * Markers for tour destinations should load on map
 * Directions should display on map, starting at the first destination in the tour
 * Origin field in the sidebar form should remain blank
 * Set the origin in the form
 * Directions should update to route from user set origin to tour destinations

Closes #1164

### DIFF
--- a/src/app/scripts/cac/control/cac-control-directions.js
+++ b/src/app/scripts/cac/control/cac-control-directions.js
@@ -144,8 +144,8 @@ CAC.Control.Directions = (function (_, $, moment, Control, Places, Routing, User
         showSpinner();
         showPlaces(false);
         var tourMode = UserPreferences.getPreference('tourMode');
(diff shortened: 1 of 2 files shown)


flibbertigibbet:
I've fixed up the origin clear issue (good find!) and rebased on `tours`.

PR #1168, opened 2019-10-16
Tour sorting styles and other misc polish

lederer:
## Overview

Style tour destination sorting UI.
Link tour/event map sidebar title to detail page.
Style tours list in map infowindow.
Fix tour sidebar scrolling bug.
Hide tour card carousel dots when showing trip duration.

### Demo

image

image

image


src/app/scripts/cac/control/cac-control-explore.js
```diff
@@ -396,7 +396,7 @@ CAC.Control.Explore = (function (_, $, MapTemplates, HomeTemplates, Places, Rout
                                                 tabControl.isTabShowing(tabControl.TABS.HOME));
         $(options.selectors.placesContent).html(places);
         $(options.selectors.imageCarousel).each(function (index, carouselNode) {
-            tns(Object.assign({container: carouselNode}, Utils.defaultCarouselOptions));
+            tns(Object.assign({container: carouselNode}, Utils.defaultCarouselOptions, {autoplay: false}));
```
flibbertigibbet: :+1:

lederer:
> I'm seeing a scroll bar within the draggable cards on Chrome Linux

Whack. Just pushed a fixup that _should_ address that.

src/app/scripts/cac/map/cac-map-templates.js
```diff
@@ -279,12 +279,17 @@ CAC.Map.Templates = (function (Handlebars, moment, Utils) {
...
-        '<ul class="tour-list">',
+        '<div class="tour-list">',
             '<div class="tour-heading">',
                 '<div class="tour-label">',
                     '{{#if tour.is_event}}Event{{else}}Tour{{/if}}',
                 '</div>',
-                '<h1 class="tour-name">{{ tourName }}</h1>',
+                '<h1 class="tour-name">',
+                    '<a class="tour-name-link" href=',
+                        '"/{{#if tour.is_event}}event{{else}}tour{{/if}}/{{ tour.id }}/">',
```
flibbertigibbet: This should probably `target="_blank"` to open in a new tab, as the other links do.

lederer: The app generally only opens _external_ pages in new tabs, whereas this link is for navigating within the app, so I'm going to leave as is. (I poked around the code to verify this and found a `target="blank"` that breaks this logic, so I've pushed a fix for that.)

PR #1175, opened 2019-10-17
Limit up to two total events and tours in explore

### LINKED ISSUE #1171: Show max 2 total events and tours in Home/Explore All tab
We currently show up to two events and up to two tours on the All tab of Home/Explore. Instead we should show up to 2 events and tours total.

Normally this would be 1 event and 1 tour. But in case there are zero events (or tours), we can show up to 2 tours (or events).

flibbertigibbet:
## Overview

Show one event and one tour at top, unless one type is missing, then show up to two of the other.

### Demo

image

## Testing Instructions

 * Reload home page (uses Django template)
 * Should have expected number of events and tours
 * Toggle home page filter to another tab and back to all (uses JS template)
 * Should still have expected number of events and tours

Closes #1171

### DIFF
--- a/python/cac_tripplanner/destinations/views.py
+++ b/python/cac_tripplanner/destinations/views.py
@@ -62,8 +62,13 @@ def home(request):
     article = Article.objects.random()
     # Show all destinations, events, and tours
     destinations = list(Destination.objects.published().order_by('priority'))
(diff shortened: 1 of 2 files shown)


flibbertigibbet:
Derp, fixed.

PR #1176, opened 2019-10-17
Fix explore markers when no origin set

### LINKED ISSUE #1173: In no-origin explore mode, map markers are translucent
On the tours branch, when in explore mode with no origin set, all map markers are translucent. 

Compare to the correct behavior in production, where zero map markers are translucent when in that state.

flibbertigibbet:
## Overview

Do not present markers as translucent (unmatched) in explore mode when no origin is set.

### Demo

Markers are orange when no origin is set:
image

Markers are still translucent when origin set, but no locations matched:
image

## Testing Instructions

 * Go to explore map page with no origin set
 * Map markers should be solid orange

Fixes #1173


flibbertigibbet:
build again

PR #1186, opened 2019-10-22
Feature/kak/undo tour order#1180

### LINKED ISSUE #1180: Add undo/reset button for tour ordering
On tour map page, add and implement the undo button that resets the place order.

image

(Reset button should be hidden for events, unlike in mock above.)

flibbertigibbet:
## Overview

Add and implement undo and remove buttons for user-ordered tours.

If there are at least three tour destinations remaining in a tour, show an 'X' button for the user to remove a place from the tour and requery for the itinerary.

The undo button resets the order to the admin-assigned default order and adds back any removed destinations to the waypoints.

### Demo

image

## Testing Instructions

 * Go to the map page for a tour
 * Reordering the list by drag-and-drop should still function as expected
 * Clicking the 'X' to remove a destination should remove it from itinerary on map and sidebar list
 * 'X' should be hidden for events and tours with less than three destinations
 * Undo button at top of list should reset to initial state

Closes #1180

### DIFF
--- a/src/app/scripts/cac/control/cac-control-directions.js
+++ b/src/app/scripts/cac/control/cac-control-directions.js
@@ -10,7 +10,7 @@ CAC.Control.Directions = (function (_, $, moment, Control, Places, Routing, User
     var DIRECTION_THROTTLE_MILLIS = 750;
 
     // Number of milliseconds to wait on destination list reorder before requerying directions
-    var REORDER_TOUR_THROTTLE_MILLIS = 1000;
+    var REORDER_TOUR_THROTTLE_MILLIS = 500;
 
     var defaults = {
         selectors: {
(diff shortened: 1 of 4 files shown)


flibbertigibbet:
> * Disable location removal after user removes enough locations that only 2 remain. (Less important)

This is the current behavior.

lederer:
>  It was the behavior, before adding undo.

I might not have described it clearly. Here's the current behavior I'm observing:
- Open map of tour with 3 locations
- Remove 1 location, so there are 2 remaining
- The 2 remaining tours each have a remove button. On removing one, the sidebar contents are replaced with a "Could not plan trip" error.
- EXPECTED: Once there are 2 locations remaining, they cannot be removed and their remove buttons are hidden.

flibbertigibbet:
Thanks, that was my expected behavior as well; you found a bug I introduced adding undo functionality.

flibbertigibbet:
Okay, I fixed the undo removal count bug and hid the undo button for clean tours.

PR #1217, opened 2019-11-19
Bugfix/kak/tour reorder highlighting#1212

### LINKED ISSUE #1212: Fix segment highlighting on tour reorder
After reordering or deleting tour destinations, the map route segment highlight on hover interaction no longer matches. The highlight segments should update with the new order.

flibbertigibbet:
## Overview

Fix tour map segment highlighting on hover after removing or reordering destinations.

### Notes

There was two separate issues: an off-by-one error in assigning order after reordering, and not handling changing the assigned order of subsequent places after removing one.

## Testing Instructions

 * Go to a tour map page
 * Remove a place by clicking the 'x' in the top right of its card
 * Hovering over places should still highlight the correct map segment
 * Reorder places by dragging a card
 * Hovering over places should still highlight the correct map segment
 * Set and/or clear the origin
 * Hovering over places should still highlight the correct map segment

Closes #1212

### DIFF
--- a/deployment/ansible/group_vars/all
+++ b/deployment/ansible/group_vars/all
@@ -17,9 +17,6 @@ packer_version: "1.4.4"
 nodejs_version: 12.13.1
 nodejs_npm_version: 6.13.1
 
-otp_version: "1.4.0"
-otp_jar_sha1: "0367b1a15bac5f587807a5b897a9734209f8135c"
-
 otp_router: "default"
 
 # used by nginx and gunicorn to set timeouts; OTP defaults to 30s
--- a/deployment/ansible/roles.yml
+++ b/deployment/ansible/roles.yml
@@ -1,5 +1,5 @@
 - src: azavea.opentripplanner
-  version: 2.0.0
+  version: 2.0.1
 
 - src: azavea.nginx
   version: 0.3.1
--- a/python/cac_tripplanner/destinations/migrations/0057_default_tours.py
+++ b/python/cac_tripplanner/destinations/migrations/0057_default_tours.py
@@ -24,8 +24,10 @@ def add_sample_tours(apps, schema_editor):
 
     tour = Tour(**tour_data)
     tour.save()
+    dest_num = 1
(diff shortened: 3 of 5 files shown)


KlaasH:
Actually it looks like it works correctly after I've reordered the destinations, but not when I first load the directions or after removing one but not reordering.

flibbertigibbet:
The default tour created by fixture doesn't properly set the order; I think this may be the result of that. Going in to the tour in the admin site and saving it should trigger the logic to automatically fix the order.

I'll make an issue to fix the default tour fixture ordering.

flibbertigibbet:
I rebased this branch on `tours` to fix a merge conflict and pushed a commit to add explicit ordering to the default tour migration.

KlaasH:
I resaved the tour and that made the highlighting work properly on initial load.  And it still seems to work properly after reordering.  I'm still seeing a bug after deletion, though.  Some kind of off-by-one error again: when I delete a destination, the highlighting for the one before it gets broken.  So e.g. starting from the default ordering (Waterworks, Palmyra Cove, Schuylkill Center, Seaport Museum, etc), if I delete the Seaport Museum, once the new directions have loaded, hovering over the Schuylkill Center card highlights the Palmyra Cove route and marker. The others still work correctly, only the one before the deleted destination gets messed up.

flibbertigibbet:
I cannot reproduce any issue with highlight ordering on this branch on delete or reorder, with or without an origin set. Did you maybe switch branches without rebuilding the JS?

KlaasH:
I checked the "Disable cache" box in the developer tools network tab and found a modified section of the code in the Sources tab to be sure the front-end was showing the branch.  It seems like it is.  Here's an animation showing the bug, starting from the initial unmodified tour, showing the highlighting working properly before the destination is removed then malfunctioning after.

cec-tour-highlighting-after-deletion2

flibbertigibbet:
Thanks for the recording. I was able to reproduce it and it was indeed an off-by-one error.

Issue #1219, opened 2019-11-19
Tour sidebar place card text cut off


flibbertigibbet:
The bottom of the tour sidebar place card name text gets cut off slightly (see the 'g' below).

![cac_tour_place_card_text_cutoff](https://user-images.githubusercontent.com/960264/69110850-df17f100-0a49-11ea-91b8-96b50f64077d.png)

flibbertigibbet:
Closed in #1221

PR #1221, opened 2019-11-19
Fix compact place card layout

lederer:
## Overview

Adjusts layout of tour place cards so descenders in place name are visible.

### Demo

image

image

Connects #1219


lederer:
> a scrollbar has appeared in the mobile carousel

Thanks for catching. Which browser is that? I was unable to recreate on iOS, Android 9, macos Chrome at mobile width, or macos FF at mobile width.

PR #1227, opened 2019-11-20
Bugfix/kak/fix back nav#1207

### LINKED ISSUE #1207: Fix origin recognition back navigation to event
Steps to reproduce:

 - go to map page for an event with no origin set
 - click 'directions' link for a place in the event
 - set origin
 - go back to event places list
 - click 'directions' link for a different place in the event
 - origin text is still set, but not recognized (highlighted in error and not set in URL or used as origin)

Origin should continue to be recognized through state navigation on the map.

### LINKED ISSUE #1223: Errors on reversing origin and destination with a tour
Steps to reproduce:

 - set a tour as the destination on the map page
 - click to reverse origin/destination
 - click again to reverse origin/destination

There are console errors, and the trip plan fails.

image

Tours cannot be used as an origin, so how to resolve is unclear.

flibbertigibbet:
## Overview

Fixes some state management issues around the new tours feature:

 - Fixes syncing the form fields with the URL on browser back navigation.
 - Fixes tour destination being lost on page reload (not getting synced with URL)
 - Handles user attempting to reverse origin/destination, when new origin is a tour
 - Removes tours and events from origin field autocomplete

### Demo

Error set when user reverses origin and destination, making a tour the origin:
image

## Testing Instructions

 * Tours and events should not show in origin field autocomplete
 * Tours and events should still be in destination field autocomplete
 * Go to map page for a tour
 * Refresh page; destination should still be set to the tour
 * Set the origin
 * Click the form button to reverse origin and destination
 * Form error should be set on tour origin
 * Refreshing page should preserve destination and clear origin
 * Set destination to an event and clear origin
 * Click directions button on an event place card
 * Set the origin
 * Click back
 * Origin should clear
 * Click back again
 * Should return to event page with origin unset
 * Click directions button for another event place card
 * Should go to directions for the place with the origin still unset

Fixes #1207 
Fixes #1223

### DIFF
--- a/src/app/scripts/cac/control/cac-control-directions-form.js
+++ b/src/app/scripts/cac/control/cac-control-directions-form.js
@@ -87,10 +87,14 @@ CAC.Control.DirectionsFormControl = (function ($, Typeahead, Geocoder, UserPrefe
 
         if (origin && origin.location) {
             typeaheads.origin.setValue(UserPreferences.getPreference('originText'));
+        } else {
+            typeaheads.origin.setValue('');
         }
 
         if (destination && destination.location) {
             typeaheads.destination.setValue(UserPreferences.getPreference('destinationText'));
+        } else {
+            typeaheads.destination.setValue('');
         }
     }
 
--- a/src/app/scripts/cac/control/cac-control-directions.js
+++ b/src/app/scripts/cac/control/cac-control-directions.js
@@ -698,6 +698,8 @@ CAC.Control.Directions = (function (_, $, moment, Control, Places, Routing, User
             // results from API search endpoint.
             // If result did not come from Typeahead, the full tour will be
             // loaded in the Places query in setFromUserPreferences.
+            result.address = result.name;
+            directionsFormControl.setStoredLocation('destination', result);
             tour = result;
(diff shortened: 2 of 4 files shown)


KlaasH:
Most of these are working for me.  A couple exceptions:

- Click the form button to reverse origin and destination
- Form error should be set on tour origin

I get an error, but not the intended one:
image

The tour is one I made locally, consisting of the Schuylkill Center and the Independence Seaport Museum.

- Should return to event page with origin unset

This works, but it makes some noise in the console:
image

flibbertigibbet:
I've pushed a fix for the 'could not find trip' error. A page refresh after setting the origin and tour destination was necessary to reproduce.

I haven't seen the markers `getLatLng` console error in revisiting this, but saw it before and resolved it, maybe in another recent PR? I don't remember the cause, but I'm not particularly worried about it now. We can check that it's gone on staging before final deployment.

flibbertigibbet:
I considered trying to clear the error when the user swaps back, but concluded that it would be an edge case not worth the effort, and easy enough for the user to clear by re-selecting the tour destination.

Issue #1201, opened 2019-10-31
Refreshing mobile tour does not initiate carousel


lederer:
On mobile, navigate to a tour map. Refresh browser. Locations are in a normal flow layout rather than arranged in a carousel. The tiny-slider wrapper elements are not present in the DOM.

This affects tours and multilocation events, but does **not** affect direction itineraries.

![image](https://user-images.githubusercontent.com/128699/67978893-7b8c6780-fbf1-11e9-9df5-4d1a9c32d527.png)

flibbertigibbet:
When this happens, there are two `.tour-list` results in the DOM (there should only be one).

flibbertigibbet:
Requires production JS build to reproduce (npm run gulp-production).

flibbertigibbet:
Closed in #1234.

PR #1234, opened 2019-11-27
Workaround for extra tour list

flibbertigibbet:
Work around having two tour lists in the DOM in the production build when the page is loaded directly by removing the first.

Connects #1201.

### DIFF
--- a/python/cac_tripplanner/templates/home.html
+++ b/python/cac_tripplanner/templates/home.html
@@ -217,7 +217,7 @@
     <div class="directions-results">
         {% include "partials/spinner.html" %}
         <div class="directions-list"></div>
-        <div class="tours"></div>
+        <div class="tour-places"></div>
     </div>
 
     <div class="directions-step-by-step"></div>
--- a/src/app/scripts/cac/control/cac-control-tour-list.js
+++ b/src/app/scripts/cac/control/cac-control-tour-list.js
@@ -15,7 +15,7 @@ CAC.Control.TourList = (function (_, $, MapTemplates, Utils) {
         showShareButton: false,
         selectors: {
             alert: '.alert',
-            container: '.tours',
+            container: '.tour-places',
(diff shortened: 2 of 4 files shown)


KlaasH:
I tried to look into why there might be two copies of the thing, and I have a potential lead.  `MapTemplates.tourDestinationList` is getting added to the DOM [a few lines above where you added the workaround](https://github.com/azavea/cac-tripplanner/pull/1234/files#diff-c85918693db112e87a556d707e953a81R89-R90), and when I logged the pieces of that, I noticed the `$container` selector had two members.
The `Tours` tab on the home view uses the class `tours` (added in https://github.com/azavea/cac-tripplanner/commit/90f115cfb2dc26b2e65314ee99e43c72aeaec96f), and so does the `directions-results` panel (see https://github.com/azavea/cac-tripplanner/commit/24ab904cd71268ab0a68c821feda9dca07f83567).  I don't think they're intended to be related, and it looks to me like the tab class isn't actually used for anything.  But the selector for `$container` in `cac-control-tour-list.js` is just `'.tours'`.  So it's matching both.

Here's where my story gets a little more vague, because I didn't actually track it this far, but my guess is that it's inserting the new copy of the list into the tab rather than replacing the one in the `directions-results` panel.  And maybe that the carousel activation code isn't firing because the selector doesn't match?  Or it's attached to the old copy and won't see what's happening in this other part of the DOM?

Anyway, when I change the tab class to `tours-tab`, reloading the mobile tours view doesn't produce the bug.  It does this, which isn't quite right either, but it at least has a carousel...
image

flibbertigibbet:
That was it, thanks :confetti_ball:

I've renamed the `tours` div in the sidebar results rather than the filter tab, as the filter tab classes correspond to the data filters in use.

flibbertigibbet:
I also snuck in a fix for Chrome mobile not always fully loading the map tiles.

Issue #1235, opened 2019-12-03
User edits to tours do not clear on browser back


flibbertigibbet:
Clear any user reordering or removal of tour places when tour returned to via browser back button.

flibbertigibbet:
Closed in #1236.

PR #1242, opened 2019-12-04
Preserve tour edits on returning to tour

### LINKED ISSUE #1241: Preserve user edits to tour on browser back
Preserve changes to tour order and/or removed places in the tour on browser back from place-to-place directions to the tour place list. Changes should appear both on map and in sidebar. All tour changes are lost on page refresh.

flibbertigibbet:
## Overview

On clicking to go back to the tour places list after viewing place-to-place directions,
preserve any user changes to place order or removed places.

### Notes

Supersedes #1235 by fixing edits appearing in the sidebar but not on the map.

## Testing Instructions

 * Go to the map page for a tour
 * Reorder the tour and/or remove one or more places from it
 * Click on a place card 'directions' button
 * Click either browser back or the back button at the head of the directions list
 * Tour edits made above should be preserved on both map and in sidebar
 * Page refresh should load tour with no edits

Closes #1241.

### DIFF
--- a/src/app/scripts/cac/control/cac-control-directions.js
+++ b/src/app/scripts/cac/control/cac-control-directions.js
@@ -377,6 +377,7 @@ CAC.Control.Directions = (function (_, $, moment, Control, Places, Routing, User
             showPlaces(false);
             planTrip();
         } else {
+            clearItineraries();
             showPlaces(true);
             exploreControl.getNearbyPlaces();
         }
@@ -618,6 +619,8 @@ CAC.Control.Directions = (function (_, $, moment, Control, Places, Routing, User
             }
         } else {
             $(options.selectors.reverseButton).show();
+            tour = null;
+            tourListControl.resetTour();
         }
 
         clearItineraries();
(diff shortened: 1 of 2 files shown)


KlaasH:
The actions listed above all work as expected.
While testing this out, though, I ran into two weird behaviors:
- The reset button doesn't fully reset.  On the Delaware Watershed Alliance tour, I removed John Heinz and moved Bartram's Garden up to the 2nd stop.  Clicking the reset button didn't put it back to its original state (as reloading the page does), it left Bartram's Garden in the 2nd position and added John Heinz at the end (in the unmodified tour it's the 3rd stop).
image

- I made a smaller test tour (for variety and faster directions loading) and discovered that when I loaded that tour, modified it, then clicked "Explore" and selected the other tour, it loaded the name of the other tour but kept the stops and directions of the first tour.  It seems like my customizations stuck around and are being applied to the new tour (i.e. replacing all its stops with the ones I had set for the first tour).
image

flibbertigibbet:
Updated. Good catch! It should now also reset the tour changes if the edited tour name is re-selected from the dropdown, and also I fixed an issue with the itinerary on the map not always clearing on switching to the explore tab.

flibbertigibbet:
Thanks!

Deleting the default tour and rerunning [destinations migration 0045](https://github.com/azavea/cac-tripplanner/blob/develop/python/cac_tripplanner/destinations/migrations/0057_default_tours.py) should fix the default tour order.

PR #1248, opened 2020-01-16
Bugfix/kak/fix invalidate size uninitialized map

### LINKED ISSUE #1247: Mobile directions broken when map not yet initialized
Page does not go to directions when it should while in mobile mode with an uninitialized map (as when getting directions from the home page.)

Console error:

image

flibbertigibbet:
## Overview

Fixes issue with mobile site not loading directions if the map had not yet been loaded.

Also includes minor updates to some JS dependencies.

## Testing Instructions

 * (Re-)load mobile site home page
 * Should be able to get directions from home page form/directions button on place cards
 * Map tiles should still load consistently on Chrome

Fixes #1247


flibbertigibbet:
Trivial; will merge on CI build pass.

PR #1254, opened 2020-03-25
Temporary transit warning

flibbertigibbet:
## Overview

Show alert on all transit trips and change default mode to bicycling, as a temporary measure.


flibbertigibbet:
Merging to deploy.

PR #1266, opened 2020-10-13
Upgrade OS

flibbertigibbet:
## Overview

This PR upgrades the Ubuntu version used in the VMs and deployment AMIs from `bionic` to `focal`. Both are long-term support releases. This PR also installs `osmctools` in the app VM, as a convenience for managing OSM data downloads.

This is not a necessary upgrade to make right now, but may be helpful to do because it enables installing a version of `osmconvert` (in `osmctools`) that can support the `complete-boundaries` parameter for merging and clipping OSM extracts; the version of `osmconvert` in `bionic` is not recent enough for that (> 0.7 needed).

### Demo

In the app VM:
image

### Notes

Replacing the `nginx` role was necessary, because the unofficial Ubuntu `ppa` has not been updated to support `focal`. This installs the [official repository](http://nginx.org/en/linux_packages.html#Ubuntu), and copies the template files and configuration tasks from the `nginx` role.

## Testing Instructions

If this might be a useful upgrade to make at this time, it can be tested during VM and AMI building as part of paring to redeploy.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -71,7 +71,7 @@ VAGRANT_PROXYCONF_ENDPOINT = ENV["VAGRANT_PROXYCONF_ENDPOINT"]
 VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
-  config.vm.box = "bento/ubuntu-18.04"
+  config.vm.box = "bento/ubuntu-20.04"
 
   # Wire up the proxy if:
   #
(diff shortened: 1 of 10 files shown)


flibbertigibbet:
rebuild now

flibbertigibbet:
build again

flibbertigibbet:
build again

flibbertigibbet:
It looks like the log file location for OTP would need to be updated to move it out of `/var/log/upstart`.

Issue #1316, opened 2022-03-15
Automate process for checking otp_url before deploy


rachelekm:
Our current deployment process involves manually following a complicated checklist of tasks that leaves us prone to human error each month. While we don't have the bandwidth to automate this process right now, we can create a script to doublecheck our otp url and confirm it is the expected stack color before deleting the previous stacks and causing a break in routing functionality in production.

We will need to update the deployment README to include this script as well and any preliminary tasks for it to function correctly.

PR #1317, opened 2022-03-15
Create otp_url check for deployment

rachelekm:
## Overview

This script prints the `otp_url` used by the latest AWS `WebServerStack` so that a user can manually confirm it matches the color of the `OtpServerStack` before proceeding with deployment. This is intended as a stopgap until we can better automate the deployment process.

### Notes

- This script will connect to AWS EC2 instances through `ssh` and will need the appropriate identity file. Ensure file is within local `~/.ssh` folder and given 600 permissions.
- We will need to include an instruction to run this script in the deployment `README.md` after new stacks are up and before previous ones are deleted.
- I wasn't sure how far to take this script, but AWS cli _does_ have an `aws cloudformation describe-stacks` command that I can filter, grab a color tag, and potentially compare to the color contained in`otp_url`. I think it would be a reasonable enough addition and would allow us to print a more clear cut `TEST PASS` vs `FAIL`. However, in previous "hybrid" stacks (ie a color stack mismatch), the Cloudfront cache was used instead of the `otp_host` defined in the `group_vars/production` file. It's unclear whether this will be the case in future color stack mismatches, but I figured it might be worth having a user just manually confirm the url until we sort it out.

## Testing Instructions
 * Set up identity file
 * Run `./scripts/check-deploy-otp-url.sh`
 * Confirm `otp_url` prints and matches color expected for latest stack (at time of writing this PR it should be `blue`)

Connects #1316

### DIFF
--- a/scripts/check-deploy-otp-url.sh
+++ b/scripts/check-deploy-otp-url.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+set -e
+
+echo "Fetching otp url used by latest WebServer instance..."
+
+WEBSERVER=$(AWS_PROFILE=gophillygo aws ec2 describe-instances \
+    --filters 'Name=tag:Name,Values=WebServer' \
+    --query 'Reservations[].Instances[].{ip:PrivateIpAddress,time:LaunchTime}' | \
+    jq -r 'sort_by(.time) | reverse | .[0].ip')
+
+#There is only one, but sort by latest in event of multiple instances
+BASTION=$(AWS_PROFILE=gophillygo aws ec2 describe-instances \
+    --filters 'Name=tag:Name,Values=BastionHost' \
+    --query 'Reservations[].Instances[].{dns:PublicDnsName,time:LaunchTime}' | \
+    jq -r 'sort_by(.time) | reverse | .[0].dns')
+
+ssh -o ProxyCommand="ssh -i ~/.ssh/cac.pem -W %h:%p ubuntu@$BASTION" -i ~/.ssh/cac.pem ubuntu@$WEBSERVER /bin/bash << PRINTURL
+    sed -n 14p /etc/cac_secrets
+    exit
+PRINTURL


ddohler (commented):
Thanks for putting this together--I learned some fun new SSH tricks from this!

I made some suggestions for things to change for what's here, but I also think that it would be worth rethinking the overall form of this solution. This script has enough dependencies (specifically the AWS CLI and `jq`) that I think adding some form of encapsulation (e.g. Docker) would be important to ensure that the correct dependencies (and versions of those dependencies) are available to run the script successfully. However, adding that encapsulation is going to add further complexity to what is already a complex set of deployment tools.

I was thinking that perhaps if we structured the SSH command as a part of the deployment instructions that we expect people to copy-paste and then modify, it might provide us with most of the user-friendliness of the script approach, without the additional engineering effort required to maintain a shell script and its dependencies. We know that anyone who has reached this point in the instructions will have the AWS web console open anyway, so it shouldn't be too much of a lift for them to pull out the necessary IP addresses and insert them into the command. I was able to get this to work as a single copy/pasted command from my development machine:

`ssh -o ProxyCommand="ssh -i ~/.ssh/cac.pem -W %h:%p ubuntu@[bastion ip]" -i ~/.ssh/cac.pem ubuntu@[app server internal ip] 'cat /etc/cac_secrets | grep otp_url'`

What do you think?

rachelekm:
Thanks for the review and edits @ddohler! I think pulling out the ssh command and adding it to the deployment instructions is a good idea. In an earlier conversation you had also mentioned how querying the needed stack ip addresses by date could potentially lead to error if a desired instance was unknowingly not the most recent, so having a user just grab the addresses while they have the console open eliminates that risk as well without the additional extra engineering. That command worked well for me, and I've created an issue to add it to the deployment instructions, [here](https://github.com/azavea/geospatial-apps/issues/259).

PR #1320, opened 2022-04-06
Upgrade to Django 3.2

### LINKED ISSUE #1265: Upgrade to Django 3.x
Upgrade Django from 2.x to 3.x. The project has already been upgraded to Python 3.

ddohler:
## Overview

Gets us onto the most recent long-term support release.

### Demo

Not much to demo -- everything should remain the same.

### Notes

The app actually worked mostly fine on 3.2 without additional work, probably because anything deprecated in 3.2 had already been addressed by earlier rounds of upgrade work. I tried to do the same thing -- I addressed as many deprecation warnings as I could, even though they were mostly for Django 4+.

Many of the issues I encountered stemmed from `django-wpadmin`, which is a Django admin theme that attempts to replicate the look and feel of the WordPress admin. The upstream package is no longer maintained, so we've long since forked it, and I continued updating our fork.

## Testing Instructions

 * I assume it's because we're installing `django-wpadmin` from a GitHub repo rather than PyPi, but reprovisioning didn't pick up the changes to `requirements.txt`. Instead, I had to do the following to get it to install:
   * `vagrant up && vagrant ssh app`
   * `sudo pip3 uninstall django-wpadmin` (and then logout)
   * `vagrant provision app`
 * Next, run tests and the dev server to make sure that you don't see any deprecation warnings from Django (there will be a couple from boto):
   * `vagrant ssh app`
   * `cd /opt/app/python/cac_tripplanner`
   * `python3 -Wall manage.py test`
   * `sudo systemctl stop cac-tripplanner-app` (This stops the server process that comes up by default when you start the VM; you need to do this before starting up the Django dev server manually because you'll get a port conflict if you don't)
   * `python3 -Wall manage.py runserver`
 * Last, click around the app and make sure everything works as it used to.
 * Pay special attention to the admin site -- I encountered quite a few CSS issues that caused the layout to get messed up, so it's very possible that there are still some more lingering that I've missed.

## Checklist
- [x] No gulp lint warnings
- [x] No python lint warnings
- [x] Python tests pass

Closes #1265

### DIFF
--- a/python/cac_tripplanner/cac_tripplanner/publish_utils.py
+++ b/python/cac_tripplanner/cac_tripplanner/publish_utils.py
@@ -1,14 +1,17 @@
-from django.utils.translation import ugettext_lazy
+from django.utils.translation import gettext_lazy
 
 
 class PublishableMixin:
     """Helper to add Django admin actions to publish and unpublish model objects."""
-    actions = ('make_published', 'make_unpublished')
+
+    actions = ("make_published", "make_unpublished")
 
     def make_published(self, request, queryset):
         queryset.update(published=True)
-    make_published.short_description = ugettext_lazy("Publish selected %(verbose_name_plural)s")
+
+    make_published.short_description = gettext_lazy("Publish selected %(verbose_name_plural)s")
 
     def make_unpublished(self, request, queryset):
         queryset.update(published=False)
-    make_unpublished.short_description = ugettext_lazy("Unpublish selected %(verbose_name_plural)s")
+
+    make_unpublished.short_description = gettext_lazy("Unpublish selected %(verbose_name_plural)s")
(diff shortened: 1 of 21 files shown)


rachelekm (approved):
This worked well for me! The site functions as expected after updates and I did a more close comparison of admin but couldn't find anything css-wise that stuck out. One note for future testers, since I briefly got turned around in the testing instructions: it's `sudo pip3 uninstall django-wpadmin` instead of `django-admin`

ktohalloran (changes requested):
It looks great! There was a little bit of CSS weirdness in my admin, however, on the 'Articles' and 'About faqs' table views. The list of pages were really far over to the right on my screen and didn't quite look like the rest of the admin table views. They were also accompanied in my terminal with a couple of deprecation in Django 4.0 warnings: `The {% ifequal %} template tag is deprecated in favor of {% if %}.`

A few other things I noticed but that are more notes than edits: on opening individual page views, like "Crazy Philadelphia Eddie", I saw unclosed file errors in my terminal. These seem to be something unaddressed within Django's own source code, however, as it pointed to `site-packages/django/forms/fields.py` where images are being opened using `Image.open()` rather than the `with(open(... as Image))` like you did. I also ran into similar issues with provisioning my VM as Rachele; I fixed it by running `pip3 install -r requirements.txt` directly inside the VM, which was something I also had to do while updating Django for Cicero.

ktohalloran (approved):
Ok, my VM got into a bit of a tizzy and needed to be rebuilt but you were right, uninstalling `django-wpadmin` and then re-provisioning worked so it looks ok now. So looks good! 👍

PR #1354, opened 2023-06-14
Upgrade Troposphere to version 3

### LINKED ISSUE #1347: Upgrade troposphere
During deployment, attempting to run `pip install -r python/cac_tripplanner/deployment_requirements.txt` with a freshly-created virtual environment results in 
```
Collecting troposphere==2.6.2                                                                              
  Using cached troposphere-2.6.2.tar.gz (179 kB)
  Preparing metadata (setup.py) ... error         
  error: subprocess-exited-with-error                                                                      
              
  × python setup.py egg_info did not run successfully.
  │ exit code: 1                                                                                           
  ╰─> [1 lines of output]                                                                                  
      error in troposphere setup command: use_2to3 is invalid.                                                                                                                                                        
      [end of output]                         
```

Which is due to `setuptools` [removing support for `use_2to3` in version 58](https://setuptools.pypa.io/en/latest/history.html#v58-0-0). Manually downgrading to `setuptools==57.5.0` fixes the problem and allows deployment to proceed, but ideally, we should upgrade to a new version of troposphere (we are on 2.6.2 and the latest is 4.2.0) that will hopefully fix the issue.

ddohler:
## Overview

Upgrading Troposphere broke majorkirby because it uses `add_version` in its `set_up_stack()` method, which was changed to `set_version()` in Troposphere v3. To work around the issue without pushing a new release to majorkirby, this removes calls to `super()` so that the code using `add_version()` doesn't get called, and then calls `set_version()` in local project code so that functionality is preserved. If it turns out we have more majorkirby projects hidden in a closet somewhere then we should probably push this fix up to majorkirby, but for the time being this is the most time-efficient thing to do. Also CC @rajadain in case this pops up elsewhere. 

### Notes

- I had apparently encountered this once before and forgotten about it because there was already an issue open. Oops.
- This doesn't upgrade troposphere all the way because there are more breaking changes in v4 that affect us and they look substantial enough that I think they're better left alone for now.

## Testing Instructions

 * Follow the deployment instructions to launch a new stack. As long as you've done it at least once before, you shouldn't need to build AMIs because we're not actually going to do anything with the stack we launch. Double-check the [existing stack color in the Cloudformation console](https://us-east-1.console.aws.amazon.com/cloudformation/home?region=us-east-1#/stacks?filteringText=&filteringStatus=active&viewNested=true) and launch the other color (it doesn't matter whether it matches the color built into the AMI, but we don't want to clobber anything in the existing stack). If you haven't launched the app before you can [grab some AMI ids from the console](https://us-east-1.console.aws.amazon.com/ec2/home?region=us-east-1#Images:visibility=owned-by-me;sort=desc:creationDate) to use.
 * Confirm that the stack launches successfully. You don't need to check that the stack is actually functional, although you can if you want.
 * Return to the [CloudFormation console](https://us-east-1.console.aws.amazon.com/cloudformation/home?region=us-east-1#/stacks?filteringText=&filteringStatus=active&viewNested=true) and (very carefully) delete the stacks you just launched (and not the existing production stacks).

## Checklist
- [x] ~No gulp lint warnings~
- [x] No python lint warnings
  - I actually did run this and there were some warnings, but not in any of the code that I touched, so I'm going to ignore them unless / until we actually need to touch that code again.
- [ ] ~Python tests pass~

Closes #1347

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,16 +29,18 @@ Note that if there is an existing build Graph.obj in `otp_data`, vagrant provisi
 
 Building AMIs
 ------------------------
+1. Configure an AWS profile with `aws configure --profile gophillygo` if you haven't already
 1. Make a production group_vars file (similarly to how is described above with development). Make sure production is set to true, and also specify an app_username, which should be set to: ubuntu
 2. If building the `otp` machine, make sure the latest GTFS are in `otp_data`, then build a graph when them in the development environment provisioning.  This will result in a new `Graph.obj` file being written to `otp_data`.
-3. In the project directory within the app VM, run: `deployment/cac-stack.py create-ami --aws-access-key-id YOUR_ACCESS_KEY --aws-secret-access-key YOUR_SECRET_KEY --aws-role-arn YOUR_ASSUMED_ROLE_ARN`
+3. In the project directory within the app VM, run: `AWS_PROFILE=gophillygo deployment/cac-stack.py create-ami`
 4. The previous command builds all AMIs. To only build a single AMI, run the same command, but also specify the `--machine-type` parameter, which may be set to one of: `bastion`, `otp`, or `app`.
 
 
 Launching AWS Stacks
 ------------------------
 1. Copy `deployment/default_template.yaml` to `deployment/default.yaml` and edit variables
-2. In the project directory, for a set of `Blue` stacks in the `Production` environment, run: `deployment/cac-stack.py launch-stacks --stack-type prod --stack-color blue --aws-access-key-id YOUR_ACCESS_KEY --aws-secret-access-key YOUR_SECRET_KEY --aws-role-arn YOUR_ASSUMED_ROLE_ARN`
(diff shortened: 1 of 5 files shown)


KlaasH (approved):
:+1: Looks good!  I as able to successfully create the stacks.

The one thing I noticed was that the instructions call for passing the AWS key as options to `cac-stack.py`, i.e. `--aws-access-key-id YOUR_ACCESS_KEY --aws-secret-access-key YOUR_SECRET_KEY --aws-role-arn YOUR_ASSUMED_ROLE_ARN`, but when I tried that it said "error: unrecognized arguments".  But configuring a profile and setting it in the environment worked fine.  I could open an issue for that, but I'm not sure if there's an existing issue that might cover it or any broader context around the deployment documentation.

ddohler:
Good catch! I don't normally refer to the deployment instructions in this repo because we have the detailed instructions for the production instance in a private location, but the ones that are here shouldn't be flagrantly wrong like this, so I've updated them to match the instructions in the document we use for ongoing deployments.

Issue #1356, opened 2023-07-10
Upgrade Python in the app VM


ddohler:
We're on Python 3.6, which is [now considered end-of-life](https://devguide.python.org/versions/), and we will not be able to upgrade to Django 4.2 without upgrading Python.

rachelekm:
To be completed as part of #1372 task list. As part of this task we also need to upgrade `psycopg2` from 2.8.6 to 2.9.0.

Issue #1302, opened 2021-12-02
Update Vagrant internal IPs to use 192.168.56.0/21 range


ktohalloran:
### Overview
Virtualbox 6+ has added a restriction on IPs used for host-networking, and no longer allows the IPs we are using (192.168.8.XXX) by default.

### Motivation
By updating our internal IPs to match the default ranges, it will be easier to spin up a new development instance of GoPhillyGo.

If we don't do this, our alternative would be to update our README instructions to include how to configure the allowed IP ranges.

### Notes
A similar issue can be found in Cicero [issue 2370](https://github.com/azavea/cicero/issues/2370).

PR #1357, opened 2023-07-10
Don't install deployment dependencies in app VM

ddohler:
## Overview

There was some fallout from [upgrading to Troposphere 3](https://github.com/azavea/cac-tripplanner/pull/1354), namely that the deployment dependencies were getting installed in the app VM, and the app VM is running Python 3.6, which isn't compatible with the version of Troposphere that we upgraded to, which caused app provisioning (and hence AMI building) to fail.

We don't need the deployment dependencies to be in the app VM because we use a Python virtualenv to encapsulate those dependencies separately, so this removes them from the app VM provisioning completely and updates the README instructions. The other option would have been upgrading Python, but that would have been a bigger lift. We've moved toward managing deployment dependencies separately from application dependencies on this and other projects anyway, so this seemed like a good opportunity to cement that change here.

I did open #1356 to cover upgrading Python, and we'll need to do that whenever we upgrade Django, because Django 4.2 no longer supports Python 3.6.

## Notes

- This PR is from the release branch to `master`, so it includes the troposphere upgrade changes, which have been reviewed but hadn't been deployed yet. Only commit 14be417b7964bbfbf0bd6a6f90540daaa5f6fc09 needs to be reviewed.

## Testing Instructions

 * Running `vagrant up app --provision` should not fail (but it will fail on `develop` if you want to check the fix).

## Checklist
- [ ] ~No gulp lint warnings~
- [ ] ~No python lint warnings~
- [ ] ~Python tests pass~

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,14 +32,16 @@ Building AMIs
 1. Configure an AWS profile with `aws configure --profile gophillygo` if you haven't already
 1. Make a production group_vars file (similarly to how is described above with development). Make sure production is set to true, and also specify an app_username, which should be set to: ubuntu
 2. If building the `otp` machine, make sure the latest GTFS are in `otp_data`, then build a graph when them in the development environment provisioning.  This will result in a new `Graph.obj` file being written to `otp_data`.
-3. In the project directory within the app VM, run: `AWS_PROFILE=gophillygo deployment/cac-stack.py create-ami`
-4. The previous command builds all AMIs. To only build a single AMI, run the same command, but also specify the `--machine-type` parameter, which may be set to one of: `bastion`, `otp`, or `app`.
+3. Install the deployment dependencies, ideally in a virtualenv: `python3 -m venv .venv && source .venv/bin/activate && pip install -r python/cac_tripplanner/deployment_requirements.txt`
+4. Build AMIs by running (within the virtualenv): `AWS_PROFILE=gophillygo deployment/cac-stack.py create-ami`
+5. The previous command builds all AMIs. To only build a single AMI, run the same command, but also specify the `--machine-type` parameter, which may be set to one of: `bastion`, `otp`, or `app`.
 
 
 Launching AWS Stacks
 ------------------------
 1. Copy `deployment/default_template.yaml` to `deployment/default.yaml` and edit variables
 1. Configure an AWS profile with `aws configure --profile gophillygo` if you haven't already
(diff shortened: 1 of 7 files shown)


KlaasH (approved):
Looks good :+1: 

I got as far as confirming that provisioning the `app` box works on this branch and crashes when trying to install requirements on `develop`.  I didn't get the whole app up, or even the whole `app` machine provisioned, because I didn't want to try messing with the database VM and without that, the `app` provisioning fails when it tries to run migrations.  But yeah, I got far enough to see it pass vs. crash on the pip installation step.

Other road blocks, and how I got around them, just for the record:
- Problem: `npm install` crashed
  Fix/workaround: I was using node 18. Tried again with node 12 and it worked. Or possibly I was in the wrong Python venv when I ran it? But that doesn't seem like it would matter…
- Problem: Host-only network error. First "address is not within the allowed ranges" then an E_ACCESSDENIED error.
  Fix/workaround: added a `/etc/vbox/networks.conf` file containing:
  ```
  * 192.168.56.0/21
  * 192.168.8.0/24
  ```
  I think the second error was because I had a typo in the file the first time.
- Problem: [This error](https://github.com/mhallin/vagrant-notify-forwarder/issues/8)
  Fix/workaround: I was already on the latest version of the plugin, so upgrading it didn't help. But I just uninstalled it, and that did.

ddohler:
Thanks for testing! Here are my thoughts on the errors you saw:

> Problem: npm install crashed

Installing on-host is only necessary for doing frontend development, which we haven't done in years, so this shouldn't affect our monthly deployments; I'll open an issue to set up NVM to make this easier.

> Problem: Host-only network error. First "address is not within the allowed ranges"

Oops, sorry, this is a [known issue](https://github.com/azavea/cac-tripplanner/issues/1302). I'll prioritize this next sprint since it should be very quick and we have some time.

> Problem: https://github.com/mhallin/vagrant-notify-forwarder/issues/8

I have no idea what's up with this or why we're using this plugin. Based on what it does I would have expected to see it being used for hot-reloading the frontend but this project uses on-host development for that. Perhaps it's for enabling hot-reloading in Gunicorn, but I don't know for sure. I'll open an issue for this too.

ddohler:
Sorry, this has been merged, I used the built-in `git flow release finish` functionality and it deleted this branch remotely before pushing the merged branches, so GitHub sees this as a "Close" rather than "Merge" operation.

Issue #1368, opened 2023-12-18
Update web app to communicate Android sunsetting


rachelekm:
The Android app will be sunsetting before the end of the year (ideally the week of Dec 18th 2023 so users can be notified as part of the week's newsletter) and we should update the web application accordingly with the following UI changes:
- Take down the button promoting the Android app in the top right corner of the application
- Add a banner communicating that the Android app is sunsetting and the site will still be active, copy TBD from Nick

Ensure these updates are part of the upcoming Dec deploy

PR #1369, opened 2023-12-20
Update site UI for android sunsetting

rachelekm:
## Overview

Update the UI with the two most immediate tasks to provide clarity on the Android app sunsetting:

- [x] Remove the Android link from the header
- [x] Add a top banner across pages communicating the app sunset and promoting site use instead 

~~TODO BEFORE MERGE: update banner with approved text from Nick~~

### Demo

https://github.com/azavea/cac-tripplanner/assets/36374797/30534f4d-e441-4b11-9d7d-9bff857659de

### Notes

- Having a banner display and its show/hide state persist across each template page was trickier than I anticipated. The "Home" and "Explore" pages share a template page and get loaded and initialized with the `Home` JS module, but including click events here to show/hide the banner aren't set up to listen on other pages like "Learn" and "About". I tried creating a separate JS module that could load jQuery to a template, listen for a banner click event and add a "hidden" class once clicked, but still ran into issues with class modifications resetting once routing between `Home` or `Explore` and `Learn`. I settled on adding a jQuery block directly in `base.html`, since all pages extend this one, and then adding a boolean to localStorage which could be referenced to conditionally show or hide the banner on template render. However, this is unideal because there's a half second for the jQuery to evaluate and the class styling to update, so the banner doesn't show up until after the page is already displayed. An alternative would be to send the localStorage boolean as part of the template page request, use it to set a `hideBanner` variable in the view that could be loaded in the template context and then use context inline to conditionally add the `hidden` class, but given the tight turnaround and that the banner fits the immediate needs I wasn't sure if this would be worth it. Maybe there's another quick alternative that makes sense here and falls within the scope?

## Testing Instructions

 * If can get spun up, run `vagrant provision app`
 * Confirm banner displays on `http://localhost:8024/` and Android image/link is not longer in the header
 * Click around different pages and confirm banner persists and looks ok, no styling regressions in below content
 * Close top banner, confirm below content takes up entire page like before, and click around other pages and confirm banner no longer displays

## Checklist
- [x] No gulp lint warnings
- [x] No python lint warnings
- [x] Python tests pass
- [x] Update copy with correct message

Connects #1368


rajadain (approved):
+1 looks good, tested with jumping around a few pages. Good to merge once the final text is in.

rajadain:
Nice work and well segmented commits!

rachelekm:
Thanks for the review, @rajadain! Final message and some small accompanying styling to better fit more text added in a5fcddbd3b6c04171b2c757a73e636da78d3a557. Screenshots of final banner:
Screenshot 2023-12-20 at 3 22 14 PM
Screenshot 2023-12-20 at 3 22 01 PM

rajadain:
Thanks for the screenshots. Everything looks good!

Issue #1116, opened 2019-07-12
Switch from boto from boto3


flibbertigibbet:
Use `boto3` instead of the deprecated `boto` library.

PR #1377, opened 2024-02-23
Upgrade boto to boto3 in deployment scripts

rachelekm:
## Overview

Upgrades `boto`, used for our deployment scripts, with `boto3` version `1.15.9`, now at parity with `boto3` used elsewhere in app. 

### Notes
- There's one instance where we use `boto` outside of deployment scripts for the `app` vm, [here](https://github.com/azavea/cac-tripplanner/blob/develop/python/cac_tripplanner/cac_tripplanner/settings.py#L15), used to add the instance IP to `ALLOWED_HOSTS` for the ELBs. There's currently an [open issue](https://github.com/boto/boto3/issues/313) to add a comparable method to `boto3` and until then the preferred solutions seem to continue to use `boto.utils` or a 3rd party package, [ec2-metadata](https://github.com/adamchainz/ec2-metadata#ec2metadatasessionnone). There *is* a [describe_instances](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/describe_instances.html#) method in `boto3` that looks like we can get the public IP address from, but I'm not sure which instance we should be filtering for here that would be comparable to `get_instance_metadata` in production. Ultimately this is a pretty narrow use case for `boto` and it felt worth keeping this one utility in since it's not part of our deployment.

## Testing Instructions

 * We're not able to test this locally since the actions affected by these updates don't actually run until launching a stack. I tried simulating launching a stack locally by manually running the action methods in a python script, but it raises a `majorkirby.MKUnresolvableInputError` when running the commands outside the context of a `GlobalConfigNode`. To test we will need to temporarily launch a web stack to confirm no issues and then delete once successful.

## Checklist
- [ ] No gulp lint warnings
- [ ] No python lint warnings
- [ ] Python tests pass

Connects #1116

### DIFF
--- a/deployment/cloudformation/privatehostedzone.py
+++ b/deployment/cloudformation/privatehostedzone.py
@@ -1,9 +1,8 @@
-from boto import route53
-
 from majorkirby import CustomActionNode
 
 import json
-
+import boto3
+import uuid
 
 class R53PrivateHostedZone(CustomActionNode):
     """Represents a Route53 private hosted zone"""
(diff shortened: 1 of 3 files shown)


deployment/cloudformation/privatehostedzone.py
```diff
@@ -19,21 +18,32 @@ class R53PrivateHostedZone(CustomActionNode):
...
+            'VPCId': self.get_input("VpcId")
+        }
+        hosted_zone_config = {
+            'Comment': comment,
+            'PrivateZone': True
+        }
+
+        hosted_zone = client.create_hosted_zone(
+            Name=self.get_input("PrivateHostedZoneName"),
+            CallerReference=str(uuid.uuid4()),
```
rachelekm, line 43: This method in `boto` had `CallerReference` (`caller_ref`) as an optional parameter and it would generate a Type 4 UUID to use by default. It is now required in `boto3` and can be any unique string, so I'm continuing to use uuid by default here.

rajadain (approved):
+1 tested in pairing with @rachelekm. Spun up a new Green stack, confirmed that it worked, and tore it down. Great job!

Re: remaining boto issue, as long as the upcoming Python upgrades #1356 don't force an upgrade, I think it's fine to keep it.

rachelekm:
Thanks the review and your help pairing!

PR #1380, opened 2024-03-28
Upgrade to Django 4.2

rachelekm:
## Overview

Upgrades to latests Django 4.2 version, which maintains security support April 2026

### Notes
**Remaining deprecation warnings for Django 5+**
Derek largely tackled deprecation warning for Django 4 as part of the work in #1320, so there were minimal changes. Historically we've addressed the next versions's warning as part of current upgrade work, and there are deprecation warning for Django 5, but we're at the end of our contract with minimal budget so I left those warning be for now.

**Django WP Admin**
This is an upstream package that's no longer maintained that we've long since forked and continued updating, used to replicate the look and feel of the WordPress admin in Django. Incompatibilities with our current version of `django-wpadmin` and Django 4+ were largely addressed in #1320 so we didn't have to fix, make a new release, and then update the release tag in our requirements.txt file. One side effect of this however is the original package expects its versioning to maintain parity with Django so we'll continue to see this warning:
```bash
/usr/local/lib/python3.10/dist-packages/wpadmin/__init__.py:13: UserWarning:
Django WP Admin 3.2.14 is not compatible with Django 4.0.1 and may not work properly.
You should install Django WP Admin from 4.0.x branch.
  warnings.warn(
```
Since theres no updates to include in a new `django-wpadmin` release (and given the end of our budget) this felt fine to ignore for now.

**Remaining Boto warnings**
We still use `boto` for one specific method that is now unsupported in `boto3` (and there are no intentions of future support) and has minimal use within our app, see notes of #1377 on the decision to leave `boto` in here. As a result we'll see the following boto deprecation warnings:
```bash
/usr/local/lib/python3.10/dist-packages/boto/plugin.py:40:DeprecationWarning:
the imp module is deprecated in favour of importlib and slated for removal in Python 3.12;
see the module's documentation for alternative uses
  import imp
```
```bash
<frozen importlib._bootstrap>:914: ImportWarning: 
_SixMetaPathImporter.find_spec() not found; falling back to find_module()
```
```bash
<frozen importlib._bootstrap>:671: ImportWarning:
_SixMetaPathImporter.exec_module() not found; falling back to load_module()
  import imp
```
[...]

PR #1378, opened 2024-03-01
Upgrade Python to 3.10 in app VM and psycopg2 to 2.9

rachelekm:
## Overview

Upgrades the app VM from Python 3.6.9 to 3.10.12 as well as upgrades `psycopg2` from 2.8.6 to 2.9 in both app and database VM. Both of these upgrades support upcoming Django and PostgreSQL upgrades in the #1372 task list.

Ideally we would get to Python version 3.12 since that is currently in `bugfix` maintenance status into 2025, but I ran into a variety of issues trying to upgrade to this version so am leaving it at 3.10.12. This version is still receiving security support through October 2026. (see notes for more detail)

### Demo
```
RUNNING HANDLER [cac-tripplanner.app : Restart nginx] **************************
changed: [app]

PLAY RECAP *********************************************************************
app                        : ok=37   changed=32   unreachable=0    failed=0    skipped=17   rescued=0    ignored=0   

rachelemorino@Racheles-MacBook-Pro cac-tripplanner % vagrant ssh app
Welcome to Ubuntu 22.04.3 LTS (GNU/Linux 5.15.0-92-generic x86_64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/pro

  System information as of Fri Mar  1 07:04:49 PM UTC 2024

  System load:  0.39599609375      Processes:             159
  Usage of /:   19.1% of 30.34GB   Users logged in:       0
  Memory usage: 54%                IPv4 address for eth0: 10.0.2.15
  Swap usage:   1%                 IPv4 address for eth1: 192.168.56.24


This system is built by the Bento project by Chef Software
More information can be found at https://github.com/chef/bento
Last login: Fri Mar  1 19:03:31 2024 from 192.168.56.1
vagrant@app:~$ python3 --version
Python 3.10.12
```

```
vagrant@app:~$ pip show psycopg2-binary
Name: psycopg2-binary
Version: 2.9
Summary: psycopg2 - Python-PostgreSQL Database Adapter
Home-page: https://psycopg.org/
Author: Federico Di Gregorio
Author-email: fog@initd.org
License: LGPL with exceptions
Location: /usr/local/lib/python3.8/dist-packages
Requires: 
```

```
vagrant@database:~$ pip show psycopg2
Name: psycopg2
Version: 2.9.9
Summary: psycopg2 - Python-PostgreSQL Database Adapter
Home-page: https://psycopg.org/
Author: Federico Di Gregorio
[...]

PR #1384, opened 2024-05-06
Update AWS Ubuntu version in prod

rachelekm:
## Overview

As part of building the AMI's for the `3.1.46` release, which included Django upgrade work #1380, the build for the `app` AMI errored with `No matching distribution found for Django==4.2.11` on the `Install pip packages for app` task. The AMI was building using `Ubuntu Server 18.04 LTS` which would only have `Python 3.6.8`. This work should have been covered as part of the Python upgrade work in #1378, and was for the local dev environment in https://github.com/azavea/cac-tripplanner/pull/1378/commits/d909581c30650c3eb50a042b632f664543c5d3af, but not the deployment setup which uses an AWS Ubuntu AMI, independently set by querying for an AMI ID from the current release of a given image name. In order to bring production in line with our dev Ubuntu environment and unblock Django upgrades in production, we need to upgrade the server image name in the path to query as part of our packer deployment script.

## Testing Instructions

 * Test AMI's were built using this updated path and used to launch test stack as part of release 3.1.46. Confirm test stack url has no regressions and maintains current functionality: green.production.web.gophillygo.org

## Checklist
- [ ] No gulp lint warnings
- [ ] No python lint warnings
- [ ] Python tests pass

### DIFF
--- a/deployment/packer/cac_packer.py
+++ b/deployment/packer/cac_packer.py
@@ -21,7 +21,7 @@ def get_ubuntu_ami(region):
       creds (Dict): Dictionary containing AWS credentials
     """
 
-    response = urllib.urlopen('http://cloud-images.ubuntu.com/query/bionic/'
+    response = urllib.urlopen('http://cloud-images.ubuntu.com/query/jammy/'
                               'server/released.current.txt').readlines()
     response = [x.decode('utf-8') for x in response]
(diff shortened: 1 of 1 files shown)


rajadain (approved):
+1 tested. Clicked around https://green.production.web.gophillygo.org/ and found it to work correctly. AMI builds also passed. Nice work.

PR #1387, opened 2024-10-18
Provisioning Fixes

rajadain:
## Overview

Updates to provisioning to get this project to setup. With `virtualbox` now being the default for this app, we would no longer need to specify `CAC_APP_SHARED_FOLDER_TYPE=virtualbox` before running `vagrant` commands on Windows and Linux.

### Notes

By tying our NPM version to Ubuntu we make it less explicit, and potentially risk having it change when we do an Ubuntu upgrade. However, it was the easiest way to get a reliably working environment, and since the associated Ansible Role is old and unmaintained, I think it's alright.

## Testing Instructions

- Check out this branch
- Remove your local `node_modules` folder
    ```bash
    rm -rf src/node_modules
    ```
- Delete and reprovision the `app` VM
    ```bash
    vagrant destroy -f app && vagrant up app
    ```
  - [x] Ensure it provisions correctly
- Build the front-end
    ```bash
    ./scripts/serve-js-dev.sh
    ```
- Go to http://localhost:8024/
  - [x] Ensure you see the running app

## Checklist
- [x] No gulp lint warnings
- [x] No python lint warnings
- [x] Python tests pass

Connects #XXX

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,9 +17,9 @@ Development Installation
 3. (Optional) Generate a graph file with (takes approx 3 hours) `docker-compose run --rm otp otp --build /var/otp` in the deployment/graph directory.
 4. Copy `deployment/ansible/group_vars/development_template` to `deployment/ansible/group_vars/development`
 5. Change into the `src/` folder and run `npm install` to install the node modules on the host machine
-6. Run `vagrant up`. You can choose to change the Virtualbox shared folder type for the `app` VM from its default NFS by:
+6. Run `vagrant up`. You can choose to change the Virtualbox shared folder type for the `app` VM from its default VirtualBox by:
 ```
-CAC_APP_SHARED_FOLDER_TYPE=virtualbox vagrant up
+CAC_APP_SHARED_FOLDER_TYPE=nfs vagrant up
 ```
 7. See the app at http://localhost:8024! See OpenTripPlanner at http://localhost:9090.
 8. Running `./scripts/serve-js-dev.sh` on the host will rebuild the front-end app on file change (the browser must be reloaded manually to pick up the change). Alternatively, `cd /opt/app/src && npm run gulp-development` can be run manually in the VM to pick up changes to the static files.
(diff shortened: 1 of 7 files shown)


rachelekm (commented):
Thank you for these fixes! I'm able to successfully provision and see the running app, but I'm having trouble accessing the django admin at `localhost:8024/admin`. I get a 301 status code and it will redirect to `localhost:443/admin` are you seeing something similar?

rajadain:
I'm not exactly sure if this is the true explanation, but I believe this block:

https://github.com/azavea/cac-tripplanner/blob/71a8966c870116fe3c579e1a1fdb83dc54b73c68/deployment/ansible/roles/cac-tripplanner.app/templates/nginx-default.j2#L9-L13

Redirects `http://localhost:8024/admin` (which doesn't match any existing nginx configuration) to `http://localhost:443/admin` (via the `return 301` line), which then fails because the front-end then goes to :443 rather than :8024 as the Vagrantfile mapping expects:

https://github.com/azavea/cac-tripplanner/blob/71a8966c870116fe3c579e1a1fdb83dc54b73c68/Vagrantfile#L138-L139

By making the trailing slash optional, we now allow `http://localhost:8024/admin` to behave identically to `http://localhost:8024/admin/`, both of which now correctly redirect.

Fixed in [bd5b3d4](https://github.com/azavea/cac-tripplanner/pull/1387/commits/bd5b3d447eddd44277f3eb50cf26b7631c385980).

Will require reprovisioning app. Ready for another look.

rachelekm (approved):
Thank you, @rajadain! I re-provisioned and the app as well as the admin page work as expected 👍 Thanks for this work, those fixes will really improve any future spin-up and dev experience!

rajadain:
Before I merge this, I tried out the tests, and this one is failing:

https://github.com/azavea/cac-tripplanner/blob/71a8966c870116fe3c579e1a1fdb83dc54b73c68/python/cac_tripplanner/cac_tripplanner/tests.py#L58

Can you check on your end? It's possible I'm missing something in my local setup.

Issue #1399, opened 2025-02-27
Evaluate graph building in remote runner


rachelekm:
Could either be in github actions or aws ec2

Issue #1395, opened 2025-02-27
Setup AWS for graph building in GA workflow


rachelekm:
We'll move processes to fetch monthly data feed updates and build the new graph off of local and into a Github Action's workflow that can be manually triggered when cutting a release, specifically fetching GTFS feed files, getting the OpenStreetMap data and building the graph. Files that aren't fetched are stored in the fileshare or locally. In order to automate these actions using a workflow we need to first:

1. Setup a new S3 file repo and buckets, used for:
  -  storing elevation file, cac.tif, to be pulled and used in workflow
  - storing versioned graph object, generated from workflow and pulled for local development and monthly deployments
2. Setup CI user with appropriate permissions to above buckets

PR #1403, opened 2025-03-20
Add basic graph build workflow

rachelekm:
## Overview

Adds a basic github actions workflow to build a network graph and store it remotely. In addition to initial workflow steps, configures the following to allow build to successfully run:
1. Repurposes[ existing S3 bucket previously used for CI](https://element84.slack.com/archives/C04UC67U8SJ/p1556810309001800) for use by new workflow, including: 
     - Updating data (OSM, GTFS, and elevation) files in bucket to latest
     - Turning off public access to bucket
     - Turning on bucket versioning
2. Creates new `AmazonS3ReadWriteAccessCacOtpData` permissions policy that only allows read/write access to the `cleanair-otp-data` bucket
3. Create new `e84-graph-builder` IAM user for github actions access, set with `AmazonS3ReadWriteAccessCacOtpData` policy
4. Create access keys for `e84-graph-builder` user and adds keys to github actions repo secrets for use within workflow as well as saves .csv to 1password entry, `GoPhillyGo Github User AWS Access`

### Demo

Successful graph build: https://github.com/azavea/cac-tripplanner/actions/runs/14046461986

## Testing Instructions

 * Confirm no issues in graph build runner log
 * Navigate the `cleanair-otp-data` S3 bucket and confirm date of `Graph.obj` is the same as the last successful build
 * Follow new README instructions to fetch latest graph from S3
 * Provision OTP VM with latest graph and confirm no errors routing once up

## Checklist
- [ ] No gulp lint warnings
- [ ] No python lint warnings
- [ ] Python tests pass

Connects #1399 #1395

### DIFF
--- a/scripts/download-latest-graph.sh
+++ b/scripts/download-latest-graph.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+graph_path="otp_data/Graph.obj"
+
+if [[ ! -f "$graph_path" ]] || [[ "$*" == *"force"* ]]; then
+    echo "Pulling latest Graph.obj from S3..."
+    aws --profile=gophillygo s3 cp "s3://cleanair-otp-data/Graph.obj" "$graph_path"
+else
+    echo "Built graph already in otp_data directory, skip pulling latest from S3"
+fi
--- a/.github/workflows/build-graph.yml
+++ b/.github/workflows/build-graph.yml
@@ -10,19 +10,25 @@ on:
 jobs:
   docker-compose:
     runs-on: ubuntu-latest
-
+    env:
+      CAC_OTP_DATA_BUCKET: 'cleanair-otp-data'
     steps:
       - name: Checkout GitHub Repo
         uses: actions/checkout@v4
 
-      - name: Run Docker Compose
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_REGION }}
+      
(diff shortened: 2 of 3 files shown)


scripts/download-latest-graph.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+graph_path="otp_data/Graph.obj"
+
+if [[ ! -f "$graph_path" ]] || [[ "$*" == *"force"* ]]; then
+    echo "Pulling latest Graph.obj from S3..."
+    aws s3 cp "s3://cleanair-otp-data/Graph.obj" "$graph_path"
```
rajadain, line 9:
```suggestion
    aws --profile=gophillygo s3 cp "s3://cleanair-otp-data/Graph.obj" "$graph_path"
```

rajadain (approved):
> Turning on bucket versioning

We should keep an eye on the bucket size here. By storing past versions of very heavy objects, the bucket will likely grow quickly. At some point we may implement a strategy of keeping only the `N` most recent versions, or deleting versions more than `M` days old.

Nice job on reducing the permissions for the CI AWS account.

I'm building this on my computer right now with the downloaded graph from S3, which will take a while, so am approving this here with the expectation that will succeed. If it fails, can open a follow up issue.

Nice work!

scripts/download-latest-graph.sh
rachelekm: Thank you! That fix is added in [75ba6bb](https://github.com/azavea/cac-tripplanner/pull/1403/commits/75ba6bb3fe757afdfdd5ac189a74970f8e26cc46)

rachelekm:
Thanks for the review, @rajadain! I fixed the script with the above suggestion and also opened issue #1405 to capture updating the bucket versioning so we don't blow up the bucket size quickly (I would think only having last month's resources/graph build should be sufficient, which I noted in the issue)

Issue #1397, opened 2025-02-27
Write script for updating/validating feed files


rachelekm:
Refactor the ["Getting the GTFS file" ]("https://github.com/azavea/geospatial-apps/blob/master/gophillygo_deployment.md#getting-the-gtfs-files) section of the deployment instructions so that any manual actions are automated, including:
1. Editing SEPTA bus `routes.txt` files to convert trolleybuses
2. Confirm no error messages in feed corresponding .html file
   - If newline error, edit corresponding file
   - Else, continue to `feedvalidator` script
3. Updating the `feedvalidator` script
   - This used to work and could validate each file, confirming if issues with GTFS spec. May be outdated.
   - If invalid, stop build for manual fix

Bring these scripts into the new GA workflow

PR #1407, opened 2025-04-08
Add manual updates/validation to GTFS workflow

rachelekm:
## Overview

- Adds script updating SEPTA bus `routes.txt` file to convert replacement trolleybuses to be within OTPv1 level support
- Parses validation file with each GTFS feed and handles error in runner with message (warnings can pass)
- Updates README with base deployment instructions

### Demo

Successful validation & modifying GTFS feed files job: https://github.com/azavea/cac-tripplanner/actions/runs/14342009743/job/40203301765

Error correctly thrown with invalid GTFS feed file in job (see notes): https://github.com/azavea/cac-tripplanner/actions/runs/14338871722/job/40192902782 

### Notes

Temporarily pushed up a commit to force validating an altered septa_bus.zip in place of first downloading that GTFS file with the intention that this feed file should be caught by validation, the build stops, and the error correctly displays in the github action logs. See [error output here in latest build.](https://github.com/azavea/cac-tripplanner/actions/runs/14338871722/job/40192902782 )

## Testing Instructions

 * Confirm passing "Force & process source data" job with new modify step and updates to validation
 * For the sake of time/runner minutes latest passing workflow skips the actual graph build
 * Confirm updated date/times for all GTFS and OSM data files in `cleanair-otp-data` S3 bucket 
 * Download `septa_bus_extended.zip`, unzip, and open `routes.txt`. Confirm there are no rows with `route_type` of `11` (can search by `,,11,,`, if script unnsuccesful there will be 3 instances of this type)

## Checklist
- [ ] No gulp lint warnings
- [ ] No python lint warnings
- [ ] Python tests pass

Connects #1397

### DIFF
--- a/deployment/dataproc/gtfs-feed-fetcher/migrate_feeds_to_otpv1.py
+++ b/deployment/dataproc/gtfs-feed-fetcher/migrate_feeds_to_otpv1.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python
+"""Modifies GTFS files to OTP v1 standard.
+
+OTPv2 updates as GTFS standard updates, however not backported to OTPv1.
+We can't migrate to OTPv2, so manually modify GTFS for OTPv1 support level.
+"""
+import logging
+import zipfile
+import os
+
+logging.basicConfig()
+LOG = logging.getLogger()
+LOG.setLevel(logging.INFO)
+
+gtfs_dir = 'gtfs'
+
+def unzip_file(filename):
+    feed_path = gtfs_dir + "/" + filename + ".zip"
+    if zipfile.is_zipfile(feed_path):
+        with zipfile.ZipFile(feed_path, 'r') as zip_ref:
+            zip_ref.extractall(gtfs_dir)
+            LOG.debug('Unzipped %s', filename)
+            return
+    else:
+        LOG.warn('File %s does not look like a valid zip file.', filename)
+
+def clean_up_txt_files():
+    for root, _, files in os.walk(gtfs_dir):
+        for file in files:
+            if file.endswith('.txt'):
+                os.remove(os.path.join(root, file))
+    LOG.debug('Cleaned up .txt files')
+
+def re_zip_files(filename):
+    feed_path = gtfs_dir + "/" + filename + ".zip"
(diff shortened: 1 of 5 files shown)


deployment/dataproc/gtfs-feed-fetcher/migrate_feeds_to_otpv1.py
```diff
@@ -0,0 +1,74 @@
...
+    content=''
+    try:
+        with open(txt_path_to_update, 'r') as file:
+            content = file.read()
+            content = content.replace(',,11,,', ',,3,,')
+        with open(txt_path_to_update, 'w') as file:
+            file.write(content)
+        LOG.debug('Modified %s', txt_path_to_update)
+    except Exception as e:
+        raise ValueError('Error modifying %s: %s', txt_path_to_update, e)
```
rajadain, line 58: This is a little fragile. These .txt files are actually formatted as CSV, so we should be able to use a CSV reader and update the `route_type` column. However, this is working well enough for now, and we do have an `except` that will pause on this failure, so perhaps this can be made more robust in the future, when it fails.

image

rajadain (approved):
+1 tested, inspected the output on S3 and the GitHub Actions logs. This looks good!

deployment/dataproc/gtfs-feed-fetcher/migrate_feeds_to_otpv1.py
rachelekm: This is a great point, thank you! This is definitely a file subject to change with the new SEPTA rollout so it would benefit to be robust. Documenting that we talked briefly about solutions, I ended up moving forward with a csv reader and iterating through the rows and replacing values in the correct column as needed. This is a smaller file (~10kb, <200 rows) so bringing in a library to replace these values more efficiently doesn't feel worth it. Here is t[hat new step in the latest successful build ](https://github.com/azavea/cac-tripplanner/actions/runs/14480346972/job/40615799660)(note I pushed up a temp commit to test this job but that prevents building the graph). Let me know what you think!

rachelekm:
Thank you for the second look! Rebased and dropped the commit containing the commented out code/build trigger for this branch 👍

